PR #30, opened 2024-12-02
Update readme GIFs

bwateratmsft:
The cost.gif readme is not yet used but will be once we release the extension.

cost.png should not be deleted because it is still in use by the released version of the extension.

The other GIFs are close enough to the previous recording that replacing them is not an issue.


MRayermannMSFT:
lgtm

PR #449, opened 2025-10-30
Foundry agent sdk sample

JasonYeMSFT:
Add agent sdk samples so tools can consume them from publicly available urls.


JasonYeMSFT:
The motivation is to make them publicly available, like the cli installation instructions. https://github.com/microsoft/GitHub-Copilot-for-Azure/tree/main/docs/cli-install

fanyang-mono:
I feel that it would be hard to remember to update these content as time comes, since it lives in a different repo than the tools.

PR #513, opened 2026-01-13
Add the azure-resource-visualizer skill

tmeschter:
Adds a skill that can analyze an Azure resource group and create a Mermaid diagram showing the relationships between the various resources.


claude/skills/azure-resource-visualizer/SKILL.md
```diff
@@ -0,0 +1,230 @@
+---
+name: azure-resource-visualizer
+description: Analyze Azure resource groups and generate detailed Mermaid architecture diagrams showing the relationships between individual resources. Use this skill when the user asks for a diagram of their Azure resources or help in understanding how the resources relate to each other.
```
kvenkatrajan, line 3: does the resource need to be within a given resource group or it can pick up the relationship from across the entire subscription? Maybe good t call that out.

tmeschter: The skill looks at all resources within a resource group, but nothing would prevent the user from specifying multiple resource groups.

kvenkatrajan (approved):

PR #514, opened 2026-01-13
Add the azure-role-selector skill

tmeschter:
Add the azure-role-selector skill to the Claude plugin.


msalaman (approved):
Validated that it worked on my own VS code editor

PR #519, opened 2026-01-17
Add an agent to create Azure-related skills

tmeschter:
Add an agent that gathers user requirements for a new agent skill, including which scenarios it should address and any tools it should utilize. The output of the agent is a REQUIREMENTS.md file that can be handed off to the "Plan" agent to utilize (which will in turn hand off to "Agent" for the actual implementation).


kvenkatrajan (commented):
Should it be for Azure specific skills and specify azure mcp and other azure tooling?

PR #539, opened 2026-01-23
Add a brainstorming agent

tmeschter:
Add a brainstorming agent that helps users identify relevant scenarios for a skill. This skill then hands off to the Skill Creator agent to gather further requirements, then the Plan agent to create a plan, and finally just "agent" to implement the plan.

### DIFF
--- a/.github/agents/SkillBrainstormer.agent.md
+++ b/.github/agents/SkillBrainstormer.agent.md
@@ -1,11 +1,11 @@
 ---
-name: Skill Brainstormer
-description: Helps brainstorm ideas for a new agent skill based on user provided tools and documentation.
+name: Azure Skill Brainstormer
+description: Helps brainstorm ideas for a new Azure-related agent skill based on user provided tools and documentation.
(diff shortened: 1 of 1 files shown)


.github/agents/SkillBrainstormer.agent.md
```diff
@@ -0,0 +1,46 @@
+---
+name: Skill Brainstormer
```
kvenkatrajan: Is this generic enough for any tool or do we need to put Azure MCP here? Im fine with either just asking. The vision is the coding agent can pick up the list of tools from Azure MCP for a given server and brainstorm on the skills. Have we tested the agent that way?

tmeschter: Yeah, I'll update the name and description to be more Azure-specific.

I tested it against azqr, and it did pick up on the azqr tool in Azure MCP without me explicitly mentioning that.

tmeschter: > Yeah, I'll update the name and description to be more Azure-specific.

Though since this is an agent the name and description primarily need to be meaningful to the user, not the LM.

PR #547, opened 2026-01-26
Added skill for deploying to azure function apps

harryli0108:
image
image

### DIFF
--- a/claude/skills/azure-function-app-deployment/SKILL.md
+++ b/claude/skills/azure-function-app-deployment/SKILL.md
@@ -59,19 +59,48 @@ async function validatePrerequisites() {
     throw new Error('Not authenticated with Azure CLI. Run: az login');
   }
   
-  // Check Azure Functions Core Tools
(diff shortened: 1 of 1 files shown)


claude/skills/azure-function-app-deployment/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+  try {
+    await exec('az account show');
+    checks.push({ name: 'Azure CLI', status: 'authenticated' });
+  } catch (error) {
+    throw new Error('Not authenticated with Azure CLI. Run: az login');
+  }
+  
+  // Check Azure Functions Core Tools
+  try {
+    await exec('func --version');
```
tmeschter, line 64: You might want to add instructions for installing `func`, but that could be a follow-up change.

harryli0108: Addressed in the latest commit

PR #559, opened 2026-01-27
Add warning for potential race condition with azd up when image is still being published and container app is getting deployed.

tdaroly:
This is a race condition when using azd up with container app where the image while still being published is being referenced in the deployment. It is transient BUT i saw it twice when testing with vscode+custom skills, id imagine itd be there in cli as well.

Other folks have seen this to throught the years and the fix may be to just use az cli:
 [az containerapp up](https://learn.microsoft.com/en-us/cli/azure/containerapp?view=azure-cli-latest#az-containerapp-up)

Issues:
https://github.com/Azure/azure-cli/issues/29974
https://github.com/Azure/azure-dev/issues/1753

### DIFF
--- a/plugin/skills/azure-aca-deployment/SKILL.md
+++ b/plugin/skills/azure-aca-deployment/SKILL.md
@@ -183,10 +183,37 @@ When using `azd up`, Bicep provisioning runs BEFORE images are built and pushed.
 ContainerAppOperationError: MANIFEST_UNKNOWN: manifest tagged by "latest" is not found
 ```
 
-**Workarounds:**
-1. **Pre-push an image** before running `azd up`
-2. **Run separately**: `azd provision` then manually push image, then `azd deploy`
-3. **Use `az containerapp up`** instead, which handles this automatically
+**Recommended Solution: Use a Placeholder Image**
+
+Use a placeholder "hello world" image for initial Container App provisioning. During `azd deploy`, the actual application image will be built, pushed to ACR, and deployed to replace the placeholder.
+
+```bicep
+// In your main.bicep, use a placeholder for initial deployment
+param containerImage string = 'mcr.microsoft.com/k8se/quickstart:latest'
+
+resource containerApp 'Microsoft.App/containerApps@2024-03-01' = {
+  // ...
+  properties: {
+    template: {
+      containers: [
+        {
+          name: projectName
+          image: containerImage  // Placeholder on provision, real image on deploy
+          // ...
+        }
+      ]
+    }
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


plugin/skills/azure-aca-deployment/SKILL.md
```diff
@@ -175,6 +175,19 @@ async function generateInfrastructure(projectName: string) {
...
+When using `azd up`, Bicep provisioning runs BEFORE images are built and pushed. If your Container App Bicep references your ACR image directly (e.g., `myregistry.azurecr.io/myapp:latest`), provisioning will fail with:
+
+```
+ContainerAppOperationError: MANIFEST_UNKNOWN: manifest tagged by "latest" is not found
+```
+
+**Workarounds:**
+1. **Pre-push an image** before running `azd up`
+2. **Run separately**: `azd provision` then manually push image, then `azd deploy`
+3. **Use `az containerapp up`** instead, which handles this automatically
```
wbreza: A better workaround would be to use a placeholder "hello world" image for the initial container app deployment. Then during `azd deploy` the actual image would be built, pushed and deployed.

tdaroly: I agree AND that is what i had initially as well. Use a placeholder startup image and that gets replaced with the generated image. Only thing im worried/not sure of is that will the place holder image get replaced with the correct image when azd deploy/azd up is ran

tdaroly: Pushed commit to use placeholder image :)

plugin/skills/azure-aca-deployment/SKILL.md
```diff
@@ -175,6 +175,46 @@ async function generateInfrastructure(projectName: string) {
 
 **Key insight**: Always validate infrastructure before deploying using preview commands to catch configuration issues early.
 
+### ⚠️ Known Issue: Image Race Condition with `azd up`
```
JasonYeMSFT, line 178: I don't feel using an external placeholder image is a good idea. From a security point of view, the user will technically need to vet a completely different image becuase it can execute any code before azd create a new revision of the user's app to override it. I think build/tag/load an initial version of the container image of the user's app outside azd up is a better approach.

plugin/skills/azure-aca-deployment/SKILL.md
kvenkatrajan: @wbreza are all these included in the new restructuring?

kvenkatrajan:
ACA skill was moved over to azure-deploy.

PR #584, opened 2026-01-28
Document azd usage for plain HTML site deployment

Petermarcu:
Added instructions for using azd with plain HTML sites, including project structure and configuration details.

### DIFF
--- a/plugin/skills/azure-deploy/SKILL.md
+++ b/plugin/skills/azure-deploy/SKILL.md
@@ -501,6 +501,68 @@ swa deploy ./dist --deployment-token "$TOKEN" --env production
 rm -rf dist
 ```
 
+**Using azd for Plain HTML Sites:**
+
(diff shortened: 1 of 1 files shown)


plugin/skills/azure-deploy/SKILL.md
```diff
@@ -501,6 +501,68 @@ swa deploy ./dist --deployment-token "$TOKEN" --env production
...
+  }
+  properties: {}
+}
+
+output name string = staticWebApp.name
+output uri string = 'https://${staticWebApp.properties.defaultHostname}'
+```
+
+Then deploy with:
+```bash
```
charris-msft, line 562: do we need to include PowerShell? 
I know the command is the same so it seems the model will probably have no issue.

Petermarcu: I think we can leave it as is for now.

PR #605, opened 2026-01-28
Improve microsoft-foundry skill with better triggers and structure

annaji-msft:


plugin/skills/microsoft-foundry/references/workflows.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Troubleshooting
+
+### Deployment Issues
+
+**Problem: Deployment Pending/Fails**
+
+```bash
+# Check deployment status
+az cognitiveservices account deployment show \
```
Petermarcu, line 260: @wbreza has been helping get all the other skills standardized on azd. What should the flow look like for that and is this going to cause things to flip back and forth between the two if the app is deploying with azd and IaC but the models are az?

annaji-msft: @Petermarcu I left it as draft as the original skill wasn’t generic enough and skill was not getting triggered for simple prompts. I will let the team take it from here and provide guidance and take it forward. Cc @JasonYeMSFT

JasonYeMSFT: @annaji-msft We have added integration tests for skill invocation and adopted a common pattern for the skill descriptions. Could you please apply the same pattern for the Foundry skill description change and add a skill invocation tests? If your change works as expected we can merge it for now.

PR #617, opened 2026-01-29
Add skill integration test framework

JasonYeMSFT:


integration-tests/package-lock.json
spboyer, line 1: add to gitignore

integration-tests/src/azure-role-selector/azureRoleSelectorTests.ts
```diff
@@ -0,0 +1,26 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from "assert";
+import { areToolCallsSuccess, doesAssistantMessageIncludeKeyword, isSkillInvoked, run } from "../runner.js";
+import { describe, it } from "mocha";
+
+describe("azureRoleSelectorTests", function () {
+    it("invokes azure-role-selector skill for AcrPull prompt", async function () {
+        const agentMetadata = await run({
+            prompt: "What role should I assign to my managed identity to read images in a Azure Container Registry?"
```
spboyer, line 13: We should consider being able to run various prompts here that satisfy the skill selector, perhaps a prompt.txt in the /skill-name folder

JasonYeMSFT: Do you mean running multiple prompts within one agent flow? The current runner doesn't have a way to do multiple rounds of interaction with the agent. And it will be challenging since the response of prior rounds are nondeterministic. If you mean testing multiple prompts separately, we can have a dedicated test case for each prompt.

integration-tests/package-lock.json
JasonYeMSFT: Any reason to gitignore the package-lock.json? Most of the projects I worked with commit package-lock.json to track the last working state of the depencies. This is handy in case some ^ denoted dependencies upgrades and breaks the code.

PR #706, opened 2026-01-31
Add non-interactive mode to copilot runner

jeo02:
- Added an option to run the copilot runner in non-interactive mode and yolo mode. This will make sure we have a full cycle of the skill to evaluate upon. 
- Added examples for azure-deploy where it completes the prompts full task.
- Need to override the global timeout since these tests take longer.

### DIFF
--- a/tests/utils/agent-runner.ts
+++ b/tests/utils/agent-runner.ts
@@ -101,7 +101,7 @@ export async function run(config: TestConfig): Promise<AgentMetadata> {
     }
 
     // Copilot client with yolo mode and non-interactive session.
-    // Longer chats but will have full conversation of skill. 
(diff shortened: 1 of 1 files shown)


tests/reports/junit.xml
```diff
@@ -1,357 +1,45 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
JasonYeMSFT: This file should not be committed. Let's wait for this to merge and resolve the conflict to remove this file. https://github.com/microsoft/GitHub-Copilot-for-Azure/pull/693

jeo02: yea i know I'm waiting to merge it out :)

kvenkatrajan (approved):

PR #719, opened 2026-02-02
fix: add cross-env to all jest scripts for Windows compatibility

spboyer:
## Problem
On Windows, running `npm run test:skill -- appinsights-instrumentation` would run **all tests** instead of filtering to the specified pattern.

## Solution
Added `cross-env` prefix to all jest-based npm scripts in `tests/package.json`. This ensures consistent environment variable and argument passing across Windows, Mac, and Linux.

## Changes
- Added `cross-env` to: test, test:unit, test:verbose, test:coverage, test:ci, test:watch, test:skill, update:snapshots

## Testing
Verified locally that `npm run test:skill -- appinsights-instrumentation` now correctly filters to only appinsights-instrumentation tests.

## Usage
```bash
# Run specific skill tests
npm run test:skill -- appinsights-instrumentation

# Or use the base test command
npm run test -- --testPathPattern=appinsights-instrumentation
```

### DIFF
--- a/tests/package.json
+++ b/tests/package.json
@@ -4,16 +4,16 @@
   "description": "Test suite for Azure Copilot Skills",
   "private": true,
   "scripts": {
-    "test": "jest",
-    "test:unit": "jest --testPathIgnorePatterns=\"node_modules|_template|integration\"",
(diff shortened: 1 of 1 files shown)


spboyer (commented):
✅ `cross-env` is already listed in devDependencies at version `^7.0.3` (line 21 of package.json). No changes needed.

PR #722, opened 2026-02-02
Refine entra-app skill

JasonYeMSFT:
Refine the frontmatter to following the common pattern. Added a Bicep template example, which is a better way to manage app registration in a few mentioned scenarios than Azure CLI.

Skill invocation test hit 5/5 for all existing tests.


kvenkatrajan:
@JasonYeMSFT - Were we able to add an integration test for this?

plugin/skills/entra-app-registration/references/BICEP-EXAMPLE.bicep
```diff
@@ -0,0 +1,199 @@
+// Bicep template for Microsoft Entra App Registration
```
saikoumudi, line 1: nit: lowercasing on reference file name

PR #728, opened 2026-02-02
sensei: improve azure-keyvault-expiration-audit frontmatter

tmeschter:
1. Improve frontmatter
   - Switch to a multi-line string that can accommodate colon (":") characters properly.
   - Add explicit "USE FOR" and "DO NOT USE FOR" sections
2. Add trigger tests.
3. Add unit tests.


tests/azure-keyvault-expiration-audit/triggers.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        const result = triggerMatcher.shouldTrigger(prompt);
+        expect(result.triggered).toBe(true);
+        // Trigger matcher requires >= 2 keywords or >= 20% confidence
+        expect(result.matchedKeywords.length).toBeGreaterThanOrEqual(2);
+      }
+    );
+  });
+
+  describe('Should NOT Trigger', () => {
+    // Prompts that should NOT trigger - avoid matching >= 2 keywords
```
saikoumudi, line 46: I think once we add all the tests, we need to improve on the prompts in should not trigger, since the prompts are very similar in all other tests too

PR #746, opened 2026-02-03
Deploy coverage for each resource types

jeo02:
- Added tests to cover SWA, appservice, azure functions and ACA.
- Each can take a while so just be aware to expect a long wait while running.

### DIFF
--- a/tests/azure-deploy/integration.test.ts
+++ b/tests/azure-deploy/integration.test.ts
@@ -91,42 +91,144 @@ describeIntegration(`${SKILL_NAME} - Integration Tests`, () => {
 
   // Need to be logged into azd for these tests. 
   // azd auth login
+
+  // Static Web Apps (SWA)
   test('creates whiteboard application and deploys to Azure', async () => {
     const agentMetadata = await run({
-      prompt: 'Create a whiteboard application and deploy to azure',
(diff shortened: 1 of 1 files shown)


tests/azure-deploy/integration.test.ts
```diff
@@ -91,42 +91,144 @@ describeIntegration(`${SKILL_NAME} - Integration Tests`, () => {
...
+    expect(isValidateInvoked).toBe(true);
+    expect(isPrepareInvoked).toBe(true);
     expect(containsDeployLinks).toBe(true);
   });
 
+  // App Service
   test('creates discussion board and deploys to Azure', async () => {
     const agentMetadata = await run({
-      prompt: 'Create a discussion board and deploy to Azure',
+      prompt: 'Create a discussion board application and deploy to Azure using my current subscription in eastus region.',
```
kvenkatrajan, line 133: did we need to do this because SWA was using an invalid region? If so this should be fixed now

CC: @jongio

JasonYeMSFT (approved):

tests/azure-deploy/integration.test.ts
jeo02: I used it because it seems to be set as a requirement that even yolo mode won't get past.

PR #747, opened 2026-02-03
Add test for foundry model listing

JasonYeMSFT:
Also in this PR:
- Add CopilotClient type import back.
- Add an option to customize system prompt for the agent run.

### DIFF
--- a/tests/microsoft-foundry/integration.test.ts
+++ b/tests/microsoft-foundry/integration.test.ts
@@ -16,7 +16,8 @@ import {
   shouldSkipIntegrationTests,
   getIntegrationSkipReason,
   doesAssistantMessageIncludeKeyword,
-  areToolCallsSuccess
+  areToolCallsSuccess,
+  isSkillInvoked
 } from "../utils/agent-runner";
 import * as fs from "fs";
 import { AIProjectClient } from "@azure/ai-projects";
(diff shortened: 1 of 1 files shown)


tests/microsoft-foundry/integration.test.ts
```diff
@@ -91,6 +93,31 @@ describeIntegration(`${SKILL_NAME} - Integration Tests`, () => {
...
+        content: `Use ${projectEndpoint} as the project endpoint when calling Foundry tools.`
+      },
+      prompt: "What's the official name of GPT 5 in Foundry?",
+      nonInteractive: true
+    });
+
+    const areFoundryToolCallsSuccess = areToolCallsSuccess(agentMetadata, "azure-foundry");
+    const isCorrectModelNameInResponse = doesAssistantMessageIncludeKeyword(agentMetadata, "gpt-5", { caseSensitive: true });
+    expect(areFoundryToolCallsSuccess).toBe(true);
+    expect(isCorrectModelNameInResponse).toBe(true);
```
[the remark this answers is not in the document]
JasonYeMSFT: I didn't notice this before but the existing behavior is wrong. People run these tests expecting copilot-sdk to be available. If it isn't then they should be notified about it by a failed test.

tests/microsoft-foundry/integration.test.ts
```diff
@@ -91,6 +93,31 @@ describeIntegration(`${SKILL_NAME} - Integration Tests`, () => {
...
+    const agentMetadata = await run({
+      systemPrompt: {
+        mode: "append",
+        content: `Use ${projectEndpoint} as the project endpoint when calling Foundry tools.`
+      },
+      prompt: "What's the official name of GPT 5 in Foundry?",
+      nonInteractive: true
+    });
+
+    const areFoundryToolCallsSuccess = areToolCallsSuccess(agentMetadata, "azure-foundry");
```
tmeschter, line 116: Should we be checking for a skill as well as a tool?

JasonYeMSFT: Yes

PR #766, opened 2026-02-04
Include diagnosis

tmeschter:
Update the `file-test-bug` skill to include an diagnosis. This won't always be correct but at least it provides some high-level explanation for what went wrong.


tmeschter:
See #765 for an example of the output of the updated skill.

PR #794, opened 2026-02-05
Consolidate azure-security-hardening

tmeschter:
Consolidate the `azure-security-hardening` skill into `azure-prepare`. The existing `azure-security-hardening` skill contains a number of important requirements and best practices that should be considered during the development of an application, so it is useful to incorporate it into that workflow rather than keeping it as a standalone skill that would only be picked up when the user specifically asks about security. 

Here we move the content of the `azure-security-hardening` skill to azure-prepare/references/security.md, and reference it from various places.

### DIFF
--- a/plugin/skills/appinsights-instrumentation/SKILL.md
+++ b/plugin/skills/appinsights-instrumentation/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: appinsights-instrumentation
-description: 'Foo Instrument a webapp to send useful telemetry data to Azure App Insights'
+description: 'Instrument a webapp to send useful telemetry data to Azure App Insights'
 ---
 
 # AppInsights instrumentation


plugin/skills/_shared/security/data-protection.md
```diff
@@ -0,0 +1,38 @@
+# Data Protection
```
kvenkatrajan: what other areas do we for see under shared? should sdks be in here? CC: [ronniegeraghty](https://github.com/ronniegeraghty)

tmeschter: I don't think skills can actually reference files that are outside their folder--I never saw the files under _shared/security being picked up. When I moved them under the azure-prepare skill they started to get picked up.

plugin/skills/appinsights-instrumentation/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: appinsights-instrumentation
-description: 'Instrument a webapp to send useful telemetry data to Azure App Insights'
+description: 'Foo Instrument a webapp to send useful telemetry data to Azure App Insights'
```
fanyang-mono: Is this change intentional?

fanyang-mono (approved):
LGTM!

plugin/skills/appinsights-instrumentation/SKILL.md
tmeschter: Oops. No, let me fix that.

PR #849, opened 2026-02-10
Fix issues when timeout occurs in test

jeo02:
- Currently when a test times out and it is running the agent runner it will clean up the resources in a finally block. However, if the test times out it will never reach that finally block leading to a memory leak. 
- The markdown files would also not be generated as it will hang on the `send` or `sendAndAwait` promises. 
- Instead, if we have an `afterEach` to clean up the resources and generate the file we are guaranteed to clean up and produce the markdown.


tests/utils/agent-runner.ts
```diff
@@ -250,132 +260,145 @@ function writeMarkdownReport(config: TestConfig, agentMetadata: AgentMetadata):
...
  */
-export async function run(config: TestConfig): Promise<AgentMetadata> {
-  const testWorkspace = fs.mkdtempSync(path.join(os.tmpdir(), "skill-test-"));
-  const FOLLOW_UP_TIMEOUT = 1800000; // 30 minutes
+export function useAgentRunner() {
+  let currentCleanups: RunnerCleanup[] = [];
+
+  if(isTest()) {
+    afterEach(async () => {
+      for (const cleanup of currentCleanups) {
```
JasonYeMSFT: The clean up logic should be universal no matter whether the agent run is for a test or not. And ideally reuse the same logic even if we have to conditionally do it at different places depending on whether the run is for a test or not.

The markdown generation is only applicable in the test context. Let's pull it out from the clean up logic and only do it when it's in a test context.

PR #877, opened 2026-02-11
Add typecheck to pr ci pipeline

JasonYeMSFT:
This would allow us to catch type errors like this in PR CI jobs. https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/81d3e02c715982c1948aeaf014e1f8917ab77af6/tests/azure-ai/integration.test.ts#L71

### DIFF
--- a/tests/azure-ai/integration.test.ts
+++ b/tests/azure-ai/integration.test.ts
@@ -68,7 +68,7 @@ describeIntegration(`${SKILL_NAME} - Integration Tests`, () => {
 
       for (let i = 0; i < RUNS_PER_PROMPT; i++) {
         try {
-          const agentMetadata = await run({
+          const agentMetadata = await agent.run({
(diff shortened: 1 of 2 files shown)


tests/azure-ai/integration.test.ts
```diff
@@ -68,7 +68,7 @@ describeIntegration(`${SKILL_NAME} - Integration Tests`, () => {
 
       for (let i = 0; i < RUNS_PER_PROMPT; i++) {
         try {
-          const agentMetadata = await run({
+          const agentMetadata = await agent.run({
```
tmeschter, line 71: What happens if you call `run` instead of `agent.run`?

JasonYeMSFT: The latent transpiler will throw an error before any code gets executed for it because there is no such exported run function.

PR #916, opened 2026-02-13
Add Azure Resource Graph integration across plugin skills

charris-msft:
## Summary

Add Azure Resource Graph (ARG) integration across plugin skills for efficient cross-subscription, cross-resource-group resource discovery.

## What changed

- **New `azure-resource-lookup` skill** that uses ARG via `az graph query` for cross-subscription resource discovery
- **ARG reference docs** with curated KQL query patterns added to 5 existing skills: `azure-compliance`, `azure-cost-optimization`, `azure-diagnostics`, `azure-resource-visualizer`
- **Full test suite**: trigger tests, unit tests, integration tests, and snapshot tests for the new skill

**19 files changed, +1,227 lines**

## Why it matters

Many MCP tools require a resource group name upfront (e.g., `sql_server_list`), and some resource types like Container Apps and Web Apps have no MCP list command at all. This forces the agent to iterate through every resource group one-by-one — slow, incomplete, and often misses resources in unexpected RGs or other subscriptions.

Resource Graph solves this with a single KQL query that spans all resource groups and subscriptions instantly.

## Measured impact (19-prompt test suite, 8 runs across plugin versions)

- Skills fired **0/19 times** in all previous plugin builds (private-v1, v2)
- With this change: **3/19 skills invoked**, including 2 Resource Graph queries
- Container Apps discovery: went from PARTIAL (couldn't find apps without RG names) to PASS (single ARG query found all 11 apps)
- SQL Server discovery: went from iterating 17 RGs to a single ARG query finding 9 servers across subscriptions
- Pass rate held steady at 89% (17/19) — the 2 partials are unrelated permission issues (Key Vault elicitation, AI Search RBAC)

### DIFF
--- a/tests/azure-resource-lookup/integration.test.ts
+++ b/tests/azure-resource-lookup/integration.test.ts
@@ -14,7 +14,6 @@
 import {
   useAgentRunner,
   isSkillInvoked,
-  areToolCallsSuccess,
   doesAssistantMessageIncludeKeyword,
   shouldSkipIntegrationTests
 } from "../utils/agent-runner";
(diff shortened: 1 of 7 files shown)


tests/azure-resource-lookup/integration.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+import {
+  useAgentRunner,
+  isSkillInvoked,
+  areToolCallsSuccess,
+  doesAssistantMessageIncludeKeyword,
+  shouldSkipIntegrationTests
+} from "../utils/agent-runner";
+
+const SKILL_NAME = "azure-resource-lookup";
```
kvenkatrajan, line 21: can we have a test for isSkillInvoked? Here's an example -https://github.com/microsoft/GitHub-Copilot-for-Azure/pull/888/changes#diff-7e42c3802abbe802c3021893ef54e1093310afd467c4d67f93b5f7bc248e1c68R65

charris-msft: Added

PR #918, opened 2026-02-13
Fix ci report generation

JasonYeMSFT:
Failed tests were aborting the steps, causing the report generation command to be skipped. This PR fixes it by always executing the report generation.

Two more not directly related changes:
- Removed integration test related code in the Test All Skills workflow. Appended a "non-integration" to its name to clarify what it does.
-  Set cancel-in-progress: false for integration test all workflows.

### DIFF
--- a/.github/workflows/test-all-skills.yml
+++ b/.github/workflows/test-all-skills.yml
@@ -3,7 +3,7 @@
 # Runs the complete test suite for all skills.
 # Triggered manually or on significant changes.
 
-name: Test All Skills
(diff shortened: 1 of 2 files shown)


.github/workflows/test-all-skills.yml
jeo02, line 1: (nit) might want to update the header

JasonYeMSFT: I'll do it in a follow up.

PR #1012, opened 2026-02-23
fix(azure-prepare): route Functions to composition algorithm instead of non-existent skill

### LINKED ISSUE #1011: azure-prepare skill: Agent bypasses base+recipe composition algorithm for Functions
## Problem

When using the `azure-prepare` skill to create Azure Functions projects, the agent synthesizes IaC from scratch instead of following the **base + recipe composition algorithm** defined in `services/functions/templates/recipes/composition.md`.

This leads to:
- Subtly incorrect IaC (e.g., wrong Flex Consumption resource properties)
- Missing security best practices that base templates include
- Errors that require manual fixes (e.g., `storage_container_endpoint` vs `storage_account_name`)

## Root Cause

There is a **discoverability gap** in the skill references. The composition algorithm exists and is well-documented, but the agent never loads it.

### The Gap in `research.md`

Line 76-78 in `references/research.md`:
```markdown
| Scenario | Invoke Skill |
|----------|--------------|
| Using Azure Functions | `azure-functions` |
```

**Problem:** There is no `azure-functions` skill. The agent cannot invoke it, so it falls back to synthesizing IaC.

### What the Agent Did

1. Loaded `services/functions/templates/http.md` (template list only)
2. Loaded `recipes/cosmosdb/terraform/cosmos.tf` (recipe file only)
3. **Never saw** `selection.md` or `composition.md` which contain the algorithm
4. Synthesized Terraform from scratch instead of using `azd init -t`

### What Should Have Happened

1. Load `services/functions/templates/selection.md` → decision tree
2. Follow `recipes/composition.md` → Step 1: Fetch Base Template with `azd init -t`
3. Apply recipe on top of base

## Evidence

The composition algorithm in `recipes/composition.md` explicitly says:

> **Critical Rules:**
> 1. **NEVER synthesize Bicep or Terraform from scratch** — always start from base template IaC

But without a clear path from `research.md` → `selection.md` → `composition.md`, the agent never sees this rule.

## Suggested Fix

Update `references/research.md` to explicitly route Functions to the composition algorithm:

```markdown
### Step 2: Invoke Related Skills (When Deeper Guidance Needed)

| Scenario | Action |
|----------|--------|
| Using Azure Functions | Load `services/functions/templates/selection.md` → Follow `recipes/composition.md` algorithm |
| PostgreSQL with passwordless auth | Invoke `azure-postgres` skill |
| ...
```

Alternatively, create an `azure-functions` skill that simply loads the template selection and composition references.

## Files to Update

- `plugin/skills/azure-prepare/references/research.md` - Fix the "azure-functions skill" reference
- Optionally: `plugin/skills/azure-prepare/references/generate.md` - Add prominent link to composition algorithm for Functions

## Reproduction

1. Ask: "Create a URL shortener using Azure Functions with Cosmos DB and Terraform"
2. Observe: Agent synthesizes IaC from scratch
3. Expected: Agent runs `azd init -t functions-quickstart-typescript-azd-tf` then applies cosmosdb recipe

paulyuk:
## Summary

Fixes #1011

The agent was bypassing the base+recipe composition algorithm for Azure Functions because:
1. `research.md` referenced a non-existent `azure-functions` skill
2. `README.md` did not contain explicit instructions to use the composition workflow

## Fix

Put the mandatory instructions where they belong — in `services/functions/README.md`:

- Add prominent callout box with mandatory composition workflow
- Link directly to `selection.md` (decision tree) and `composition.md` (algorithm)
- Update References section to prioritize these guides
- Simplify `research.md` to just point to `README.md` (standard pattern)

## Agent Flow (After)

```
research.md → README.md → selection.md → composition.md → azd init -t
```

### `services/functions/README.md`
- Added mandatory composition workflow callout at top
- Added links to `selection.md` and `composition.md` in References

### `research.md`
- Simplified Functions row to standard pattern: `| Azure Functions | [Functions](services/functions/README.md) | — |`
- Simplified Serverless Event-Driven pattern (README.md now contains the instructions)

## Testing

Verified all linked files exist:
- `services/functions/README.md` ✅
- `services/functions/templates/selection.md` ✅  
- `services/functions/templates/recipes/composition.md` ✅

### DIFF
--- a/plugin/skills/azure-prepare/references/research.md
+++ b/plugin/skills/azure-prepare/references/research.md
@@ -116,24 +116,24 @@ Add research findings to `.azure/plan.md`:
 
 ### Web Application + API + Database
 
-1. Load: `services/container-apps/README.md` → `bicep.md`, `scaling.md`
-2. Load: `services/cosmos-db/README.md` → `partitioning.md`
-3. Load: `services/key-vault/README.md`
+1. Load: [services/container-apps/README.md](services/container-apps/README.md) → [bicep.md](services/container-apps/bicep.md), [scaling.md](services/container-apps/scaling.md)
+2. Load: [services/cosmos-db/README.md](services/cosmos-db/README.md) → [partitioning.md](services/cosmos-db/partitioning.md)
+3. Load: [services/key-vault/README.md](services/key-vault/README.md)
 4. Invoke: `azure-observability` (monitoring setup)
 5. Invoke: `azure-security-hardening` (security baseline)
 
 ### Serverless Event-Driven
 
-1. Load: `services/functions/README.md` (contains mandatory composition workflow)
-2. Load: `services/event-grid/README.md` or `services/service-bus/README.md` (if using messaging)
-3. Load: `services/storage/README.md` (if using queues/blobs)
+1. Load: [services/functions/README.md](services/functions/README.md) (contains mandatory composition workflow)
+2. Load: [services/event-grid/README.md](services/event-grid/README.md) or [services/service-bus/README.md](services/service-bus/README.md) (if using messaging)
+3. Load: [services/storage/README.md](services/storage/README.md) (if using queues/blobs)
 4. Invoke: `azure-observability` (distributed tracing)
 
 ### AI Application
 
 1. Invoke: `microsoft-foundry` (AI patterns and best practices)
-2. Load: `services/container-apps/README.md` → `bicep.md`
-3. Load: `services/cosmos-db/README.md` → `partitioning.md` (vector storage)
+2. Load: [services/container-apps/README.md](services/container-apps/README.md) → [bicep.md](services/container-apps/bicep.md)
+3. Load: [services/cosmos-db/README.md](services/cosmos-db/README.md) → [partitioning.md](services/cosmos-db/partitioning.md) (vector storage)
 4. Invoke: `azure-security` (API key management)
 
 ## After Research


paulyuk:
To test this, you can use this prompt:
```
Create a URL shortener service using Azure Functions that creates short links and redirects users to the original URL and deploy to Azure using my current subscription in
  eastus2 region using terraform 
```

It will not only succeed without retries, but it will not miss RBAC rules and you will see it uses the composition.md algorithm.

plugin/skills/azure-prepare/references/research.md
```diff
@@ -70,20 +70,25 @@ Selected: Container Apps, Cosmos DB, Key Vault
...
 | Building AI applications | `microsoft-foundry` |
 | Cost-sensitive deployment | `azure-cost-optimization` |
 
-**Skill Invocation Pattern:**
+**Skill/Reference Invocation Pattern:**
 ```
 For Component: Azure Functions
-→ Invoke: azure-functions skill
-→ Extract: trigger patterns, bindings, hosting options
+→ Load: services/functions/templates/selection.md (decision tree)
```
tmeschter: Should this be a markdown reference?

paulyuk (commented):
Good catch! This is inside a code block showing an example pattern, so it doesn't render as a clickable link. However, you're right that agents reading this literally won't be able to follow the path.

I'll update this to use actual markdown links outside the code block, since the agent needs to be able to load these files.

saikoumudi (approved):

PR #1022, opened 2026-02-24
Schedule nightly integration test runs

JasonYeMSFT:
Schedule nightly integration test runs in the following way:
1. 9:00 PM PST, run microsoft-foundry integration tests
2. 12:00 AM PST, run azure-deploy integration tests
3. 4:00 AM PST, run integration tests for all other skills

A new skills.json configuration file is added to tell each schedule cron job which skills to test when it runs. A PR check step is added to remind people who add/remove/rename skills to update the configuration to ensure test coverage.

### DIFF
--- a/.github/workflows/test-all-integration.yml
+++ b/.github/workflows/test-all-integration.yml
@@ -46,6 +46,7 @@ jobs:
   resolve-inputs:
     name: Resolve workflow inputs
     runs-on: ubuntu-latest
+    if: github.repository == 'microsoft/GitHub-Copilot-for-Azure'
     outputs:
       skills: ${{ steps.resolve.outputs.skills }}
       deploy-test-pattern: ${{ steps.resolve.outputs.deploy-test-pattern }}


tests/jest.config.ts
```diff
@@ -57,9 +57,6 @@ const config: Config = {
   // Verbose output for better readability
   verbose: true,
 
-  // Fail fast in CI
-  bail: process.env.CI ? 1 : 0,
```
JasonYeMSFT, line 61: bail out means stop executing remaining tests as soon as seeing one failed test suite. This is not the behavior we want.

.github/workflows/test-all-integration.yml
```diff
@@ -27,9 +34,79 @@ on:
...
+env:  
+  SCHEDULED_DEBUG_21: 'false'        # debug flag for 9:00 PM PST run
+  SCHEDULED_DEBUG_00: 'false'        # debug flag for 12:00 AM PST run
+  SCHEDULED_DEBUG_04: 'false'         # debug flag for 4:00 AM PST run
+
 jobs:
+  resolve-inputs:
+    name: Resolve workflow inputs
+    runs-on: ubuntu-latest
+    if: github.repository == 'microsoft/GitHub-Copilot-for-Azure'
```
JasonYeMSFT, line 49: This condition will make all attempts to run this workflow in a forked repo to skip this job and everything afterwards since they all depend on this job.

Anyone who wants to manually run this workflow in their forked repo can temporarily remove this condition. They can also remove the cron job at the meantime in case they need to enable manually running this workflow for an extended period of time but don't want to have the scheduled runs.

tests/skills.json
```diff
@@ -0,0 +1,29 @@
+{
+    "skills": [
```
kvenkatrajan, line 2: Assuming this has to be managed manually when a new skill is added/updated? Can we add this in agent.md to be updated here as well?

JasonYeMSFT: I intentionally left it out since this is easy to edit and won't be edited frequently. I don't want to distract the coding agent on such basic task by adding tokens to its context window.

kvenkatrajan: sure thing we should keep an eye out for the next new skill added

PR #1038, opened 2026-02-25
fix: convert SKILL.md descriptions from YAML folded scalars to inline

ronniegeraghty:
skills.sh cannot parse YAML multi-line scalar syntax (>-) in frontmatter description fields, causing descriptions to render as literal '>-' instead of the actual description text. 

This PR converts all 21 affected SKILL.md files in plugin/skills/ from >- folded scalars to inline double-quoted strings, preserving the exact description content.

image

[Link to skills.sh page](https://skills.sh/microsoft/github-copilot-for-azure)

### DIFF
--- a/plugin/skills/appinsights-instrumentation/SKILL.md
+++ b/plugin/skills/appinsights-instrumentation/SKILL.md
@@ -1,9 +1,6 @@
 ---
 name: appinsights-instrumentation
-description: >-
-  Guidance for instrumenting webapps with Azure Application Insights. Provides telemetry patterns, SDK setup, and configuration references.
-  USE FOR: how to instrument app, App Insights SDK, telemetry patterns, what is App Insights, Application Insights guidance, instrumentation examples, APM best practices.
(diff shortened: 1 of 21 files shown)


spboyer:
> @JasonYeMSFT, then we might also want to request that the skills.sh site do the same. [This is the repo for skills cli tool](https://github.com/vercel-labs/skills), but I can't find if vercel has a repo for the site as well, so this might be the best place to put a similar request.

Will approve this change, we made the change from "|" to ">-" as way to properly do the long descriptions, however there are other finidings for Claude/Opus models on some triggers and directional phrases we need to pivot on affecting the description length thus negating the need for long (>100 words). PRs incoming and approve here soon.

@tmeschter @kvenkatrajan

spboyer (approved):
LGTM — this is a prerequisite for #1041 (cross-model description compatibility) which builds on this change. The inline double-quoted string format is required because:

1. \>-\ folded scalars are incompatible with skills.sh
2. The cross-model analysis shows inline descriptions work better for fast pattern matching on Sonnet-class models

#1041 updates sensei and skill-authoring to recommend inline quoted strings and the \WHEN:\ trigger format going forward.

PR #1040, opened 2026-02-26
Add azure-compute skill

alex-thompson:
* Added the `azure-compute` skill to assist with VM vs VMSS, VM family selection, and pricing lookup

**Reference Guides for Recommendations**
- Added `references/vm-families.md`, a VM Family Guide mapping workloads to VM families and series, including a decision tree, trade-offs, and naming conventions.
- Added `references/vmss-guide.md`, a VMSS Guide detailing when to recommend VMSS, orchestration modes (Flexible vs Uniform), autoscale patterns, networking options, cost estimation tips, and VMSS limits.
- Added `references/retail-prices-api.md`, a guide to querying the Azure Retail Prices API for VM pricing, including example queries, filter options, response fields, pagination, and tips for cost estimation.

### DIFF
--- a/plugin/skills/azure-compute/references/retail-prices-api.md
+++ b/plugin/skills/azure-compute/references/retail-prices-api.md
@@ -15,10 +15,22 @@ https://prices.azure.com/api/retail/prices?api-version=2023-01-01-preview
 
 ## Querying VM Prices
 
+> **No Azure CLI command exists** for the Retail Prices API. Since the API is unauthenticated, use `curl` (bash) or `Invoke-RestMethod` (PowerShell) directly. The `az rest` command also works but adds no auth benefit.
+
 ### Basic VM price lookup
 
 ```http
 GET https://prices.azure.com/api/retail/prices?$filter=serviceName eq 'Virtual Machines' and armRegionName eq 'eastus' and armSkuName eq 'Standard_D4s_v5' and priceType eq 'Consumption'
+```
+
+```bash
+curl -s "https://prices.azure.com/api/retail/prices?\$filter=serviceName%20eq%20'Virtual%20Machines'%20and%20armRegionName%20eq%20'eastus'%20and%20armSkuName%20eq%20'Standard_D4s_v5'%20and%20priceType%20eq%20'Consumption'"
+```
+
+```powershell
+$filter = "serviceName eq 'Virtual Machines' and armRegionName eq 'eastus' and armSkuName eq 'Standard_D4s_v5' and priceType eq 'Consumption'"
+$response = Invoke-RestMethod "https://prices.azure.com/api/retail/prices?`$filter=$filter"
+$response.Items | Select-Object armSkuName, retailPrice, unitOfMeasure, meterName
 ```
 
 ### Filter by family (all D-series in a region)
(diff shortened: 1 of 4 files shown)


tests/azure-compute/unit.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    describe("Recommender Reference Files", () => {
+        let vmFamiliesContent: string;
+        let retailPricesApiContent: string;
+        let vmssGuideContent: string;
+
+        beforeAll(async () => {
+            const fs = await import("fs/promises");
+            const path = await import("path");
+            const refsDir = path.join(
+                SKILLS_PATH,
```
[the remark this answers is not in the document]
alex-thompson: I believe this is global?

tests/azure-compute/recommender/integration.test.ts
```diff
@@ -0,0 +1,147 @@
...
+const skipTests = shouldSkipIntegrationTests();
+const skipReason = getIntegrationSkipReason();
+
+if (skipTests && skipReason) {
+    console.log(`⏭️  Skipping integration tests: ${skipReason}`);
+}
+
+const describeIntegration = skipTests ? describe.skip : describe;
+
+describeIntegration(`${SKILL_NAME}/recommender_ - Integration Tests`, () => {
```
JasonYeMSFT: Your test name should follow this pattern 
```
${SKILL_NAME}_{some-suffix} - Integration Tests
```

We only allow using dash in the suffix.

plugin/skills/azure-compute/recommender/SKILL.md
```diff
@@ -0,0 +1,158 @@
+---
```
JasonYeMSFT: Please don't write nested SKILL.md. They won't be loaded by the agent as skills. Rename them into something that are clearly references.

plugin/skills/azure-compute/SKILL.md
```diff
@@ -0,0 +1,61 @@
+---
+name: azure-compute
+description: >-
```
JasonYeMSFT: We temporarily decided to stop using this pattern for skill descriptions. See https://github.com/microsoft/GitHub-Copilot-for-Azure/pull/1038

tests/skills.json
JasonYeMSFT: You need to add the skill to one of the scheduled test groups. I suggest adding the the 0 12 * * * one unless your test consumes a lot of resources.

tests/azure-compute/recommender/integration.test.ts
JasonYeMSFT: And you don't have to write separate integration.test.ts files. You can write all the tests in the top level integration.test.ts file.

JasonYeMSFT (changes requested):
Please see my comments at places I would like you to make changes.

plugin/skills/azure-compute/recommender/SKILL.md
alex-thompson: @JasonYeMSFT I reverted this and everything is back under the single `SKILL.md`. If we want to add more skill-like workflows/instructions that aren't really references in the future, would this be an acceptable pattern? https://github.com/anthropics/skills/tree/main/skills/skill-creator/agents

JasonYeMSFT: The agent files you linked are also reference files. All the files other than the one SKILL.md at the root of the skill directory are "reference" files. They are named so because the only way they are supposed to be used by the agent is to be dynamically loaded by directly/indirectly following a reference (e.g. markdown link) in the SKILL.md.

If you want some content to not be used like that, you need to think about whether you want to make it a new top level SKILL.md or add it somewhere else, such as Agent.md, claude.md, copilot-instructions.md, etc.

plugin/skills/azure-compute/references/retail-prices-api.md
```diff
@@ -0,0 +1,86 @@
...
+```text
+https://prices.azure.com/api/retail/prices?api-version=2023-01-01-preview
+```
+
+## Querying VM Prices
+
+### Basic VM price lookup
+
+```http
+GET https://prices.azure.com/api/retail/prices?$filter=serviceName eq 'Virtual Machines' and armRegionName eq 'eastus' and armSkuName eq 'Standard_D4s_v5' and priceType eq 'Consumption'
```
JasonYeMSFT, line 23: Just wondering, is there a more user friendly way to call this API? Like Azure CLI?

tests/azure-compute/integration.test.ts
```diff
@@ -0,0 +1,218 @@
...
+            }
+
+            const invocationRate = successCount / RUNS_PER_PROMPT;
+            console.log(
+                `${SKILL_NAME} invocation rate for VM recommendation prompt: ${(invocationRate * 100).toFixed(1)}% (${successCount}/${RUNS_PER_PROMPT})`
+            );
+            fs.appendFileSync(
+                `./result-${SKILL_NAME}.txt`,
+                `${SKILL_NAME} invocation rate for VM recommendation prompt: ${(invocationRate * 100).toFixed(1)}% (${successCount}/${RUNS_PER_PROMPT})\n`
+            );
```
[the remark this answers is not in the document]
alex-thompson: Updating this to remove the writes

tests/azure-compute/integration.test.ts
```diff
@@ -0,0 +1,218 @@
...
+                    }
+                    throw e;
+                }
+            }
+
+            const invocationRate = successCount / RUNS_PER_PROMPT;
+            console.log(
+                `${SKILL_NAME} invocation rate for VM recommendation prompt: ${(invocationRate * 100).toFixed(1)}% (${successCount}/${RUNS_PER_PROMPT})`
+            );
+            fs.appendFileSync(
```
JasonYeMSFT: We have stopped using this approach to write skill-invocation results. Please use the softCheckSkill function from tests/utils/evaluate.ts for adding the invocation result into the final report.

tests/azure-compute/unit.test.ts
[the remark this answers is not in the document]
JasonYeMSFT: Don't use __dirname. It won't be available at runtime.

JasonYeMSFT (changes requested):
The skill content looks good. I'll approve this PR once you address the comments for test code.

plugin/skills/azure-compute/recommender/SKILL.md
alex-thompson: @JasonYeMSFT Sorry, should have been clearer: I mean having them under an `/agents` folder vs `/references`. The "sub-skills" could potentially be things like `recommender.md` or `troubleshooter.md`, but have no frontmatter; they could potentially refer to files under the `/references` folder.

plugin/skills/azure-compute/references/retail-prices-api.md
alex-thompson: Ah I had been picturing this as agent-only, but good point. Added additional examples using curl and PS

plugin/skills/azure-compute/recommender/SKILL.md
JasonYeMSFT: What you described is totally fine.

PR #1042, opened 2026-02-26
Use the MI service connection to query for key vaults

hallipr:
Prove agent-to-azure access by using the new `azure_msbench_eval` service connection to list the vaults in our test resource group.

### DIFF
--- a/pipelines/azure-benchmarks.yml
+++ b/pipelines/azure-benchmarks.yml
@@ -17,7 +17,7 @@ stages:
     - task: AzureCLI@2
       displayName: 'Run Copilot Benchmarks script'
       inputs:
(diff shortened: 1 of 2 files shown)


pipelines/azure-benchmarks.yml
```diff
@@ -14,7 +17,8 @@ stages:
     - task: AzureCLI@2
       displayName: 'Run Copilot Benchmarks script'
       inputs:
-        azureSubscription: 'Azure SDK Engineering System'
+        connectedServiceNameARM: 'azure-sdk_to_ghcp4a-testing'
```
fanyang-mono: Let's rename the MI to something like `msbench-CI-MI`

hallipr: updated to `azure_msbench_eval`

PR #1057, opened 2026-02-27
Add a script for validating frontmatter

tmeschter:
Currently we have some basic checks for skill frontmatter in pr.yml--specifically, we check that the frontmatter exists and includes "name" and "description" fields.

Here we convert these checks to a script that can be run both locally and in a GitHub Actions, and also adds further checks that the fields have appropriate length and formatting.

### DIFF
--- a/scripts/src/frontmatter/cli.ts
+++ b/scripts/src/frontmatter/cli.ts
@@ -20,6 +20,7 @@
 import { dirname, resolve, basename } from "node:path";
 import { fileURLToPath } from "node:url";
 import { existsSync, readdirSync, readFileSync, statSync } from "node:fs";
+import { parseSkillContent } from "../shared/parse-skill.js";
 
 // ── Paths ────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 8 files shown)


scripts/src/frontmatter/cli.ts
```diff
@@ -0,0 +1,347 @@
...
+
+  return content.substring(3, endIndex).trim();
+}
+
+/**
+ * Extract a simple YAML field value from raw frontmatter text.
+ * Handles inline double-quoted strings (with escaped quotes inside).
+ * Returns null if the field is not found.
+ */
+export function extractField(frontmatter: string, field: string): string | null {
```
JasonYeMSFT: FYI, there is an npm library called gray-matter that can parse frontmatter.
https://www.npmjs.com/package/gray-matter

tmeschter: I'm taking a look at that now.

scripts/src/frontmatter/cli.ts
```diff
@@ -0,0 +1,299 @@
...
+export function validateDescriptionFormat(rawFrontmatter: string): ValidationIssue[] {
+  const issues: ValidationIssue[] = [];
+
+  // Match description field followed by a block scalar indicator
+  if (/^description:\s*>-?\s*$/m.test(rawFrontmatter)) {
+    issues.push({
+      check: "description-format",
+      message: "Description uses >- folded scalar (incompatible with skills.sh) — use inline double-quoted string instead",
+    });
+  }
```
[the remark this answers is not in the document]
tmeschter: These aren't actual skills (even if they are named and structured that way) so the formatting requirement does not apply to them.

PR #1070, opened 2026-02-27
Update msbench script and pipeline

hallipr:


pipelines/azure-benchmarks.yml
```diff
@@ -14,6 +18,11 @@ stages:
   - job: RunBenchmark
     displayName: 'Run Copilot Benchmarks'
     steps:
+    - task: UsePythonVersion@0
+      displayName: 'Use Python 3.11'
```
fanyang-mono: I am using pythong 3.14 locally and it works well. Let's use that for CI as well.

pipelines/scripts/Invoke-CopilotBenchmarks.ps1
```diff
@@ -1,12 +1,123 @@
+<#
+.SYNOPSIS
+    Installs MSBench CLI in a local virtual environment and runs a Copilot Azure benchmark.
+
+.DESCRIPTION
+    This script is executed by the Azure DevOps benchmark pipeline to run a single Azure benchmark
```
fanyang-mono: It should run all Azure benchmark scenarios.

pipelines/scripts/Invoke-CopilotBenchmarks.ps1
```diff
@@ -1,12 +1,123 @@
...
+    instance using the github-copilot-cli agent.
+
+    The script creates a Python virtual environment in the working directory, installs MSBench CLI
```
fanyang-mono: We don't need to create python virtual environment for CI runs.

pipelines/scripts/Invoke-CopilotBenchmarks.ps1
```diff
@@ -1,12 +1,123 @@
...
+    Benchmark identifier
+
+.PARAMETER Model
+    Model name passed to msbench-cli via --model.
+    Default: claude-sonnet-4.5-autodev-test
+
+.PARAMETER NoWait
+    Whether to add --no-wait to the run command.
+    Accepted values: "true" or "false" (case-insensitive).
+    Default: true
```
fanyang-mono: Let's make it default to false. Once we make this works. I would like to add another step after the run completes to generate reports for the runs.

pipelines/scripts/Invoke-CopilotBenchmarks.ps1
```diff
@@ -1,12 +1,123 @@
...
+    Required environment variable:
+    - GITHUB_MCP_SERVER_TOKEN
+
+    MSBench CLI reference:
+    - https://github.com/devdiv-microsoft/MicrosoftSweBench/wiki
+
+.PARAMETER Benchmark
+    Benchmark identifier
+
+.PARAMETER Model
```
fanyang-mono, line 18: We probably don't need this. My plan is to add multiple runs against multiple models within this script, after we confirm one successful run.

pipelines/scripts/Invoke-CopilotBenchmarks.ps1
```diff
@@ -1,12 +1,123 @@
...
+python -m pip install --upgrade pip
+
+Write-Host "Installing artifact authentication dependencies"
+python -m pip install keyring artifacts-keyring
+
+Write-Host "Checking MSBench CLI versions from feed"
+python -m pip index versions msbench-cli --index-url $indexUrl
+
+Write-Host "Installing/upgrading MSBench CLI"
+python -m pip install --upgrade msbench-cli --index-url $indexUrl
```
fanyang-mono: I don't think we need the `--upgrade` flag here, since it will install the latest version at that time.

pipelines/azure-benchmarks.yml
```diff
@@ -1,4 +1,8 @@
-trigger: none
+trigger:
+  branches:
+    include:
+    - main
```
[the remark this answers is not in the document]
hallipr: Yaml supports block sequences with the dash at the same indention level as the block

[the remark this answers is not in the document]
hallipr: https://yaml.org/spec/1.2.2/

```yaml
american:
- Boston Red Sox
- Detroit Tigers
- New York Yankees
national:
- New York Mets
- Chicago Cubs
- Atlanta Braves
```

pipelines/scripts/Invoke-CopilotBenchmarks.ps1
hallipr: The script is meant to support local devs also (another reason we use powershell)

hallipr: but, yeah... I'll remove it.  Local users can just kill their venv if they want a clean, latest install

PR #1081, opened 2026-02-28
Fix broken auto-create evaluators step in deploy/observe loop

XOEEst:
The 'Auto-create evaluators & evaluation dataset' step was being skipped when the monolithic agent-observability-loop skill was split into separate deploy and observe skills. Neither skill owned the auto-create step, causing post-deploy users to jump directly to evaluation.

Changes:
- deploy.md: Replace generic 'set up evaluation?' prompt with automatic 6-step evaluator & dataset creation matching the reference behavior
- observe.md: Add Loop Overview, fix entry points to route post-deploy users through auto-setup, add evaluator existence check
- deploy-and-setup.md: Make auto-create primary content, demote deploy section to prerequisites


XOEEst:
wrong destination..

PR #1111, opened 2026-03-03
support for standard setup for agents and refactoring skill definitio…

smartmuki:
…n files so they align with Anthropic's recommendations


plugin/skills/microsoft-foundry/references/private-network-standard-agent-setup.md
```diff
@@ -0,0 +1,74 @@
+# Private Network Standard Agent Setup
```
kvenkatrajan, line 1: Is any of this content already in learn? If so we should pull from there. IIf not do we have a mechanism for refreshing this content?

plugin/skills/microsoft-foundry/references/standard-agent-setup.md
```diff
@@ -0,0 +1,51 @@
+# Standard Agent Setup
```
kvenkatrajan, line 1: Same comment applies here on refresh/mslearn

smartmuki: Yes, there is and I have added that in line 3 to read mslearn before proceeding. The other content are things added based on failures or things which are not obvious from mslearn.

plugin/skills/microsoft-foundry/references/private-network-standard-agent-setup.md
smartmuki: i think i can reduce this file further

PR #1112, opened 2026-03-03
Try using UV for faster installs

hallipr:


hallipr:
@copilot, can you iterate on this until it can install and run via UV?

PR #1137, opened 2026-03-05
Feature/azure quotas

rakal-dyh:
In progress. Test need to be added


JasonYeMSFT:
I merged a change to reduce the char count of other skills' descriptions. Please rebase to the latest main and check the total char count again.

PR #1215, opened 2026-03-10
VM Troubleshooting Skill

rmmue21:


plugin/skills/azure-compute-vm-troubleshooting-connectivity/references/cannot-connect-to-vm.md
```diff
@@ -0,0 +1,315 @@
+# Cannot Connect to VM
+
+When to use this reference file: when the user is facing connectivity issues with their Azure VM, such as:
```
JasonYeMSFT, line 3: Can you add these as an additional reference to azure-compute skill instead?

PR #1218, opened 2026-03-11
Feature/azure quotas update

rakal-dyh:
Some improvements to azure-quotas
1. reduce some of the text
2. add instruction that not trig this quotas when from microsoft-foundary

### DIFF
--- a/plugin/skills/azure-quotas/SKILL.md
+++ b/plugin/skills/azure-quotas/SKILL.md
@@ -38,20 +38,6 @@ Invoke this skill when:
 - **Comparing regional capacity** - Find regions with available quota
 - **Validating provisioning limits** - Ensure deployment won't exceed quotas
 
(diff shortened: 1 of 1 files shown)


plugin/skills/azure-quotas/SKILL.md
```diff
@@ -38,6 +38,20 @@ Invoke this skill when:
 - **Comparing regional capacity** - Find regions with available quota
 - **Validating provisioning limits** - Ensure deployment won't exceed quotas
 
+## ⛔ When NOT to Use This Skill
```
JasonYeMSFT: This piece of text won't prevent the agent from invoking this skill. The agent only uses the description in the frontmatter to determine if it should invoke this skill.

rakal-dyh: removed

PR #1298, opened 2026-03-16
Enable DYNAMIC_INSTRUCTIONS_RETRIEVAL

JasonYeMSFT:
Test runs with the feature enabled.
https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/23154686674
https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/23154773001

### DIFF
--- a/tests/utils/agent-runner.ts
+++ b/tests/utils/agent-runner.ts
@@ -516,7 +516,8 @@ export function useAgentRunner() {
         cliPath: getBundledCliPath(),
         env: {
           ...process.env,
-          SKILLS_INSTRUCTIONS: "true"
+          SKILLS_INSTRUCTIONS: "true",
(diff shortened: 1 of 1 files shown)


jongio (commented):
The change itself looks good - follows the existing env var pattern and CI is green.

One issue: this branch was forked before `SKILL_CHAR_BUDGET: "20000"` was added to the same env block on main. Merging will conflict. Rebase on main so the result includes all three env vars (`SKILLS_INSTRUCTIONS`, `DYNAMIC_INSTRUCTIONS_RETRIEVAL`, `SKILL_CHAR_BUDGET`).

JasonYeMSFT:
Copilot CLI has improved the default behavior of skill truncation such that truncated skills at least have their name listed in the prompt. We no longer urgently need this flag for testing. Closing.

PR #1324, opened 2026-03-17
Fixes for azure-resource-visualizer

### LINKED ISSUE #1266: Integration test failure: azure-resource-visualizer – resource connections diagram not created [Assertion mismatch]
## Prompt

```
Visualize how my Azure resources are connected and show their relationships
```

Follow-up: `Go with recommended options.`

## Summary

**Run:** [Integration Tests - all #84](https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/23050118229)
**Test:** `azure-resource-visualizer_ - Integration Tests › resource-group-visualization › visualizes resource connections and relationships`
**Result:** Fail
**Duration:** 4m 19s

## Root Cause Category

**Assertion mismatch**

### What was expected
The test expects `azure-resource-visualizer` to be invoked (`isSkillUsed = true`) and an architecture diagram file to be created in the workspace (`hasDiagramFile = true`).

### What actually happened
`isSkillUsed` passed (skill was invoked), but `expect(hasDiagramFile).toBe(true)` failed — `hasArchitectureDiagramFile(agentMetadata)` returned `false`. The skill ran but did not produce an architecture diagram file.

### Why it failed
The `azure-resource-visualizer` skill was triggered but failed to create the expected diagram file (likely a `.mmd`, `.dot`, or `.png` file). This may be because: (1) the skill generated a text-based diagram inline rather than saving to a file, (2) the file was saved under a different name/format than `hasArchitectureDiagramFile` checks, or (3) the resource visualization API call failed silently.

### Suggested fix
- Verify `azure-resource-visualizer` saves the diagram to a file and not just outputs it inline
- Update `hasArchitectureDiagramFile` to match the actual output format/filename convention used by the skill
- Add error handling to the skill for cases where Azure resource graph queries fail or return empty

## azure-resource-visualizer Skill Invocation

| Skill | Invoked |
|-------|---------|
| **azure-resource-visualizer** | **Yes** |

## Skill Report Context

<details>
<summary>Per-test section from SKILL-REPORT.md (click to expand)</summary>

_Skill report artifact was not accessible in this analysis run. See run artifacts at https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/23050118229/artifacts/5911223040_

</details>

## Environment

- **Node.js:** v24.14.0
- **Model:** claude-sonnet-4.5 (or as overridden)
- **Run URL:** https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/23050118229
- **Commit:** `fef6de1232d9a1725537d8058e39e382e78233ab`

> Generated by [Analyze Test Run](https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/23055952855) · [◷](https://github.com/search?q=repo%3Amicrosoft%2FGitHub-Copilot-for-Azure+is%3Aissue+%22gh-aw-workflow-call-id%3A+microsoft%2FGitHub-Copilot-for-Azure%2Fanalyze-test-run%22&type=issues)

tmeschter:
Fixes #1266.

1. Update the skill `description` to use "WHEN" instead of "USE FOR", and remove the "DO NOT USE FOR" section.
2. Update the integration tests to more reliably check for a generated architecture.*.md file with a MermaidJS diagram. Previously we were dependent on Copilot using the `create` tool to create the file, but in some runs we've seen it use a bash script instead. In these cases the file would be created but we would assert that it hadn't. Looking for the actual file rather than depending on the manner of its creation should be more reliable.

### DIFF
--- a/tests/azure-resource-visualizer/integration.test.ts
+++ b/tests/azure-resource-visualizer/integration.test.ts
@@ -11,14 +11,12 @@
  * 3. az login (for Azure resource access)
  */
 
-import * as fs from "fs";
-import * as path from "path";
 import {
   useAgentRunner,
   shouldSkipIntegrationTests,
   getIntegrationSkipReason
 } from "../utils/agent-runner";
-import { softCheckSkill, isSkillInvoked } from "../utils/evaluate";
+import { softCheckSkill, isSkillInvoked, doesWorkspaceFileIncludePattern } from "../utils/evaluate";
 
 const SKILL_NAME = "azure-resource-visualizer";
 const RUNS_PER_PROMPT = 5;
@@ -34,19 +32,6 @@ if (skipTests && skipReason) {
 
 const describeIntegration = skipTests ? describe.skip : describe;
 const visualizerTestTimeoutMs = 1800000;
-
-/**
- * Check if the workspace contains an architecture markdown file
- * with a Mermaid diagram (graph TB or graph LR).
- */
-function hasArchitectureDiagramFile(workspacePath: string): boolean {
-  const files = fs.readdirSync(workspacePath, { recursive: true, encoding: "utf-8" });
-  const architectureFiles = files.filter(f => /architecture.*\.md$/i.test(f));
-  return architectureFiles.some(file => {
-    const content = fs.readFileSync(path.join(workspacePath, file), "utf-8");
-    return /graph\s+(TB|LR)/i.test(content);
-  });
-}
 
 describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
   const agent = useAgentRunner();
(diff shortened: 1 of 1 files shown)


tests/azure-resource-visualizer/integration.test.ts
```diff
@@ -97,28 +82,40 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
...
+        followUp: FOLLOW_UP_PROMPT,
+        setup: async (workspace: string) => {
+          workspacePath = workspace;
+        }
       });
 
       const isSkillUsed = isSkillInvoked(agentMetadata, SKILL_NAME);
-      const hasDiagramFile = hasArchitectureDiagramFile(agentMetadata);
+      expect(workspacePath).toBeDefined();
+      const hasDiagramFile = doesWorkspaceFileIncludePattern(workspacePath!, /graph\s+(TB|LR)/i, /architecture.*\.md$/i);
```
JasonYeMSFT, line 98: Ideally we check both files and responses and pass the test if the diagram is in either of them.

PR #1331, opened 2026-03-17
fix: Add guidance on secret storage

tmeschter:
Fixes https://github.com/microsoft/GitHub-Copilot-for-Azure/issues/1323.

When an Aspire app deploys to Azure Functions we need to ensure that the application is setup to access secret storage properly. Here we add the necessary guidance.

Without this, the app may appear to deploy but then fail to start properly. Using the Azure skills, Copilot is likely to recover from this issue but it would be better to proactively check for this and avoid problems to begin with.

### DIFF
--- a/plugin/skills/azure-validate/references/aspire-functions-secrets.md
+++ b/plugin/skills/azure-validate/references/aspire-functions-secrets.md
@@ -17,7 +17,7 @@ This check is required when **all** of these are true:
 Search for `AddAzureFunctionsProject` in the AppHost source file(s):
 
 ```bash
-grep -rn "AddAzureFunctionsProject" . --include="*.cs" -l
+grep -rn "AddAzureFunctionsProject" . --include="*.cs"
 ```
 
 **PowerShell:**


plugin/skills/azure-validate/references/aspire-functions-secrets.md
```diff
@@ -0,0 +1,78 @@
...
+Search for `AddAzureFunctionsProject` in the AppHost source file(s):
+
+```bash
+grep -rn "AddAzureFunctionsProject" . --include="*.cs" -l
+```
+
+If found, check whether `AzureWebJobsSecretStorageType` is already configured in those same file(s):
+
+```bash
+# Check only the AppHost file(s) that contain AddAzureFunctionsProject
```
wbreza, line 31: **[High]** Cross-platform gap: Detection commands are bash-only

The detection commands here (`find`, `xargs`, `grep`) are bash-only, but the post-provisioning section in `aspire.md` correctly provides both bash **and** PowerShell. This is inconsistent within the PR. Repo standards state: *"Provide both bash and PowerShell versions for non-trivial scripts."*

**Suggested Fix:**
```suggestion
# Check only the AppHost file(s) that contain AddAzureFunctionsProject
# Bash:
grep -l "AddAzureFunctionsProject" $(find . -name "*.cs" -path "*AppHost*") | xargs grep -l "AzureWebJobsSecretStorageType"
```

Consider adding a PowerShell equivalent block below:
```powershell
Get-ChildItem -Recurse -Filter "*.cs" |
  Where-Object { $_.FullName -match "AppHost" } |
  Select-String "AddAzureFunctionsProject" -List |
  ForEach-Object { Select-String "AzureWebJobsSecretStorageType" -Path $_.Path }
```

plugin/skills/azure-validate/references/recipes/azd/aspire.md
```diff
@@ -0,0 +1,87 @@
...
+grep -l "AddAzureFunctionsProject" $(find . -name "*.cs" -path "*AppHost*") | xargs grep -l "AzureWebJobsSecretStorageType"
+```
+
+**If `AddAzureFunctionsProject` is present but `AzureWebJobsSecretStorageType` is NOT configured in the same file**, edit the AppHost source file that contains the `AddAzureFunctionsProject` call (often `Program.cs` in the `*.AppHost` project) to add `.WithEnvironment("AzureWebJobsSecretStorageType", "Files")` to the Functions project builder chain:
+
+```csharp
+var functions = builder.AddAzureFunctionsProject<Projects.MyFunctions>("functions")
+    .WithHostStorage(storage)
+    .WithEnvironment("AzureWebJobsSecretStorageType", "Files")  // Required for Container Apps
+    // ... other configuration
```
wbreza: **[High]** DRY: Detection logic and fix example duplicated

The detection commands (lines 25-33) and fix example (lines 35-42) are duplicated nearly verbatim from `aspire-functions-secrets.md`. While the link to the reference doc on line 45 is great progressive disclosure, the repeated content creates a maintenance burden — future changes must be synchronized in two places.

**Suggested Fix:** Consider making this section a brief summary with a direct link, e.g.:

> Check if the project uses Azure Functions within Aspire and ensure `AzureWebJobsSecretStorageType` is configured.
> See [Aspire Functions Secrets Reference](../../aspire-functions-secrets.md) for detection commands, fix examples, and full details.

plugin/skills/azure-validate/SKILL.md
```diff
@@ -4,7 +4,7 @@ description: "Pre-deployment validation for Azure readiness. Run deep checks on
 license: MIT
 metadata:
   author: Microsoft
-  version: "1.0.0"
+  version: "1.0.3"
```
wbreza: **[Medium]** Version bump skips 1.0.1 and 1.0.2

Version jumps from `1.0.0` → `1.0.3`. Were intermediate versions released elsewhere? If not, consider using `1.0.1` (next sequential patch) or adding a note in the PR description explaining the gap.

plugin/skills/azure-validate/references/recipes/azd/aspire.md
```diff
@@ -0,0 +1,87 @@
...
+Check if the project uses Azure Functions within Aspire by searching the AppHost source file(s):
+
+```bash
+grep -rn "AddAzureFunctionsProject" . --include="*.cs" -l
+```
+
+**If found**, check whether `AzureWebJobsSecretStorageType` is already configured in those same file(s):
+
+```bash
+# Check only the AppHost file(s) that contain AddAzureFunctionsProject
```
wbreza: **[Medium]** Detection command fragile with spaces in paths

This `find | xargs` pipeline will break on paths containing spaces. The `find` output is word-split in the command substitution, and `xargs` splits on whitespace by default.

**Suggested Fix:**
```suggestion
find . -name "*.cs" -path "*AppHost*" -print0 | xargs -0 grep -l "AddAzureFunctionsProject" 2>/dev/null | xargs grep -l "AzureWebJobsSecretStorageType"
```

PR #1336, opened 2026-03-17
Add PR check for undesired file change

fanyang-mono:
Add PR check to see if plugin version or CHANGELOG.md were modified. They are supposed to be modified automatically.

### DIFF
--- a/.github/workflows/pr-plugin-version-check.yml
+++ b/.github/workflows/pr-plugin-version-check.yml
@@ -39,8 +39,6 @@ jobs:
           echo "Checking if plugin versions were modified in this PR..."
           cd scripts && npm run checkPluginVersionPr
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          PR_NUMBER: ${{ github.event.number }}
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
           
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-plugin-version-check.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          cd scripts
+          npm ci
+
+      - name: Check plugin version changes
+        run: |
+          echo "Checking if plugin versions were modified in this PR..."
+          cd scripts && npm run checkPluginVersionPr
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
JasonYeMSFT: The token and the pr number aren't used.

.github/workflows/pr-plugin-version-check.yml
```diff
@@ -0,0 +1,55 @@
...
+  pull-requests: read
+
+jobs:
+  check-plugin-modifications:
+    runs-on: ubuntu-latest
+    name: Check Plugin Version and CHANGELOG Not Modified
+    
+    steps:
+      - name: Checkout PR head
+        uses: actions/checkout@v6
```
JasonYeMSFT, line 22: Can you lock each step with its hash? Like what this PR does. https://github.com/microsoft/GitHub-Copilot-for-Azure/pull/1329

JasonYeMSFT: A follow up PR is fine.

.github/workflows/pr-plugin-version-check.yml
fanyang-mono: They have been removed.

.github/workflows/pr-plugin-version-check.yml
fanyang-mono: Will do in a follow-up PR.

PR #1522, opened 2026-03-26
update test prompts to better reflect real life cases

tendau:
Update tests to better reflect real life invocation scenarios and lower run amount to speed up test runs.

### DIFF
--- a/plugin/skills/microsoft-foundry/SKILL.md
+++ b/plugin/skills/microsoft-foundry/SKILL.md
@@ -1,10 +1,10 @@
 ---
 name: microsoft-foundry
-description: "Deploy, evaluate, and manage Foundry agents end-to-end: Docker build, ACR push, hosted/prompt agent create, container start, batch eval, prompt optimization, prompt optimizer workflows, agent.yaml, dataset curation from traces. USE FOR: deploy agent to Foundry, hosted agent, create agent, invoke agent, evaluate agent, run batch eval, optimize prompt, improve prompt, prompt optimization, prompt optimizer, improve agent instructions, optimize agent instructions, optimize system prompt, deploy model, Foundry project, RBAC, role assignment, permissions, quota, capacity, region, troubleshoot agent, deployment failure, create dataset from traces, dataset versioning, eval trending, create AI Services, Cognitive Services, create Foundry resource, provision resource, knowledge index, agent monitoring, customize deployment, onboard, availability. DO NOT USE FOR: Azure Functions, App Service, general Azure deploy (use azure-deploy), general Azure prep (use azure-prepare)."
+description: "Microsoft Foundry agent skill — send messages to Foundry agents, invoke, deploy, evaluate, and manage agents in Azure AI Foundry. USE FOR: send test message to Foundry agent, invoke Foundry agent, chat with Foundry agent, deploy agent to Foundry, create agent, evaluate agent, run batch eval, optimize prompt, improve agent instructions, deploy model, Foundry project, RBAC, role assignment, permissions, quota, capacity, region, troubleshoot agent, deployment failure, create dataset from traces, dataset versioning, eval trending, create AI Services, Cognitive Services, create Foundry resource, knowledge index, agent monitoring, onboard, availability. DO NOT USE FOR: Azure Functions, App Service, general Azure deploy (use azure-deploy), general Azure prep (use azure-prepare)."
 license: MIT
 metadata:
   author: Microsoft
-  version: "1.0.7"
+  version: "1.0.8"
 ---
 
 # Microsoft Foundry Skill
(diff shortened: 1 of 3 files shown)


tests/microsoft-foundry/foundry-agent/create/integration.test.ts
```diff
@@ -18,6 +18,9 @@ import {
 import { softCheckSkill, isSkillInvoked, getToolCalls, withTestResult } from "../../../utils/evaluate";
 
 const SKILL_NAME = "microsoft-foundry";
+// These prompts trigger heavy agent work (file creation, scaffolding).
+// 3 runs keeps the test within the 20-minute Jest timeout while still
```
jongio, line 22: The comment says `3 runs` but `RUNS_PER_PROMPT` is still 5. Looks like either the comment should say `5 runs` or the value was supposed to be changed to 3 and wasn't. If the intent was to lower it to 3, note that `ceil(3 * 0.8) = 3` - so a 0.8 threshold with 3 runs effectively requires 100% success rate.

tests/microsoft-foundry/foundry-agent/invoke/integration.test.ts
```diff
@@ -11,9 +11,11 @@ import {
   shouldSkipIntegrationTests,
   getIntegrationSkipReason,
 } from "../../../utils/agent-runner";
-import { isSkillInvoked, withTestResult } from "../../../utils/evaluate";
+import { softCheckSkill, isSkillInvoked, shouldEarlyTerminateForSkillInvocation, withTestResult } from "../../../utils/evaluate";
 
 const SKILL_NAME = "microsoft-foundry";
+const RUNS_PER_PROMPT = 5;
+const invocationRateThreshold = 0.8;
```
jongio, line 19: The invoke test previously ran the prompt once. It now runs 5 times with rate tracking, which is a good reliability pattern - but it conflicts with the PR description's goal to `lower run amount to speed up test runs.` This test's runtime increased ~5x. If speed is a goal, consider using 3 runs here (and updating the threshold accordingly).

plugin/skills/microsoft-foundry/SKILL.md
```diff
@@ -1,10 +1,10 @@
...
+description: "Microsoft Foundry agent skill — send messages to Foundry agents, invoke, deploy, evaluate, and manage agents in Azure AI Foundry. USE FOR: send test message to Foundry agent, invoke Foundry agent, chat with Foundry agent, deploy agent to Foundry, create agent, evaluate agent, run batch eval, optimize prompt, improve agent instructions, deploy model, Foundry project, RBAC, role assignment, permissions, quota, capacity, region, troubleshoot agent, deployment failure, create dataset from traces, dataset versioning, eval trending, create AI Services, Cognitive Services, create Foundry resource, knowledge index, agent monitoring, onboard, availability. DO NOT USE FOR: Azure Functions, App Service, general Azure deploy (use azure-deploy), general Azure prep (use azure-prepare)."
```
jongio, line 3: Several trigger phrases were removed from the description: `hosted agent`, `improve prompt`, `prompt optimization`, `prompt optimizer`, `optimize agent instructions`, `optimize system prompt`, `provision resource`, `customize deployment`. Were these intentional removals? Some of these (especially the prompt optimization variants) seem like they'd still be valid user queries that should route to this skill.

jongio (changes requested):
The prompt improvements are solid - the new invoke prompt is more realistic and specific. But there are a few things that need fixing before this lands.

CI is failing - 26 snapshot tests across 13 suites are broken because of the SKILL.md description change. You'll need to update the snapshots (`npm run npx -- -u` per the CI output).

plugin/skills/microsoft-foundry/SKILL.md
JasonYeMSFT, line 3: I don't see it necessary to repeat the skill name in description. When agent presents the skill to the LLM, the name is already there as a part of the structured data, like
```xml
<name>microsoft-foundry</name>
<description>the_description_text</description>
```

JasonYeMSFT:
@tendau The change to the prompt looks good to me. Could you please resolve the remaining comments, resolve the merge conoflicts and address the failed ci check.

plugin/skills/microsoft-foundry/SKILL.md
```diff
@@ -1,10 +1,10 @@
...
 license: MIT
 metadata:
   author: Microsoft
-  version: "1.0.7"
+  version: "1.0.8"
```
jongio, line 7: Per repo convention (`skill-files.instructions.md`), skills under `plugin/` should use `"0.0.0-placeholder"` for the version field since NBGV stamps the real version at build time. Other skills (`azure-deploy`, `azure-prepare`, etc.) follow this pattern. Consider switching to the placeholder instead of manually bumping.

jongio (changes requested):
Three things still need addressing before this can merge:

1. **Merge conflicts** need resolution (base branch has diverged).
2. **CI is failing** on snapshot tests due to the SKILL.md description change (noted by prior reviewers, still unresolved).
3. **Unresolved inline comments** from prior reviews about the runs/comment mismatch and removed trigger phrases.

Additionally, the manual version bump (1.0.7 to 1.0.8) doesn't match repo convention. Other skills under `plugin/` use `"0.0.0-placeholder"` in source because NBGV stamps the real version at build time. See `.github/instructions/skill-files.instructions.md`.

PR #1683, opened 2026-04-02
Refactor VM Troubleshooting for Tokens

rmmue21:
## Checklist

- [x] **If modifying skill descriptions:** verified routing correctness with integration tests (`npm run test:skills:integration -- <skill>`)
- [x] **If modifying skill `USE FOR` / `DO NOT USE FOR` / `PREFER OVER` clauses:** confirmed no routing regressions for competing skills
- [x] Version bumped in skill frontmatter (if skill files changed)

### DIFF
--- a/tests/azure-compute/triggers.test.ts
+++ b/tests/azure-compute/triggers.test.ts
@@ -85,20 +85,138 @@ describe(`${SKILL_NAME} - Trigger Tests`, () => {
     );
   });
 
-  describe("Should Trigger - VM Troubleshooting", () => {
-    const vmTroubleshootingPrompts: string[] = [
+  describe("Should Trigger - VM Troubleshooting: Unable to RDP", () => {
+    const rdpPrompts: string[] = [
       "I can't connect to my Azure VM via RDP",
-      "SSH connection to my Azure VM is refused",
-      "My Azure VM is unreachable after updating NSG rules",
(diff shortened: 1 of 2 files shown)


plugin/skills/azure-compute/workflows/vm-troubleshooter/vm-troubleshooter.md
```diff
@@ -52,9 +52,19 @@ If the user shares an Azure VM name or resource ID, attempt to use the azure-res
 
 ### Phase 2: Route to Solution
 
-Open [references/cannot-connect-to-vm.md](references/cannot-connect-to-vm.md) and find the **Symptoms → Solutions** table that matches the user's category. Narrow down to the specific row matching their symptom.
-
-If additional details are needed to narrow to a specific solution row, ask the user. For example:
+Open [references/cannot-connect-to-vm.md](references/cannot-connect-to-vm.md) to identify the symptom category. Then open the matching sub-reference for the full **Symptoms → Solutions** table and Quick Commands:
```
alex-thompson: There's some duplication here. I think we can keep the original example prompts that lead to `cannot-connect-to-vm.md` and then let it handle the routing to the other references.

rmmue21: fixed

PR #1747, opened 2026-04-07
Update test as deployment should not succeed

### LINKED ISSUE #1743: Integration test failure: azure-deploy – aspire health-checks-ui brownfield dotnet [Assertion mismatch]
## Prompt

```
Please deploy this application to Azure. Use the eastus2 region. Use my current subscription. This is for a small scale production environment. Use standard SKUs. The app can be found under samples/health-checks-ui.
```

## Summary

**Run:** [Integration Tests - all #143](https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/24071855639)  
**Test:** `azure-deploy - Integration Tests › brownfield-dotnet > deploys aspire health-checks-ui`  
**Result:** Fail  
**Duration:** 6m 50s

## Root Cause Category

**Assertion mismatch**

### What was expected
The test asserts `containsDeployLinks` is `true` — the assistant output should contain at least one Azure deployment URL after completing the deployment workflow.

### What actually happened
`azure-validate` correctly identified that the application contains a custom Aspire resource type (`HealthChecksUIResource`) that cannot be deployed to Azure. It blocked deployment with error `"generating bicep from manifest: unsupported resource type"`. `azure-deploy` was never invoked (correctly prevented by validation). As a result, no deployment links were produced.

### Why it failed
The test unconditionally expects deploy links, but the application intentionally uses a non-deployable Aspire custom resource type. `azure-validate` doing the right thing (blocking an unsupported deployment) causes the test to fail. The test should either detect the expected validation block and pass, or this sample should be excluded from the `containsDeployLinks` assertion.

### Suggested fix
- Update test to handle the expected `unsupported resource type` case from `azure-validate` and assert that the agent clearly communicates the blocker (rather than requiring deploy links)
- Or add a negative assertion variant: `brownfield-dotnet > does not deploy aspire health-checks-ui` mirroring the existing `does not deploy aspire custom resources` pattern

## azure-deploy Skill Invocation

| Skill | Invoked |
|-------|---------|
| **azure-deploy** | No (correctly blocked by azure-validate) |

> In azure-deploy integration tests, the full skill chain is azure-prepare → azure-validate → azure-deploy. In this test, azure-prepare and azure-validate ran successfully; azure-validate blocked azure-deploy because `HealthChecksUIResource` is an unsupported Aspire resource type.

## Skill Report Context

<details>
<summary>Per-test section from SKILL-REPORT.md (click to expand)</summary>

**Pass Rate:** 0% (Blocked) | **Avg Confidence:** 65%

The agent executed azure-prepare, azure-quotas, and azure-validate in sequence (177s total). Validation correctly identified that the application contains a custom Aspire resource type (`HealthChecksUIResource`) that cannot be deployed to Azure. The azure-validate skill blocked deployment with error `"generating bicep from manifest: unsupported resource type"`. azure-deploy was not invoked because validation prevented progression to deployment phase.

**What Went Well:** Successfully identified custom resource blocker before attempting provision; followed proper validation workflow.

**What Went Wrong:** Application uses custom resource type not supported by Azure deployment tooling; could detect custom resources earlier.

</details>

## Environment

- **Node.js:** (from workflow)
- **Model:** claude-sonnet-4.5
- **Run URL:** https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/24071855639
- **Commit:** d0479c2c4961c40ee103869a016e14040813fee5

> Generated by [Analyze Test Run](https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/24081616374/agentic_workflow) · ● 3.4M · [◷](https://github.com/search?q=repo%3Amicrosoft%2FGitHub-Copilot-for-Azure+is%3Aissue+%22gh-aw-workflow-call-id%3A+microsoft%2FGitHub-Copilot-for-Azure%2Fanalyze-test-run%22&type=issues)

fanyang-mono:
## Description
This app contains custom Aspire resource types that Azure Developer CLI cannot deploy to Azure. Update test expectations accordingly.

## Related Issues
Fixes #1743

### DIFF
--- a/tests/azure-deploy/integration.test.ts
+++ b/tests/azure-deploy/integration.test.ts
@@ -894,7 +894,7 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
       });
     }, brownfieldTestTimeoutMs);
 
-    test("deploys aspire health-checks-ui", async () => {
+    test("does not deploy aspire health-checks-ui", async () => {
       await withTestResult(async () => {
         const HEALTH_CHECKS_SPARSE_PATH = "samples/health-checks-ui";
 


tests/azure-deploy/integration.test.ts
```diff
@@ -922,7 +922,8 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
         softCheckDeploySkills(agentMetadata);
         const containsDeployLinks = hasDeployLinks(agentMetadata);
 
-        expect(containsDeployLinks).toBe(true);
+        // This app contains custom Aspire resource types that Azure Developer CLI cannot deploy to Azure.
```
kvenkatrajan, line 925: Is it supposed to generate custom Aspire resource types?

fanyang-mono: Yes, I looked at previous passed runs. Copilot usually modify the source code to remove those custom aspire resources

JasonYeMSFT: Can you copy the project locally, remove the custom aspire resources and give the trimmed project to setup for this test case? Alternatively, repurpose this test case to make sure the agent doesn't deploy such apps or remove this test case.

jongio, line 925: The assertion fix is right. One thing though - the test name at line 897 still says `"deploys aspire health-checks-ui"` which now contradicts this `false` assertion. The existing convention (line 833) uses `"does not deploy aspire custom resources"` for the same pattern.

Consider renaming to:
```typescript
test("does not deploy aspire health-checks-ui", async () => {
```

This keeps CI output consistent and makes the intent clear at a glance.

fanyang-mono: Updated test name to indicate that it is expected that deployment won't happen.

fanyang-mono: Good catch. Updated test name.

PR #1766, opened 2026-04-07
Fix azure cost rate limiting

taylorak:
Description

Addresses three issues with the azure-cost skill across 11 files:

1. Rate-Limit Retry Handling — AB#37441409

The skill had a 40% integration test failure rate caused by Cost Management API 429 responses. The agent was not properly waiting between requests,
violating the per-scope limit of 4 requests/minute.

Changes:

 - Added rate-limit threshold tables to cost-query/guardrails.md and cost-forecast/guardrails.md documenting all five tiers (per-user, per-scope, 
per-tenant, per-client-type)
 - Added 429 handling sections instructing the agent to read all three x-ms-ratelimit-microsoft.costmanagement-*-retry-after headers and wait for the 
longest value
 - Updated error-handling.md in both cost-query and cost-forecast to reference all three retry-after headers
 - Expanded rate-limit guidance in cost-query/workflow.md and cost-optimization/workflow.md
 - Added rate-limit best practice to SKILL.md
 - Reduced test RUNS_PER_PROMPT from 5→3 and invocationRateThreshold from
  0.8→0.6 to stay within per-scope limits

2. ClientType Request Header

Added --headers "ClientType=GitHubCopilotForAzure" to all az rest commands in cost-query, cost-forecast, and cost-optimization workflows. This enables
per-client-type rate-limit bucketing (2,000 req/min) and telemetry attribution for requests originating from GitHub Copilot.

3. Filter Schema Casing — AB#37400748

Fixed filter JSON property names from PascalCase → lowercase to match the official Query API docs (
https://learn.microsoft.com/en-us/rest/api/cost-management/query/usage?view=rest-cost-management-2023-11-01):

 - And→and, Or→or, Not→not
 - Dimensions→dimensions, Tags→tags
 - Name→name, Operator→operator, Values→values

Updated in request-body-schema.md, examples.md, and guardrails.md.

Checklist

 - [X]  Tests pass locally — 441 tests ✅
 - [ ]  If modifying skill descriptions: verified routing correctness
 - [ ]  If modifying USE FOR/DO NOT USE FOR/PREFER OVER: confirmed no routing regressions
 - [X]  Version bumped in skill frontmatter (1.0.0 →
  1.1.0)

Related Issues

 - Fixes AB#37441409 — azure-cost 40% failure rate due to rate-limit retry handling
 - Fixes AB#37400748 — Forecast filter schema casing mismatch causing API 400 errors
 - Related to #1626 — azure-cost-optimization integration test keyword assertion mismatch

### DIFF
--- a/tests/azure-cost/integration.test.ts
+++ b/tests/azure-cost/integration.test.ts
@@ -260,17 +260,6 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
         doesAssistantMessageIncludeKeyword(agentMetadata, "estimate");
       expect(hasForecast).toBe(true);
     }));
-
-    test("response addresses Cost Management 429 rate limiting", () => withTestResult(async () => {
-      const agentMetadata = await agent.run({
-        prompt: "I'm getting 429 rate limit errors from the Cost Management API, how do I handle them?"
-      });
-      const addresses429 = doesAssistantMessageIncludeKeyword(agentMetadata, "retry-after") ||
-        doesAssistantMessageIncludeKeyword(agentMetadata, "retry after") ||
-        doesAssistantMessageIncludeKeyword(agentMetadata, "rate limit") ||
-        doesAssistantMessageIncludeKeyword(agentMetadata, "429");
-      expect(addresses429).toBe(true);
-    }));
   });
 
   describe("aks-cost-optimization", () => {
--- a/tests/azure-cost/integration.test.ts
+++ b/tests/azure-cost/integration.test.ts
@@ -21,7 +21,7 @@ import { softCheckSkill, isSkillInvoked, withTestResult, shouldEarlyTerminateFor
 
 const SKILL_NAME = "azure-cost";
 const RUNS_PER_PROMPT = 3;
-const invocationRateThreshold = 0.6;
+const invocationRateThreshold = 0.8;
 
 const skipTests = shouldSkipIntegrationTests();
 const skipReason = getIntegrationSkipReason();
--- a/tests/azure-cost/integration.test.ts
+++ b/tests/azure-cost/integration.test.ts
@@ -20,8 +20,8 @@ import {
 import { softCheckSkill, isSkillInvoked, withTestResult, shouldEarlyTerminateForSkillInvocation } from "../utils/evaluate";
 
 const SKILL_NAME = "azure-cost";
-const RUNS_PER_PROMPT = 5;
-const invocationRateThreshold = 0.8;
+const RUNS_PER_PROMPT = 3;
+const invocationRateThreshold = 0.6;
 
 const skipTests = shouldSkipIntegrationTests();
 const skipReason = getIntegrationSkipReason();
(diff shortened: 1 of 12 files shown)


tests/azure-cost/integration.test.ts
```diff
@@ -20,8 +20,8 @@ import {
 import { softCheckSkill, isSkillInvoked, withTestResult, shouldEarlyTerminateForSkillInvocation } from "../utils/evaluate";
 
 const SKILL_NAME = "azure-cost";
-const RUNS_PER_PROMPT = 5;
-const invocationRateThreshold = 0.8;
+const RUNS_PER_PROMPT = 3;
+const invocationRateThreshold = 0.6;
```
jongio: Dropping `invocationRateThreshold` from 0.8 to 0.6 means the skill only needs correct routing 2/3 of the time to pass. The `RUNS_PER_PROMPT` reduction (5 to 3) makes sense for reducing API load, but the threshold drop is a separate concern - if the rate-limit fix is effective, 80% should be more achievable now, not less. Consider keeping the threshold at 0.8 (or at least 0.7) and only reducing the run count.

jongio (commented):
Two things to address:

1. `cost-optimization/azure-aks-anomalies.md` (line 12) has an `az rest` call that's missing the new `ClientType` header. Since SKILL.md now requires it on all Cost Management API requests, this file should be updated too. (Not in this diff, so noting here.)

2. The test invocation threshold drop (0.8 to 0.6) weakens the quality gate - see inline comment.

Everything else looks solid - the casing fix is consistent across all cost-query files, the 429 retry guidance is thorough, and the rate-limit threshold tables are a useful addition.

tests/azure-cost/integration.test.ts
kvenkatrajan: Agreed we should keep it at 80% and drop the number of runs to avoid rate limiting

taylorak: @jongio having a threshold of .8 means that the skills must be invoked >80 of the time, which is 3/3 runs. There is no leeway there. Whereas before .8 with 5 runs use to mean 4/5 runs needs to pass. 

I have updated the threshold, but am worried the test will break much easier now on a single failure.

jongio (commented):
The three fixes are solid and consistent:

- 429 retry handling now covers all three retry-after headers across every file that references rate limiting
- ClientType header is on all `az rest` code blocks (including the azure-aks-anomalies.md that was missing before)
- Filter schema casing is lowercase everywhere it needs to be

One thing still worth resolving: the threshold math on the test. With `RUNS_PER_PROMPT=3` and `invocationRateThreshold=0.8`, the effective requirement is `ceil(3 * 0.8) = 3` - all 3 runs must pass. Before, `ceil(5 * 0.8) = 4`, so 4/5 had to pass (one failure tolerated). The effective pass rate actually went *up* from 80% to 100%, which makes the test more brittle, not less.

If the rate-limit fix + ClientType bucketing are expected to eliminate 429 flakiness entirely, keeping 0.8 is fine - it just means zero tolerance. If there's still residual risk, `threshold=0.67` would restore the previous one-failure tolerance with 3 runs.

Also: `cost-query/workflow.md` Key Guardrails table (line 102) still says "Filter AND/OR" - should be lowercase to match the rest of the casing changes. One-line fix.

tests/azure-cost/integration.test.ts
taylorak: had a discussion with kay, and will be keeping it at 80% and making our descriptions crisp.

kvenkatrajan:
casing issue and threshold fixed. approved

kvenkatrajan (approved):

PR #1773, opened 2026-04-08
Fix DTS integration test to read Bicep content from workspace filesystem

### LINKED ISSUE #1724: Integration test failure: azure-prepare – Durable Task Scheduler Bicep not via create tool [Assertion mismatch]
## Diagnosis

The `generates Durable Task Scheduler infrastructure and workflow code for a workflow app` test fails because the agent creates Bicep files using **bash heredoc commands** (`cat > file.bicep << 'EOF'`) rather than the MCP `create` tool. The test collects Bicep content only from `create` MCP tool calls, so it sees `bicepContent.length === 0`.

### Failure

| Field | Value |
|-------|-------|
| **Test** | `generates Durable Task Scheduler infrastructure and workflow code for a workflow app` |
| **Test Class** | `azure-prepare_ - Integration Tests › durable-task-scheduler` |
| **File** | `tests/azure-prepare/integration.test.ts:1008` |
| **Assertion** | `expect(bicepContent.length).toBeGreaterThan(0)` |
| **Received** | `0` — no `.bicep` files found via `create` tool calls |

### Root Cause

The test uses:
````typescript
const createCalls = getToolCalls(agentMetadata, "create");
const bicepContents = createCalls
  .filter(event => args?.path?.endsWith(".bicep"))
  .map(event => args?.file_text ?? "");
const bicepContent = bicepContents.join("\n");
expect(bicepContent.length).toBeGreaterThan(0); // line 1008
```

Analysis of the agent metadata shows the agent created files using bash (29 bash tool calls vs. 2 create tool calls), with only `deployment-plan.md` written via the `create` tool. Bicep files (`infra/main.bicep`, `infra/app/durable-task-scheduler.bicep`) were written via bash heredocs.

Subsequent assertions also fail as a result:
- `expect(/Microsoft\.DurableTask\/schedulers/i.test(bicepContent)).toBe(true)` → false
- `expect(/Microsoft\.DurableTask\/schedulers\/taskHubs/i.test(bicepContent)).toBe(true)` → false
- `expect(/0ad04412-c4d5-4796-b79c-f76d14c8d402/i.test(bicepContent)).toBe(true)` → false

### Prompt Context

```
Prepare the Azure deployment infrastructure for a new workflow app that will orchestrate
a multi-step order processing pipeline. Generate the Bicep templates, RBAC assignments,
and azure.yaml. Use the eastus2 region and my current subscription.
````

### azure-prepare Skill Invocation

azure-prepare was invoked ✅. The failure is in how the agent creates output files (bash vs. create tool).

### Skill Report Context

From the SKILL-REPORT (Test 33):
- Agent reported **95% confidence** and "PASSED" — it generated all required files including DTS scheduler Bicep, TypeScript orchestrator code, RBAC assignments, and azure.yaml
- Files listed: `infra/app/durable-task-scheduler.bicep`, `src/functions/orderProcessingOrchestrator.ts`, `host.json`
- The skill completed the task correctly; the test could not verify Bicep content because files were written via bash

### Environment

| Field | Value |
|-------|-------|
| **Run** | [23978564120](https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/23978564120) |
| **Branch** | main (`a3bdc6e`) |
| **Date** | 2026-04-04 |
| **Model** | claude-sonnet-4.5 |

### Expected Fix

Options to investigate:
1. **Preferred:** Instruct the agent to use the `create` MCP tool (not bash) when writing Bicep and other infrastructure files, so that test assertions on tool call arguments work correctly
2. **Alternative:** Update test to inspect the workspace filesystem directly instead of relying solely on `create` tool calls

> Generated by [Analyze Test Run](https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/23980905110/agentic_workflow) · ● 3.7M · [◷](https://github.com/search?q=repo%3Amicrosoft%2FGitHub-Copilot-for-Azure+is%3Aissue+%22gh-aw-workflow-call-id%3A+microsoft%2FGitHub-Copilot-for-Azure%2Fanalyze-test-run%22&type=issues)

### PR DESCRIPTION
**Pull request by greenie-msft**
The 'generates Durable Task Scheduler infrastructure and workflow code' test failed because it collected Bicep content only from 'create' MCP tool calls, but the agent wrote .bicep files via bash heredocs instead.

Changed the test to read Bicep files directly from the workspace filesystem using listFilesRecursive + fs.readFileSync, which works regardless of the file creation mechanism (create tool, bash, etc.).

Fixes #1724

### DIFF
--- a/tests/azure-prepare/integration.test.ts
+++ b/tests/azure-prepare/integration.test.ts
@@ -17,7 +17,7 @@ import {
 import { hasValidationCommand } from "../azure-validate/utils";
 import { hasPlanReadyForValidation, getDockerContext, hasServicesSection, getServiceProject } from "./utils";
 import { cloneRepo } from "../utils/git-clone";
-import { expectFiles, getToolCalls, listFilesRecursive, softCheckSkill, isSkillInvoked, shouldEarlyTerminateForSkillInvocation, withTestResult } from "../utils/evaluate";
+import { doesWorkspaceFileIncludePattern, expectFiles, getToolCalls, listFilesRecursive, softCheckSkill, isSkillInvoked, shouldEarlyTerminateForSkillInvocation, withTestResult } from "../utils/evaluate";
 import * as fs from "fs";
 
 const SKILL_NAME = "azure-prepare";
@@ -1010,11 +1010,7 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
         expect(/0ad04412-c4d5-4796-b79c-f76d14c8d402/i.test(bicepContent)).toBe(true);
 
         // Must include the scheduler connection string app setting
-        const allFileContents = allFiles
-          .filter(f => !fs.statSync(f).isDirectory())
-          .map(f => fs.readFileSync(f, "utf-8"))
-          .join("\n");
-        expect(/DURABLE_TASK_SCHEDULER_CONNECTION_STRING/i.test(allFileContents)).toBe(true);
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /DURABLE_TASK_SCHEDULER_CONNECTION_STRING/i)).toBe(true);
 
         // Must include ipAllowlist to avoid 403 errors (empty list denies all traffic)
         expect(/ipAllowlist/i.test(bicepContent)).toBe(true);

### CODE REVIEW

File: tests/azure-prepare/integration.test.ts [hunk 1]
```diff
@@ -1017,11 +1010,9 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
         expect(/0ad04412-c4d5-4796-b79c-f76d14c8d402/i.test(bicepContent)).toBe(true);
 
         // Must include the scheduler connection string app setting
```
jongio, line 1012: This reimplements what `doesWorkspaceFileIncludePattern` already does - it's used across 4 other integration test files and handles node_modules skipping + try/catch for unreadable files. Consider replacing this block with:

```typescript
expect(doesWorkspaceFileIncludePattern(workspacePath!, /DURABLE_TASK_SCHEDULER_CONNECTION_STRING/i)).toBe(true);
```

That drops 4 lines and stays consistent with how other tests check for patterns.

greenie-msft: @jongio Updated per your suggestion

kvenkatrajan (approved):

PR #1780, opened 2026-04-09
Onboard azure-skills to gemini cli

saikoumudi:
## Description

In order to onboard to gemini-sli, we need to have gemini-extension.json. It's recommended to add `GEMINI.md` file and reference in gemini-extension.json. Eg: ` "contextFileName": "GEMINI.md"`. I think it's also recommended to have CLAUDE.md for claude code and we dont have it yet. we can skip the context file for now. Also there is a recommendation to add `gemini-cli-extension` topic to `azure-skills` repo, like tag for better discoverability but we can also skip for now if we dont want to tag in main repo.

Gemini hooks also work differently from copilot-cli/ claude/ open plugin hooks configuration. - https://geminicli.com/docs/hooks/

Gemini can only read from hooks/hooks.json file and so does Claude Code. we cannot do something like this in the same file. that causes errors.
image
So, currently, hooks are not supported for gemini CLI since we are supporting hooks for Claude Code. 

As soon as user installs our extension in gemini cli, they will see this warning message. 
image

Skills functionality won't impacted. Here's a screenshot from /skills in gemini cli
image

## Checklist


plugin/gemini-extension.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "azure",
```
saikoumudi, line 2: Are we okay with this name and description? it will appear here similar to the vault 

Image

kvenkatrajan (approved):

plugin/gemini-extension.json
```diff
@@ -0,0 +1,15 @@
...
+  "version": "1.0.1",
+  "description": "Microsoft Azure MCP integration for cloud resource management, deployments, and Azure services. Manage your Azure infrastructure, monitor applications, and deploy resources directly from Gemini CLI.",
```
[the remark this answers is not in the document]
saikoumudi: we are not adding context file name. added in the description so we don't forgot about the conscious decision.

plugin/gemini-extension.json
saikoumudi: we are okay with the name since that's what we are using everywhere else.

PR #1807, opened 2026-04-09
Onboard to cursor

saikoumudi:
Checked out https://github.com/[cursor/plugin-template](https://github.com/cursor/plugin-template) and other plugins for inspiration.

Screenshot from Local Plugin setup. However, i am unable to test skill invocation. Maybe onboarding to marketplace will help with it.
image

We need to submit a form here - https://cursor.com/marketplace/publish

I restructured the copilot-hooks and also added cursor-hooks as placeholder into hooks folder and updated the workflow file. The .cursor-plugin need to be in the repo root.

Added azure logo being used in vscode extension to be consumed in cursor marketplace. Couple of ways to display a logo:
1. define it in plugin.json under `logo` attribute mentioned here - https://cursor.com/docs/reference/plugins
2. Another way is to share the logo url while onboarding the cursor plugin - https://cursor.com/marketplace/publish
image

We can try doing both.


plugin/.cursor-plugin/marketplace.json
```diff
@@ -0,0 +1,16 @@
...
+    "name": "Microsoft"
+  },
+  "metadata": {
+    "description": "Microsoft Azure MCP integration for cloud resource management, deployments, and Azure services. Manage your Azure infrastructure, monitor applications, and deploy resources directly from Cursor."
+  },
+  "plugins": [
+    {
+      "name": "azure",
+      "description": "Microsoft Azure MCP integration for cloud resource management, deployments, and Azure services. Manage your Azure infrastructure, monitor applications, and deploy resources directly from Cursor.",
+      "source": "./.github/plugins/azure-skills"
```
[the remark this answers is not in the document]
saikoumudi: I have to discuss with @kvenkatrajan  and @tmeschter to see what we have decided on marketplace.json file maintenance for claude and take a similar approach for cursor as well

plugin/.cursor-plugin/plugin.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "azure",
+  "description": "Microsoft Azure MCP integration for cloud resource management, deployments, and Azure services. Manage your Azure infrastructure, monitor applications, and deploy resources directly from Cursor.",
+  "version": "1.0.1",
+  "author": {
```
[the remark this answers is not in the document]
saikoumudi: yea, the repo structure is a little different. .cursor-plugin needs to be at root level in azure-skills repo. so we are good here since we copy folders over to azure-skills repo from here

plugin/.cursor-plugin/plugin.json
```diff
@@ -0,0 +1,25 @@
...
+    "infrastructure",
+    "deployment",
+    "microsoft",
+    "devops",
+    "foundry",
+    "diagnostics"
+  ],
+  "skills": "./skills/",
+  "mcpServers": "./.mcp.json",
+  "hooks": "./hooks/cursor-hooks.json"
```
saikoumudi, line 25: good idea to exclude hooks until we test it and know it works

.github/workflows/publish-to-marketplace.yml
```diff
@@ -211,14 +211,15 @@ jobs:
...
           rsync --archive --delete --verbose "$SOURCE_DIR" "$TARGET_DIR"
 
-      # 5. Copy plugin.json, .mcp.json, .claude-plugin/plugin.json, skills/, hooks/, copilot-hooks.json, and gemini-extension.json to the repo root
+      # 5. Copy plugin.json, .mcp.json, .claude-plugin/plugin.json, .cursor-plugin/, skills/, hooks/, and gemini-extension.json to the repo root
       - name: Copy plugin files to repo root
         run: |
           cp "source-repo/plugin/.plugin/plugin.json" "target-repo/plugin.json"
           cp "source-repo/plugin/.mcp.json" "target-repo/.mcp.json"
-          cp "source-repo/plugin/copilot-hooks.json" "target-repo/copilot-hooks.json"
           mkdir -p "target-repo/.claude-plugin/"
```
jongio, line 220: The old workflow copied `copilot-hooks.json` to the target repo root. Since that copy is removed (the file now lives under `hooks/`), the stale `copilot-hooks.json` from previous runs will linger in the target repo with potentially outdated URLs - it's no longer in the URL replacement list either (though `target-repo/hooks/` covers the new location). Consider adding `rm -f "target-repo/copilot-hooks.json"` here to clean it up.

plugin/.cursor-plugin/plugin.json
jongio, line 25: These relative paths (`./skills/`, `./.mcp.json`, `./hooks/cursor-hooks.json`) need to resolve from the repo root, not from `.cursor-plugin/`. The other plugin manifests use the same pattern and it works, so this is likely fine - but since you mentioned you couldn't test skill invocation yet, worth verifying that Cursor resolves paths from the repo root when `plugin.json` is inside `.cursor-plugin/`.

plugin/.cursor-plugin/marketplace.json
```diff
@@ -0,0 +1,16 @@
...
+    }
+  ]
+}
```
jongio: Missing trailing newline. This PR fixes the same issue in `hooks.json` and `copilot-hooks.json` - should be consistent here too.

jongio (commented):
A few things I noticed that aren't covered by the existing review threads.

plugin/.cursor-plugin/plugin.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "azure",
+  "description": "Microsoft Azure MCP integration for cloud resource management, deployments, and Azure services. Manage your Azure infrastructure, monitor applications, and deploy resources directly from Cursor.",
+  "version": "1.0.1",
```
saikoumudi, line 4: shouldn't the version be 1.0.0 to begin with?

saikoumudi: and update scripts as appropriate

charris-msft: We should mention skills: "Microsoft Azure MCP and Skills integration..."

saikoumudi: yes, we aren't mentioning anything about skills in any of the client description. i will update it.

plugin/.cursor-plugin/plugin.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "azure",
+  "description": "Microsoft Azure MCP and Skills integration for cloud resource management, deployments, and Azure services. Manage your Azure infrastructure, monitor applications, and deploy resources directly from Cursor.",
+  "version": "1.0.1",
```
jongio, line 4: The version management pipeline doesn't know about this file yet. Four places need updating:

- `scripts/src/update-plugin-version.ts` - add to the default `pluginFiles` array (line 21)
- `.releaserc.json` - add `plugin/.cursor-plugin/plugin.json` to the `assets` array (line 21)
- `scripts/src/check-plugin-version-pr.ts` - add to `PLUGIN_PATHS` (line 14)
- `.github/workflows/pr-plugin-version-check.yml` - add to the `paths` trigger (line 8)

Without these, the Cursor plugin version won't get bumped by semantic-release and manual version edits won't be caught by CI.

plugin/.cursor-plugin/marketplace.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "azure-skills",
+  "owner": {
+    "name": "Microsoft"
+  },
+  "metadata": {
+    "description": "Microsoft Azure MCP integration for cloud resource management, deployments, and Azure services. Manage your Azure infrastructure, monitor applications, and deploy resources directly from Cursor."
+  },
```
jongio: This description says 'MCP integration' but `plugin.json` in the same directory says 'MCP and Skills integration'. Should be consistent - the 'and Skills' addition per charris-msft's feedback should apply here too.

jongio (commented):
Two things the existing review threads don't cover - both related to integrating .cursor-plugin into the existing version management pipeline.

plugin/.cursor-plugin/marketplace.json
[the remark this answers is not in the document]
saikoumudi: decided to add marketplace.json file in azure-skills repo alone, since we will onboard that repo to cursor marketplace. having marketplace.json file two repo may add confusion. so deleted it here

plugin/.cursor-plugin/plugin.json
saikoumudi: I was unable to test hooks both a plugin hooks and project level hooks. However, i referenced https://github.com/cursor/plugins/blob/08c2bbe2ae8a022a21dc6c32faf611f14a6e8343/continual-learning/hooks/hooks.json and this thread - https://forum.cursor.com/t/inconsistent-working-directory-for-plugin-hook-commands/153236

i see their hook working, and we can try to add to ours as custom hooks and see how that goes.
image

plugin/.cursor-plugin/plugin.json
saikoumudi: made these changes.

jongio (commented):
Previous feedback is addressed - version scripts, .releaserc, PR check workflow, descriptions, and stale copilot-hooks.json cleanup all look good.

One remaining gap: the `sync-to-microsoft-skills` job in `publish-to-marketplace.yml` doesn't include `.cursor-plugin/plugin.json` anywhere. The `sync-to-microsoft-azure-skills` job was updated correctly, but the skills repo sync still only handles `.claude-plugin/plugin.json` and `.plugin/plugin.json` in three places:

1. **Version restore** (step 5) - cursor-plugin version won't be pinned back to the current target version after rsync
2. **Version file list** (step 6) - `.cursor-plugin/plugin.json` isn't in the `VERSION_FILES` array
3. **Non-JSON change filter** (step 6) - version-only changes to cursor's plugin.json will incorrectly count as content changes and trigger unnecessary version bumps

The rsync in step 4 WILL copy `.cursor-plugin/` into the target, so the files will be there - but the version handling won't track them. This will cause version drift between `microsoft/skills` and the source once releases start flowing.

plugin/hooks/cursor-hooks.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": 1,
+  "hooks": {
+    "postToolUse": [
+      {
+        "type": "command",
+        "command": "bash ${CURSOR_PLUGIN_ROOT}/hooks/scripts/track-telemetry.sh"
```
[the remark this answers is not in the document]
saikoumudi: this is not required.

plugin/.cursor-plugin/plugin.json
saikoumudi: When Cursor installs a plugin, it caches the full repo contents under `~/.cursor/plugins/cache/cursor-public/<plugin-name>/<id>/`. I verified via other plugin installations that cursor resolves all these relative paths correctly. 

 Cursor appears to load plugins dynamically at runtime — simply placing the plugin folder under `~/.cursor/plugins/cache/` or `~/.cursor/plugins/local/` doesn't cause Cursor to discover and load the skills or hooks.

As a workaround, I verified skill invocation by placing them at the project level (`.cursor/skills/`), which works correctly. However, hooks don't appear to work even at the project level via `.cursor/hooks.json` — so I'm unable to validate hook behavior regardless of the plugin loading mechanism.

.github/workflows/publish-to-marketplace.yml
```diff
@@ -245,7 +249,7 @@ jobs:
         run: |
           CURRENT_VERSION="${{ steps.version.outputs.current }}"
           TARGET_DIR="target-repo/.github/plugins/azure-skills/"
-          for f in "${TARGET_DIR}.claude-plugin/plugin.json" "${TARGET_DIR}.plugin/plugin.json" "target-repo/plugin.json" "target-repo/.claude-plugin/plugin.json"; do
+          for f in "${TARGET_DIR}.claude-plugin/plugin.json" "${TARGET_DIR}.cursor-plugin/plugin.json" "${TARGET_DIR}.plugin/plugin.json" "target-repo/plugin.json" "target-repo/.claude-plugin/plugin.json" "target-repo/.cursor-plugin/plugin.json"; do
             if [ -f "$f" ]; then
```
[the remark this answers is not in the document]
saikoumudi: this has been addressed

plugin/.cursor-plugin/plugin.json
```diff
@@ -0,0 +1,25 @@
...
+    "infrastructure",
+    "deployment",
+    "microsoft",
+    "devops",
+    "foundry",
+    "diagnostics"
+  ],
+  "skills": "./skills/",
+  "mcpServers": "./.mcp.json",
+  "hooks": "./hooks/cursor-hooks.json"
```
jongio, line 25: The `azure-logo.svg` was added to `plugin/assets/` but this manifest doesn't have a `logo` field referencing it. The PR description mentions doing both the plugin.json field and the marketplace form upload - worth adding `"logo": "./assets/azure-logo.svg"` here if Cursor supports it, so the logo is self-contained in the plugin config.

.github/workflows/publish-to-marketplace.yml
```diff
@@ -211,24 +212,28 @@ jobs:
           #   --verbose   list changes
           rsync --archive --delete --verbose "$SOURCE_DIR" "$TARGET_DIR"
 
-      # 5. Copy plugin.json, .mcp.json, .claude-plugin/plugin.json, skills/, hooks/, copilot-hooks.json, and gemini-extension.json to the repo root
+      # 5. Copy plugin.json, .mcp.json, .claude-plugin/plugin.json, .cursor-plugin/, skills/, hooks/, and gemini-extension.json to the repo root
       - name: Copy plugin files to repo root
         run: |
           cp "source-repo/plugin/.plugin/plugin.json" "target-repo/plugin.json"
           cp "source-repo/plugin/.mcp.json" "target-repo/.mcp.json"
-          cp "source-repo/plugin/copilot-hooks.json" "target-repo/copilot-hooks.json"
```
fanyang-mono, line 219: You probably want to replace this with the new copilot-hooks.json file instead of deleting it completely.

.github/workflows/publish-to-marketplace.yml
```diff
@@ -211,24 +212,28 @@ jobs:
...
           mkdir -p "target-repo/.claude-plugin/"
           cp "source-repo/plugin/.claude-plugin/plugin.json" "target-repo/.claude-plugin/plugin.json"
+          mkdir -p "target-repo/.cursor-plugin/"
+          rsync --archive --delete --verbose "source-repo/plugin/.cursor-plugin/" "target-repo/.cursor-plugin/"
```
fanyang-mono: Please double check if this is needed.

.github/workflows/publish-to-marketplace.yml
```diff
@@ -211,24 +212,28 @@ jobs:
...
           mkdir -p "target-repo/skills/"
           rsync --archive --delete --verbose "source-repo/plugin/skills/" "target-repo/skills/"
           mkdir -p "target-repo/hooks/"
           rsync --archive --delete --verbose "source-repo/plugin/hooks/" "target-repo/hooks/"
           cp "source-repo/plugin/gemini-extension.json" "target-repo/gemini-extension.json"
 
+          # Clean up stale copilot-hooks.json from repo root (previously copied there, now lives under hooks/)
+          rm -f "target-repo/copilot-hooks.json"
```
fanyang-mono, line 229: I would suggest doing this one time file removal manually.

.github/workflows/publish-to-marketplace.yml
fanyang-mono: nvm, it will be copied over since it will be part of the hooks folder

.github/workflows/publish-to-marketplace.yml
```diff
@@ -258,9 +263,11 @@ jobs:
         run: |
           VERSION_FILES=(
             ".github/plugins/azure-skills/.claude-plugin/plugin.json"
+            ".github/plugins/azure-skills/.cursor-plugin/plugin.json"
             ".github/plugins/azure-skills/.plugin/plugin.json"
             "plugin.json"
             ".claude-plugin/plugin.json"
+            ".cursor-plugin/plugin.json"
```
fanyang-mono: Please double check if this file is necessary.

saikoumudi: synced with Tom offline. we added it for claude because onboarding to claude marketplace needed that file. However, i dont think we need it for cursor, as cursor official docs https://cursor.com/docs/reference/plugins talks about having only marketplace file at the root level.
image


once we submit the request to onboard to cursor and if there are any gaps, we can come back to adding it.

.github/workflows/publish-to-marketplace.yml
saikoumudi: i will let the pipeline remove it and i can have another PR to delete this line if needed, just to reduce the manual work again.

plugin/.cursor-plugin/plugin.json
saikoumudi: yes, good catch. i didnt push that change. will do now.

.github/workflows/publish-to-marketplace.yml
tmeschter: We don't need a top-level copilot-hooks.json file at all?

tmeschter (approved):

.github/workflows/publish-to-marketplace.yml
saikoumudi: i moved it to hooks folder so copilot cli now refers to the one in hooks folder. so, we dont need it at root level.

fanyang-mono (approved):

.github/workflows/publish-to-marketplace.yml
```diff
@@ -225,10 +225,13 @@ jobs:
           rsync --archive --delete --verbose "source-repo/plugin/hooks/" "target-repo/hooks/"
           cp "source-repo/plugin/gemini-extension.json" "target-repo/gemini-extension.json"
```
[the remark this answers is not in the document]
saikoumudi: will create another pr to azure-skills repo with marketplace.json

PR #1834, opened 2026-04-13
Add azure-iac-generator and azure-infrastructure-sync skills; enhance azure-resource-visualizer and azure-validate

Ba4bes:
## Description

This PR is based on [AzVerify](https://github.com/Azure/AZVerify). The PR is pretty big, I have structured the functionality by creating two new skills and enhancing two existing skills.
This closes the infra-as-code<>Bicep<>diagram loop: generate Bicep from live Azure or diagrams, detect drift across artifacts, and validate against Azure Policy — all with full reference documentation and test coverage.

Would like to have a conversation to discuss if this is the best approach or if other structures would work better.

## Checklist

- [x] Tests pass locally (`cd tests && npm test`) _(tested only skills that were added/changed)_
- [ ] **If modifying skill descriptions:** verified routing correctness with integration tests (`npm run test:skills:integration -- <skill>`) _Could only use npm run tests:integration. But for some reason tests are incredibly slow on my system, so would prefer to discuss first before I spend more time on these_
- [x] Version bumped in skill frontmatter (if skill files changed)


plugin/skills/azure-infrastructure-sync/references/bicep-diagram-sync-workflow.md
```diff
@@ -0,0 +1,17 @@
...
+
+| # | Action | Details |
+|---|--------|---------|
+| 1 | **Accept Inputs** | Get: (a) project folder containing Bicep files, (b) Draw.io file path, (c) optionally a specific `.bicep` file if not using `main.bicep`. |
+| 2 | **Parse Diagram** | Apply [diagram-parsing.md](procedures/diagram-parsing.md) to build diagram resource model. |
+| 3 | **Parse Bicep** | Apply [bicep-parsing.md](procedures/bicep-parsing.md) to build Bicep resource model. Read `.bicepparam` to resolve parameter values. |
+| 4 | **Match Resources** | Apply [resource-matching.md](procedures/resource-matching.md) across diagram and Bicep models. |
+| 5 | **Present Drift Report** | Show status: ✅ In Sync, ⬜ Diagram Only, 📄 Bicep Only. |
+| 6 | **Offer Resolution** | Options: update Bicep (add resources from diagram), update diagram (add resources from Bicep), selective, or no action. |
+| 7 | **Execute Resolution** | If updating Bicep: generate new resource blocks following [bicep-best-practices.md](../azure-iac-generator/references/bicep-best-practices.md). Run [azure-deployment-verification.md](azure-deployment-verification.md) before finalizing. If updating diagram: modify Draw.io XML. |
```
jongio, line 17: This references `../azure-iac-generator/references/bicep-best-practices.md` - a path into another skill's directory. If iac-generator gets renamed or restructured, this silently breaks. Either copy bicep-best-practices.md into this skill's references (matching the duplication pattern used elsewhere), or move it to a shared location both skills reference.

plugin/skills/azure-iac-generator/references/azure-to-bicep-workflow.md
```diff
@@ -0,0 +1,110 @@
...
+## Step 4 — Filter Non-Deployable Resources
+
+Remove auto-created and hidden resources per [resource-filtering.md](procedures/resource-filtering.md) (use "Exclude for Bicep" column). Keep: Application Insights, Log Analytics, user-assigned identities, diagnostic settings. If >30 resources remain, warn user and offer to filter by type.
+
+## Step 5 — Deep Property Extraction (HARD GATE)
+
+**Do NOT skip this step.** For every resource, retrieve the **full** configuration:
+
+| Resource Type | Tool | Also Extract |
+|---|---|---|
```
jongio, line 28: Step 4 links to `procedures/resource-filtering.md`, but that file doesn't exist in azure-iac-generator. It only exists in azure-infrastructure-sync and azure-resource-visualizer. This is a broken link - the agent won't be able to load filtering rules during Bicep generation.

plugin/skills/azure-iac-generator/references/version-currency.md
```diff
@@ -0,0 +1,35 @@
...
+//   DOTNET|10.0 → .NET 10 (LTS, supported until Nov 2028)
+//   NODE|22-lts → Node.js 22 LTS (supported until Apr 2027)
+//   PYTHON|3.13 → Python 3.13 (supported until Oct 2029)
+//   JAVA|21     → Java 21 LTS (supported until Sep 2028)
+param appServiceRuntimeStack = 'DOTNET|10.0'
+```
+
+## Handling Outdated Values from Azure
+
+When extracting from Azure and the current value uses an outdated version:
```
jongio, line 32: This says `Use the current versions as default` and `Never silently keep an end-of-life runtime.` But azure-to-bicep-workflow.md Step 8 says `Use the exact runtime version from Azure` and only upgrade if EOL. These directly conflict for the outdated-but-still-supported case (e.g., .NET 8 when .NET 10 is current but 8 is still in support). Since the skill's goal is `environment-identical redeployment,` the workflow's fidelity-first approach seems right - consider aligning this file to match.

plugin/skills/azure-resource-visualizer/references/drawio-diagram-conventions.md
```diff
@@ -0,0 +1,144 @@
...
+- There is no generic `Event_Grid.svg` — use `System_Topic.svg`, `Event_Grid_Topics.svg`, or `Event_Grid_Domains.svg`
+- `Microsoft.AppConfiguration/configurationStores` → `img/lib/mscae/App_Configuration.svg` — this icon is in the **mscae** library, **not** azure2. There is no `App_Configuration.svg` or `App_Configurations.svg` in azure2 at all; using any azure2 path produces an empty broken icon.
+- When uncertain about any path, verify it exists at `https://github.com/jgraph/drawio/tree/dev/src/main/webapp/img/lib/azure2`
+- If no match, use the closest available icon and note the gap
+
+---
+
+## 3. Resource Shape Rules
+
+For each resource icon cell:
```
jongio, line 54: This maps `Microsoft.Web/sites` to `img/lib/azure2/compute/App_Services.svg`, but `azure-stencil-mapping.json` maps it to `img/lib/azure2/app_services/App_Services.svg`. The JSON file even has a note: `CATEGORY: app_services - NOT compute.` The conventions file should use `app_services/` not `compute/`.

plugin/skills/azure-iac-generator/references/azure-resource-model.md
```diff
@@ -0,0 +1,102 @@
+# Azure Resource Metadata Model
```
jongio, line 1: This file is byte-identical across azure-iac-generator, azure-infrastructure-sync, and azure-resource-visualizer. Same for azure-authentication.md (3 copies), bicep-parsing.md (3 copies), azure-stencil-mapping.json (2 copies), azure-deployment-verification.md (2 copies), resource-filtering.md (2 copies), and diagram-parsing.md (2 copies). That's ~1,500 lines of duplication across 8 files. When someone updates auth procedures in one skill, the other copies go stale. Consider either introducing `plugin/skills/_shared/references/` for the identical files, or at minimum adding a canonical-copy header and opening a tracking issue.

jongio (changes requested):
Solid work closing the diagram-Bicep-Azure loop - the skill decomposition makes sense and the routing-conflicts test is a pattern other PRs should follow. A few things need fixing before this can merge:

1. There's a broken reference link (resource-filtering.md missing from iac-generator)
2. Version-currency.md and azure-to-bicep-workflow.md give contradictory advice on handling outdated runtimes
3. The stencil path for Microsoft.Web/sites in drawio-diagram-conventions.md contradicts azure-stencil-mapping.json
4. A cross-skill relative path (../azure-iac-generator/) creates fragile coupling

The biggest discussion point: 8 reference files are duplicated 2-3 times across skills (~1,500 lines). Since you mentioned wanting a conversation about approach - I'd suggest either a shared references location or at minimum a tracking issue for consolidation.

plugin/skills/azure-iac-generator/references/azure-resource-model.md
tmeschter: Skills cannot reference files outside of the skill directory (that is, the directory with the SKILL.md) so for the moment we have little choice but to duplicate the files.

We may want to invest in a build system that can take the committed files, manipulate them in various useful ways (e.g. text templating, automatic versioning, copying common files) and produce the final plugin as a separate output. That would cover this scenario as well as a number of others.

plugin/skills/azure-iac-generator/SKILL.md
```diff
@@ -0,0 +1,79 @@
+---
+name: azure-iac-generator
```
tmeschter, line 2: We also need to support Terraform--perhaps not in this change, but at some point down the line. I see two options for this:

1. Update this skill to support both, in which case the name can remain the same but in the future we'll need to do a significant refactoring of the SKILL.md.
2. Have separate skills for Bicep (this one) and Terraform (TBD). If we want to go down that route, it would be better to rename this to `azure-bicep-generator` or something like that.

My strong suggestion is to go with option 1. As such there are no changes to be made at this time, but I want you to be aware of what might be coming down the line.

tmeschter (changes requested):
I like where this is going but as a practical matter this needs to be broken up into four separate PRs, one per skill (with perhaps the PR for `azure-infrastructure-sync` dependent on the others going in first).

plugin/skills/azure-iac-generator/SKILL.md
Ba4bes: I thought about Terraform (already had requests for it), that is actually why I called this one azure-iac-generator instead of azure-bicep-generator. With the line of thinking of not having too many separate skills it made sense to me to have both Bicep and Terraform in here in the future.

plugin/skills/azure-iac-generator/references/azure-resource-model.md
Ba4bes: In my own version of this Skills-set I used a `shared` folder, which I do like a lot better than the several copies. But dependencies will be a challenge with that option. I will work with the copies for now

Ba4bes:
@jongio @tmeschter Currently working on creating 4 seperated PRs and fixing the comments while I am at it. Hope to finish tommorow. Will keep this open for just a sec to link the new pull requests so structure is clear, but can be closed afterwards

jongio:
Sounds good - splitting into 4 PRs is the right call. It'll make each one easier to review and unblock the skills that don't depend on each other. Happy to review the individual PRs as they come in.

jongio (commented):
A few more items I noticed on a deeper pass - these are about skill-files.instructions.md compliance.

plugin/skills/azure-infrastructure-sync/SKILL.md
```diff
@@ -0,0 +1,92 @@
...
+metadata:
+  author: Microsoft
+  version: "1.0.0"
+---
+
+# Azure Infrastructure Sync
+
+Compare and synchronize infrastructure artifacts to detect drift. Supports four comparison modes across Draw.io diagrams, Bicep templates, and live Azure resources.
+
+## Prerequisites
```
jongio, line 14: This skill is missing the required **Quick Reference** section (summary table with MCP tools, CLI commands, best-for) and a dedicated **MCP Tools** section with parameter tables. Both are required by `skill-files.instructions.md`. The Prerequisites table is a good start - you could expand it or add a Quick Reference above the routing section that lists the Azure MCP tools, `az` CLI fallbacks, and what each comparison mode outputs.

plugin/skills/azure-resource-visualizer/SKILL.md
```diff
@@ -1,181 +1,103 @@
...
+description: "Analyze Azure resource groups and generate architecture diagrams as Mermaid or Draw.io. Supports live Azure resource discovery, sketch/description-to-diagram conversion, and detailed relationship mapping. WHEN: create architecture diagram, visualize Azure resources, show resource relationships, generate Mermaid diagram, analyze resource group, diagram my resources, architecture visualization, resource topology, map Azure infrastructure, draw.io diagram, sketch to diagram, convert sketch to architecture, generate draw.io, create draw.io from azure. DO NOT USE FOR: comparing diagrams against live Azure or Bicep (use azure-infrastructure-sync), generating Bicep templates (use azure-iac-generator)."
 license: MIT
 metadata:
   author: Microsoft
-  version: "1.0.1"
+  version: "1.1.0"
 ---
```
jongio, line 8: The rewrite dropped the **Error Handling** section that the original had. `skill-files.instructions.md` lists Error Handling as a required section. Even though workflows moved to reference files, the main SKILL.md should still have a table covering common errors (not authenticated, empty resource group, unsupported resource types, diagram generation failures) so the agent knows what to do without loading a reference file.

plugin/skills/azure-resource-visualizer/SKILL.md
```diff
@@ -1,181 +1,103 @@
...
 
-# Azure Resource Visualizer - Architecture Diagram Generator
+# Azure Resource Visualizer
 
-A user may ask for help understanding how individual resources fit together, or to create a diagram showing their relationships. Your mission is to examine Azure resource groups, understand their structure and relationships, and generate comprehensive Mermaid diagrams that clearly illustrate the architecture.
+Generate architecture diagrams from Azure resource groups, sketches, or descriptions.
```
jongio, line 12: The original SKILL.md had valuable operational guidance that got removed in the rewrite: detailed diagram requirements (include SKUs/tiers in labels, use `-->` vs `-.->` vs `==>` for different connection types), quality standards (accuracy, completeness, clarity), and tool usage patterns. Some of this may have moved to reference files, but the core diagram construction rules in the Mermaid section here are quite minimal compared to what was there before. Consider keeping the key quality requirements in the main SKILL.md even if detailed workflows live in references.

Ba4bes:
Thank you for the extra checks @jongio  and @tmeschter .
I have created the new pull requests and have implemented the comments that were there so far. Will close this pull request now.

PR #1868, opened 2026-04-14
feat: migrate Batch 3 Jest integration tests to Vally (10 complex skills)

wbreza:
## Summary

Migrates 10 Jest integration test suites to Vally eval configs (Batch 3), building on the foundation established in #1912. Includes template, complex multi-section eval specs, and skills with file-based graders.

### Skills Migrated

| Skill | Stimuli | Key Graders |
|-------|---------|-------------|
| `_template` | 5 | skill-invocation, completed, output-contains, output-not-matches, tool-calls |
| `appinsights-instrumentation` | 4 | skill-invocation, completed, file-matches, output-contains, output-not-matches |
| `azure-cloud-migrate` | 2 | skill-invocation, completed, file-exists, output-not-matches |
| `azure-deploy` | 3 | skill-invocation, completed, file-exists, file-matches, output-matches, regex |
| `azure-enterprise-infra-planner` | 8 | skill-invocation, completed, file-exists, output-not-matches |
| `azure-hosted-copilot-sdk` | 8 | skill-invocation, completed, output-contains, output-not-matches |
| `azure-prepare` | 3 | skill-invocation, completed, file-exists, file-matches, output-matches, regex |
| `azure-resource-lookup` | 7 | skill-invocation, completed, output-contains, output-not-matches |
| `azure-resource-visualizer` | 4 | skill-invocation, completed, file-matches, output-not-matches |
| `azure-validate` | 12 | skill-invocation, completed, file-matches, output-matches, output-not-matches |

**Total: 56 stimuli across 10 skills**

### Changes

- Adds/updates 10 `evals/*/eval.yaml` configs (including `_template` reference spec)
- Uses advanced grader types: `file-exists`, `file-matches`, `regex`
- Updates CI workflow (`eval.yml`) for new eval paths
- Cleans up legacy test configs (`azure-deploy`, `azure-enterprise-infra-planner`, `azure-prepare`)

### Eval Patterns

All specs follow the validated patterns from #1912:
- Model: `claude-sonnet-4.5`
- Executor: `copilot-sdk`
- Duration format: human-friendly (`"10m"`, `"7m"`)
- Valid grader types only (per published `@microsoft/vally-cli@0.4.0`)

### Dependencies

- Requires #1912 (merged :white_check_mark:)

### DIFF
--- a/evals/appinsights-instrumentation/eval.yaml
+++ b/evals/appinsights-instrumentation/eval.yaml
@@ -28,7 +28,7 @@ tags:
 
 environment:
   skills:
-    - "plugin/skills/appinsights-instrumentation"
+    - "../../plugin/skills/appinsights-instrumentation"
 
 config:
   runs: 5
--- a/evals/appinsights-instrumentation/eval.yaml
+++ b/evals/appinsights-instrumentation/eval.yaml
@@ -26,6 +26,10 @@ tags:
   type: integration
   skill: appinsights-instrumentation
 
+environment:
+  skills:
+    - "plugin/skills/appinsights-instrumentation"
+
 config:
   runs: 5
   timeout: 600
--- a/evals/appinsights-instrumentation/eval.yaml
+++ b/evals/appinsights-instrumentation/eval.yaml
@@ -1,0 +1,160 @@
+# Vally eval config — migrated from Jest integration tests
+# Source: tests/appinsights-instrumentation/integration.test.ts
+# Migration: Jest → Vally (Batch 3A — Complex/Smaller)
+#
+# Global graders (evaluate#125 workaround): completed and no_runtime_failure
+# are duplicated into every stimulus block below.
+#
+# Jest notes:
+# - Invocation test 2 uses setup: writes package.json to workspace to indicate Node.js project
+# - Quality test 1 uses fs.cpSync("./appinsights-instrumentation/resources/aspnetcore-app/", workspace)
+#   → mapped to environment.files comment (Vally resource copying)
+# - Quality test 2 uses fs.cpSync("./appinsights-instrumentation/resources/python-app/", workspace)
+#   + doesWorkspaceFileIncludePattern for Python instrumentation patterns:
+#     /from\s+azure\.monitor\.opentelemetry\s+import\s+configure_azure_monitor/
+#     /configure_azure_monitor\s*\(/
+#   → mapped to file-matches grader
+
+name: appinsights-instrumentation-integration-eval
+description: |
+  Integration evaluation for appinsights-instrumentation skill — migrated from Jest
+  integration tests. Tests skill routing for ASP.NET Core and Node.js instrumentation
+  prompts (5 runs, 80% threshold), plus content quality checks for auto-instrumentation
+  guidance and Python Azure Monitor OpenTelemetry patterns.
+
+tags:
+  type: integration
+  skill: appinsights-instrumentation
+
+config:
+  runs: 5
+  timeout: 600
+  executor: copilot-sdk
+  model: claude-sonnet-4
+
+scoring:
+  threshold: 0.8
+
+stimuli:
+  # ═══════════════════════════════════════════════════════════════════════
+  # skill-invocation
+  # ═══════════════════════════════════════════════════════════════════════
+
+  # ── aspnet-core-instrumentation ──
+  # Jest: "invokes skill for App Insights instrumentation request"
+  # Assertions: softCheckSkill + isSkillInvoked (invocation rate ≥ 80%)
+  - name: "ASP.NET Core App Insights instrumentation"
+    prompt: "How do I add Application Insights to my ASP.NET Core web app?"
+    tags:
+      type: integration
+      tier: full
+      cost: free
+      area: routing
+    graders:
+      - type: skill-invocation
+        config:
+          required:
+            - appinsights-instrumentation
+      # Global: has_output
+      - type: completed
+      # Global: no_runtime_failure
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  # ── nodejs-telemetry ──
+  # Jest: "invokes skill for Node.js telemetry request"
+  # Assertions: softCheckSkill + isSkillInvoked (invocation rate ≥ 80%)
+  # Jest setup: writes package.json { name: "test-app", version: "1.0.0" } to workspace
+  # Vally note: workspace file creation via environment.files if supported
+  - name: "Node.js telemetry request"
+    prompt: "Add telemetry to my Node.js application"
+    # environment:
+    #   files:
+    #     - content: '{"name":"test-app","version":"1.0.0"}'
+    #       path: "package.json"
+    tags:
+      type: integration
+      tier: full
+      cost: free
+      area: routing
+    graders:
+      - type: skill-invocation
+        config:
+          required:
+            - appinsights-instrumentation
+      # Global: has_output
+      - type: completed
+      # Global: no_runtime_failure
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  # ═══════════════════════════════════════════════════════════════════════
+  # content-quality
+  # ═══════════════════════════════════════════════════════════════════════
+
+  # ── auto-instrumentation-aspnet ──
+  # Jest: "response mentions auto-instrumentation for ASP.NET Core App Service app"
+  # Assertions: doesAssistantMessageIncludeKeyword(agentMetadata, "auto-instrument")
+  # Jest setup: fs.cpSync("./appinsights-instrumentation/resources/aspnetcore-app/", workspace)
+  # Vally note: resource copying not directly supported; copilot-sdk executor may handle
+  #   workspace differently. The prompt is self-contained enough to trigger correct guidance.
+  - name: "ASP.NET Core auto-instrumentation guidance"
+    prompt: "Add App Insights instrumentation to my C# web application in Azure App Service"
+    # environment:
+    #   files:
+    #     - src: "tests/appinsights-instrumentation/resources/aspnetcore-app/"
+    #       dest: "."
+    tags:
+      type: integration
+      tier: full
+      cost: free
+      area: output
+    graders:
+      - type: output-contains
+        config:
+          substring: "auto-instrument"
+      # Global: has_output
+      - type: completed
+      # Global: no_runtime_failure
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  # ── python-instrumentation-code ──
+  # Jest: "mentions App Insights in response"
+  # Assertions:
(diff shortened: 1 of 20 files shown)


jongio (commented):
Big migration, and the Batch 3A files + `_template` + `.vally.yaml` look close to what #1912's Vally CLI will consume. Seven issues worth addressing before this comes out of draft, flagged inline. The most important ones:

1. `azure-deploy`, `azure-prepare`, and `azure-validate` still use the old Waza schema (top-level `skill`, `version`, `metrics`, `graders`; `config.trials_per_task`, `timeout_seconds`, `parallel`). Once #1912 switches CI to `@microsoft/vally-cli`, those three won't parse. Other seven files are on the new schema.

2. Every stimulus is tagged `cost: free`, but `.vally.yaml` reserves `cost: free` for static non-LLM checks (see #1912). The `pr` suite filters on `cost: free`, so this pulls 125 LLM-backed stimuli (some with 60-min timeouts) into the PR gate. They should be tagged `cost: llm` (or whatever non-free tag `.vally.yaml` settles on).

3. A few stimuli will fail by design because workspace seed data got commented out during migration: `azure-cloud-migrate` (needs `cloneRepo` of AWS samples), `azure-enterprise-infra-planner` response-quality stimuli (needed multi-turn followUp confirmations), the 3 `azure-hosted-copilot-sdk` stimuli that needed `setupExpressApp` / `setupCopilotSdkApp`, and `appinsights-instrumentation` Python/ASP.NET quality checks.

4. Negative assertions from `expectFiles` got dropped in `azure-deploy`: the Bicep stimuli assert `infra/*.bicep` exists but don't assert `infra/*.tf` is absent (and vice-versa for the Terraform section).

Happy to help triage once you've picked an approach for the seed-data gap.

evals/azure-deploy/eval.yaml
```diff
@@ -0,0 +1,1281 @@
...
+environment:
+  skills:
+    - "../../plugin/skills/azure-deploy"
+
+config:
+  trials_per_task: 1
+  timeout_seconds: 3600  # 60 min — matches deployTestTimeoutMs / brownfieldTestTimeoutMs
+  parallel: false
+  executor: copilot-sdk
+  model: claude-sonnet-4-20250514
```
jongio: This file is on the old Waza schema: `skill`, `version`, `config.trials_per_task`, `timeout_seconds`, `parallel`, top-level `metrics`, top-level `graders`. The 7 Batch 3A files and `_template` use the new Vally schema (`tags`, `scoring`, `config.runs`, `config.timeout`).

Once #1912 lands and CI switches to `@microsoft/vally-cli`, this file, `azure-prepare/eval.yaml`, and `azure-validate/eval.yaml` won't parse under the Vally schema. They'll either get rejected as having unknown fields or fall through to defaults for `runs`/`timeout` (so the 60-min timeout claimed in the PR description silently won't apply).

Also line 1 still points at `waza/main/schemas/eval.schema.json`, which contradicts the stated migration target.

Convert the 3 Batch 3B files to match Batch 3A's shape:
- drop `skill`, `version`, `metrics`, top-level `graders`
- replace `config.trials_per_task` -> `config.runs`, `timeout_seconds` -> `timeout`, drop `parallel`
- add `tags:` + `scoring: { threshold: 0.8 }`
- remove the yaml-language-server directive or repoint it at the Vally schema once it's published

evals/appinsights-instrumentation/eval.yaml
```diff
@@ -0,0 +1,164 @@
...
+
+  # ── aspnet-core-instrumentation ──
+  # Jest: "invokes skill for App Insights instrumentation request"
+  # Assertions: softCheckSkill + isSkillInvoked (invocation rate ≥ 80%)
+  - name: "ASP.NET Core App Insights instrumentation"
+    prompt: "How do I add Application Insights to my ASP.NET Core web app?"
+    tags:
+      type: integration
+      tier: full
+      cost: free
```
jongio, line 55: `cost: free` is repeated on every stimulus in this PR (125 total across all 10 files), but `.vally.yaml` reserves `cost: free` for static non-LLM checks:

```yaml
pr:
  description: "All free-tier evals for PR gate, <2min"
  filter:
    cost: free
```

All stimuli here run `executor: copilot-sdk` (real LLM) with 10-60 minute timeouts. Running `vally eval --suite pr` would pull all 125 into the PR gate, which isn't the `<2min` contract `.vally.yaml` promises.

Use `cost: llm` (or whatever non-free tag the framework settles on) for copilot-sdk stimuli, and reserve `cost: free` for static trigger/routing checks that don't call an LLM.

evals/azure-cloud-migrate/eval.yaml
```diff
@@ -0,0 +1,129 @@
...
+          path: "**/migration-status.md"
+      - type: file-exists
+        config:
+          path: "**/migration-assessment-report.md"
+      # Global: has_output
+      - type: completed
+      # Global: no_runtime_failure
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
```
jongio, line 88: This test's whole point in Jest was to migrate a real AWS Lambda repo that `cloneRepo()` dropped into the workspace. With `environment.commands` commented out, the skill runs on an empty workspace and the stimulus asks it to "migrate this Lambda to Azure" when there's no Lambda to migrate.

`file-exists: **/migration-status.md` and `file-exists: **/migration-assessment-report.md` will almost certainly never hit: the skill's migration flow expects source files to analyze, so it either bails out early or produces generic advisory text without creating those deliverables.

Same issue for `lambda-webapp-migration` below.

Options:
- Gate these stimuli on an `environment.commands` / `environment.files` primitive that Vally supports (or file an ask against `microsoft/evaluate`).
- Replace the clone with a fixture directory checked into the repo that the `copilot-sdk` executor seeds from.
- Mark both stimuli as skipped/pending until workspace setup lands, rather than leaving them as guaranteed failures that'll drag the skill's pass rate down to 0%.

evals/azure-enterprise-infra-planner/eval.yaml
```diff
@@ -0,0 +1,263 @@
...
+            - azure-enterprise-infra-planner
+      - type: file-exists
+        config:
+          path: "**/.azure/infrastructure-plan.json"
+      # Global: has_output
+      - type: completed
+      # Global: no_runtime_failure
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
```
jongio, line 222: This stimulus and `bicep-file-generation` below both expect files that the original Jest tests only produced after multi-turn confirmations:

- `plan-json-generation`: Jest followUp was `["The resource list looks good, proceed with generating the plan.", "Go with recommended options. Assume all defaults to make the plan."]`. Without those confirmations, the planner stays in resource-gathering mode and never writes `.azure/infrastructure-plan.json`.
- `bicep-file-generation`: needed a third turn `"Looks good! Let's make Bicep now."` to produce Bicep files.

With followUps commented out, both `file-exists` graders will fail ~100% of the time because the skill hasn't been told to emit the artifacts.

If Vally supports multi-turn via `messages:` or `followUp:`, wire it in here. If it doesn't yet, these two response-quality stimuli aren't portable and should be marked pending rather than left as guaranteed failures.

Same shape appears in `azure-hosted-copilot-sdk/eval.yaml` (3 stimuli that relied on `setupExpressApp` / `setupCopilotSdkApp`) and `appinsights-instrumentation/eval.yaml` (Python/ASP.NET quality checks relied on `fs.cpSync` resource copying).

evals/azure-deploy/eval.yaml
```diff
@@ -0,0 +1,1281 @@
...
+            - azure-validate
+            - azure-prepare
+      # hasDeployLinks — checks for Azure deployment URLs
+      - type: output-matches
+        config:
+          pattern: "(?i)https?://[\\w.-]+\\.(azurewebsites\\.net|azurestaticapps\\.net|azurecontainerapps\\.io|web\\.core\\.windows\\.net)"
+      # expectFiles: bicep present, no .tf
+      - type: file-exists
+        config:
+          path: "infra/*.bicep"
```
jongio: The comment on line 158 says "bicep present, no .tf" and the description on line 166 says "no Terraform", but the only assertion here is `file-exists: infra/*.bicep`. A skill that produces both `infra/*.bicep` AND `infra/*.tf` passes this grader, which loses the intent of the original Jest `expectFiles` check.

Same issue in reverse for the Terraform stimuli starting at line 520 (`tf-swa-whiteboard`, etc.): they assert `infra/*.tf` exists but don't assert `infra/*.bicep` is absent.

Add a `file-not-exists` (or equivalent) grader for the opposite IaC flavor in each stimulus. If Vally doesn't ship that grader today, `file-exists` with a negated path or `output-not-matches` on a file listing can stand in. Worth fixing across the whole section, not just the first case.

evals/azure-resource-lookup/eval.yaml
```diff
@@ -0,0 +1,219 @@
...
+    prompt: "How many resources do I have by type across all subscriptions?"
+    tags:
+      type: integration
+      tier: full
+      cost: free
+      area: output
+    graders:
+      - type: output-contains
+        config:
+          substring: "resource"
```
jongio, line 58: `output-contains: "resource"` will match almost any Azure-related response, even an empty "I can't help with that" that happens to mention resources. The Jest `doesAssistantMessageIncludeKeyword` had the same weakness, but since you're rewriting the assertion anyway, this is a chance to make it meaningful.

Same pattern on the next stimulus (`output-contains: "location"`). Consider a phrase that actually requires the skill to do the lookup, e.g. `"across subscriptions"` or a regex covering `"resources by (type|location|subscription)"`. Otherwise the grader won't catch a regression where the skill silently stops doing cross-subscription queries.

evals/appinsights-instrumentation/eval.yaml
```diff
@@ -0,0 +1,164 @@
...
+      area: output
+    graders:
+      - type: output-contains
+        config:
+          # Jest checked: "App Insights" OR "Application Insights"
+          substring: "Insights"
+      - type: file-matches
+        config:
+          path: "**/*.py"
+          pattern: "configure_azure_monitor"
```
jongio, line 158: `file-matches` on `**/*.py` looking for `configure_azure_monitor` was originally checking seed Python files copied via `fs.cpSync("./appinsights-instrumentation/resources/python-app/", workspace)`. With `environment.files` commented out (line 141-144), there are no seed `.py` files in the workspace, so this grader now behaves one of two ways depending on Vally's semantics:

- If `file-matches` requires at least one matching path and returns false when none exist: it fails whenever the skill doesn't generate a `.py` file (which it often won't for an instrumentation-guidance prompt).
- If `file-matches` passes when no files match: it becomes a no-op that always passes.

Either way, the assertion no longer means "the seeded Python app has the instrumentation pattern added" (the original Jest intent). Same story for the ASP.NET Core stimulus above: "prompt is self-contained enough to trigger correct guidance" assumes the model doesn't regress, but removes the objective evidence the test was providing.

Skip these quality checks until the seed-data / `environment.files` story lands, or inline the resource content so the test is self-contained.

wbreza:
Closing in favor of new PR from org branch (not fork) so CI has access to repo secrets. See replacement PR below.

PR #1896, opened 2026-04-15
Add Deploy Scenario Retry Table to Dashboard

msalaman:
## Description

Added a table to the Dashboard that shows the number of retries per scenario, and flags cases where retries is >= 3
image

## Checklist

### DIFF
--- a/dashboard/api/src/functions/getTestResults.ts
+++ b/dashboard/api/src/functions/getTestResults.ts
@@ -135,16 +135,18 @@ function collectAgentMetadataPaths(
  * Counts tool.execution_start events for powershell/bash tools whose command
  * matches azd up, azd deploy, or terraform apply.
  *
- * @returns max(0, deployInvocations - 1)
+ * Returns null when the input cannot be parsed or contains no events array,
+ * so callers can distinguish invalid/unreadable files from a genuine 0-retry run.
+ * @returns max(0, deployInvocations - 1), or null for invalid input
  */
-function countDeployRetries(raw: string): number {
+function countDeployRetries(raw: string): number | null {
     let parsed: { events?: Array<{ type: string; data?: { toolName?: string; arguments?: { command?: string } } }> };
     try {
         parsed = JSON.parse(raw);
     } catch {
-        return 0;
+        return null;
     }
-    if (!Array.isArray(parsed.events)) return 0;
+    if (!Array.isArray(parsed.events)) return null;
 
     let deployCount = 0;
     for (const event of parsed.events) {
@@ -321,6 +323,7 @@ async function getTestResults(request: HttpRequest, context: InvocationContext):
             retryFetchTasks.push(
                 getBlobContent(blobPath).then((raw) => {
                     const retries = countDeployRetries(raw);
+                    if (retries === null) return; // invalid/unreadable — don't skew the average
                     deployRetryTotals.set(
                         testCaseName,
                         (deployRetryTotals.get(testCaseName) ?? 0) + retries,
--- a/dashboard/api/src/functions/getTestResults.ts
+++ b/dashboard/api/src/functions/getTestResults.ts
@@ -313,22 +313,34 @@ async function getTestResults(request: HttpRequest, context: InvocationContext):
     await Promise.all(reportFetchTasks);
 
     // Fetch agent-metadata.json files for azure-deploy scenario retry counts
-    const deployRetryCounts = new Map<string, number>();
+    const deployRetryTotals = new Map<string, number>();
+    const deployRetryRunCounts = new Map<string, number>();
     const retryFetchTasks: Promise<void>[] = [];
     for (const [testCaseName, paths] of agentMetadataPathsByTestCase) {
         for (const blobPath of paths) {
             retryFetchTasks.push(
                 getBlobContent(blobPath).then((raw) => {
                     const retries = countDeployRetries(raw);
-                    deployRetryCounts.set(
+                    deployRetryTotals.set(
                         testCaseName,
-                        (deployRetryCounts.get(testCaseName) ?? 0) + retries,
+                        (deployRetryTotals.get(testCaseName) ?? 0) + retries,
+                    );
+                    deployRetryRunCounts.set(
+                        testCaseName,
+                        (deployRetryRunCounts.get(testCaseName) ?? 0) + 1,
                     );
                 }).catch(() => { /* skip unreadable files */ }),
             );
         }
     }
     await Promise.all(retryFetchTasks);
+
+    // Average retries per run to avoid inflating counts when a scenario fires across multiple runs
+    const deployRetryCounts = new Map<string, number>();
+    for (const [testCaseName, total] of deployRetryTotals) {
+        const runCount = deployRetryRunCounts.get(testCaseName) ?? 1;
+        deployRetryCounts.set(testCaseName, total / runCount);
+    }
 
     // Compute statistics per skill
     const skillTestResults: SkillTestResults = {};
--- a/dashboard/api/src/functions/getTestResults.ts
+++ b/dashboard/api/src/functions/getTestResults.ts
@@ -32,6 +32,13 @@ export interface SkillStats {
     passedTests: TestCase[];
     /** Average Confidence extracted from SKILL-REPORT.md files (0–100), or null if not available. */
     averageConfidence: number | null;
+    /**
+     * Maps sanitised test-case directory names to the number of agent runs recorded
+     * for that test case (i.e. the number of agent-metadata-*.md files found).
+     * Populated only for the azure-deploy skill; skill-invocation tests are excluded.
+     * A value greater than 1 means the test case required retries.
+     */
+    scenarioAgentRunCounts?: Record<string, number>;
 }
 
 export type SkillTestResults = Record<string, SkillStats>;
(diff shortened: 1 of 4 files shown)


dashboard/api/src/functions/getTestResults.ts
```diff
@@ -79,6 +86,48 @@ function collectSkillReportPaths(
...
+ *  Level 1 – testGroup directory (e.g. "vanilla-static-web-apps-deploy")
+ *  Level 2 – testCase directory  (e.g. "azure-deploy_-_Integration_Tests_...")
+ *
+ * The "skill-invocation" test group is excluded because only scenario tests
+ * are relevant for retry tracking.
+ *
+ * @param skillNode  The BlobTreeNode for the skill (e.g. azure-deploy)
+ * @returns Map of testCaseDirName → agent run count
+ */
+function collectScenarioAgentRunCounts(skillNode: BlobTreeNode): Map<string, number> {
```
JasonYeMSFT: The number of agent runs (aka. number of agent-metadata-.md is not the right indicator of deployment retry. A deployment retry happens in the same agent run of the failed attempt. For example, in one agent run, it may run azd up and fail, modify some files, and run azd up again. This is what we would like to track in this table instead.

JasonYeMSFT (changes requested):

dashboard/api/src/functions/getTestResults.ts
JasonYeMSFT: Btw way, each test case should have one and only one agent run, we should no longer having multiple agent runs since the 2nd one is redundant.

dashboard/api/src/functions/getTestResults.ts
```diff
@@ -32,6 +32,13 @@ export interface SkillStats {
     passedTests: TestCase[];
     /** Average Confidence extracted from SKILL-REPORT.md files (0–100), or null if not available. */
     averageConfidence: number | null;
+    /**
+     * Maps sanitised test-case directory names to the number of agent runs recorded
+     * for that test case (i.e. the number of agent-metadata-*.md files found).
+     * Populated only for the azure-deploy skill; skill-invocation tests are excluded.
```
jongio, line 40: This field only gets populated for `azure-deploy` but lives on the shared `SkillStats` interface that every skill returns. That makes the API contract confusing for consumers - they'd need to know which skill names produce this field. If you revive this after fixing the metric, consider either a separate endpoint (e.g. `/api/deploy-retries/{date}`) or a top-level response field outside the per-skill stats object.

dashboard/api/src/functions/getTestResults.ts
```diff
@@ -178,11 +227,27 @@ async function getTestResults(request: HttpRequest, context: InvocationContext):
...
+            if (skillName === "azure-deploy") {
+                const runCounts = collectScenarioAgentRunCounts(skillNode);
+                if (runCounts.size > 0) {
+                    if (!agentRunCountsBySkill.has(skillName)) {
+                        agentRunCountsBySkill.set(skillName, new Map());
+                    }
+                    const existing = agentRunCountsBySkill.get(skillName)!;
+                    for (const [testCase, count] of runCounts) {
+                        existing.set(testCase, (existing.get(testCase) ?? 0) + count);
+                    }
```
jongio: The counts are summed across every `runNode` for the date. If CI re-runs the same scenario (manual re-trigger, flaky-test retry), the numbers get inflated since both runs contribute. Even after the metric itself is corrected, you'll want either per-run counts or deduplication here - naive summing conflates separate test executions.

dashboard/assets/dashboard.js
```diff
@@ -1634,6 +1634,153 @@ function renderConfidenceLevelPanel(
...
+}
+
+/**
+ * Populate and finalise the deploy scenario retries panel.
+ * @param {HTMLElement} section
+ * @param {Array<{label:string, runs:number, retries:number}>} rows
+ * @param {string} overallStatus - pass | warn | skip
+ * @param {string|null} dateLabel
+ */
+function renderDeployRetriesPanel(section, rows, overallStatus, dateLabel) {
```
jongio, line 1684: This silently swallows all errors (network failures, bad JSON, unexpected response shape) and renders `No data`. I know other panels do the same thing, but a `console.warn("Deploy retries panel failed", err)` would cost nothing and save debugging time when the API is misbehaving.

dashboard/assets/dashboard.js
```diff
@@ -1634,6 +1634,153 @@ function renderConfidenceLevelPanel(
...
+  const summaryEl = section.querySelector(".panel-summary");
+  const itemsEl = section.querySelector(".panel-items");
+  if (!summaryEl || !itemsEl) return;
+
+  summaryEl.textContent = "";
+  itemsEl.textContent = "";
+
+  const total = rows.length;
+  const failing = rows.filter(function (r) { return r.retries >= 3; }).length;
+  const warning = rows.filter(function (r) { return r.retries > 0 && r.retries < 3; }).length;
```
jongio, line 1694: The JSDoc says `overallStatus` can be `pass | warn | skip`, but the caller also passes `"fail"` (when `retries >= 3`). Minor, but worth fixing when you're back in this code.

jongio (commented):
A few things beyond what Jason flagged - mostly relevant for when the metric is reworked. The SkillStats interface shape and cross-run aggregation strategy are worth thinking about before the next iteration.

jongio (commented):
The metric rework addresses the core concern. Counting deploy invocations per session from `tool.execution_start` events matches what Jason described, and the JSDoc/field renames clean up the earlier nits.

One carryover from the previous version: cross-run aggregation at `getTestResults.ts` lines 264-270 still sums retries across every `runNode` for the date. If a scenario fires in multiple runs on the same date (manual retrigger, or scheduled + push), the displayed retry count will be the sum across those runs rather than per-run. My earlier inline comment on `scenarioAgentRunCounts` applies equally to `scenarioDeployRetryCounts`.

dashboard/api/src/functions/getTestResults.ts
msalaman: Updated logic to count retries per run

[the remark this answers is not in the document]
msalaman: updated logic to count retries per run

dashboard/assets/dashboard.js
```diff
@@ -1634,6 +1634,153 @@ function renderConfidenceLevelPanel(
...
+      return { label, runs: /** @type {number} */ (runs), retries: /** @type {number} */ (runs) - 1 };
+    });
+
+    rows.sort(function (a, b) {
+      return b.retries - a.retries || a.label.localeCompare(b.label);
+    });
+
+    const hasFailing = rows.some(function (r) { return r.retries >= 3; });
+    const hasWarning = rows.some(function (r) { return r.retries > 0 && r.retries < 3; });
+    const overallStatus = rows.length === 0 ? "skip" : hasFailing ? "fail" : hasWarning ? "warn" : "pass";
```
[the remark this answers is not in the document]
msalaman: logic has been fixed

dashboard/api/src/functions/getTestResults.ts
msalaman: Updated logic to properly count retries instead of using agent runs

dashboard/assets/dashboard.js
msalaman: Fixed

dashboard/assets/dashboard.js
```diff
@@ -1634,6 +1634,153 @@ function renderConfidenceLevelPanel(
...
+  } catch {
+    renderDeployRetriesPanel(section, [], "skip", null);
+  }
+}
+
+/**
+ * Populate and finalise the deploy scenario retries panel.
+ * @param {HTMLElement} section
+ * @param {Array<{label:string, runs:number, retries:number}>} rows
+ * @param {string} overallStatus - pass | warn | skip
```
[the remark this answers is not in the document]
msalaman: Fixed

dashboard/api/src/functions/getTestResults.ts
```diff
@@ -226,6 +312,36 @@ async function getTestResults(request: HttpRequest, context: InvocationContext):
...
+    }
+    await Promise.all(retryFetchTasks);
+
+    // Average retries per run to avoid inflating counts when a scenario fires across multiple runs
+    const deployRetryCounts = new Map<string, number>();
+    for (const [testCaseName, total] of deployRetryTotals) {
+        const runCount = deployRetryRunCounts.get(testCaseName) ?? 1;
+        deployRetryCounts.set(testCaseName, total / runCount);
+    }
+
```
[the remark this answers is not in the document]
msalaman: Logic is fine to keep as is because there are some scenarios run multiple times within a nightly run (for example: terraform-app-service-deploy creates todo list with frontend and API using Terraform)

dashboard/assets/dashboard.js
```diff
@@ -1634,6 +1634,147 @@ function renderConfidenceLevelPanel(
...
+      const tdName = el("td", "deploy-retries-name", row.label);
+      const tdRetries = el("td", "deploy-retries-num-col deploy-retries-count");
+      tdRetries.textContent = String(row.retries);
+      if (row.retries >= 3) {
+        tdRetries.classList.add("deploy-retries-fail");
+      } else if (row.retries > 0) {
+        tdRetries.classList.add("deploy-retries-nonzero");
+      } else {
+        tdRetries.classList.add("deploy-retries-zero");
+      }
```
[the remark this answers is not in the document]
msalaman: It's okay to have decimals in this scenario

dashboard/api/src/functions/getTestResults.ts
```diff
@@ -79,6 +87,78 @@ function collectSkillReportPaths(
...
+        if (event.type !== "tool.execution_start") continue;
+        const toolName = event.data?.toolName;
+        if (toolName !== "powershell" && toolName !== "bash") continue;
+        const command = event.data?.arguments?.command ?? "";
+        if (DEPLOY_COMMAND_PATTERN.test(command)) {
+            deployCount++;
+        }
+    }
+    return Math.max(0, deployCount - 1);
+}
```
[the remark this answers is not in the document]
msalaman: Fixed

JasonYeMSFT (approved):

PR #1931, opened 2026-04-17
Enhancements for benchmarks report generation pipeline

fanyang-mono:
## Description

Key improvements:
- Remove the dependency of getting build artifacts from benchmark run pipeline
- Instead get dates to process and run ids to process from storage account
- Upload msbench run reports to storage account
- Upload key metrics to storage account: resolved or not, total consumed tokens, total steps


pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -183,15 +200,18 @@
...
     $reportsDir = Join-Path $targetDir "reports"
     $reportFiles = @()
-    $reportsDirMdFiles = Get-ChildItem -Path $reportsDir -Filter '*.md' -ErrorAction SilentlyContinue
+    $reportsDirMdFiles = Get-ChildItem -Path $reportsDir -Filter '*.md'
     if ((Test-Path $reportsDir) -and $reportsDirMdFiles) {
         $reportFiles = $reportsDirMdFiles
         Write-Host "Found $($reportFiles.Count) report(s) in $reportsDir"
     } else {
-        $reportFiles = Get-ChildItem -Path $targetDir -Filter '*.md' -ErrorAction SilentlyContinue | Where-Object { $_.FullName -notlike "*\.github\*" -and $_.FullName -notlike "*/.github/*" }
+        $reportFiles = Get-ChildItem -Path $targetDir -Filter '*.md' | Where-Object { $_.FullName -notlike "*\.github\*" -and $_.FullName -notlike "*/.github/*" }
```
jongio: The fallback `Get-ChildItem -Path $targetDir -Filter '*.md'` on this line also lost its `-ErrorAction SilentlyContinue` (same regression the bot flagged for line 209). Worth fixing both in one pass so the else branch doesn't throw when `` has no matches under `ErrorActionPreference = Stop`.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -211,5 +231,172 @@
...
+            if ($runIdsFileInfo) {
+                $Date = $runIdsFileInfo.LastWriteTimeUtc.ToString("yyyy-MM-dd")
+                Write-Host "Using benchmark run date from run_ids.json timestamp: $Date"
+            } else {
+                $Date = Get-Date -Format "yyyy-MM-dd"
+                Write-Warning "Could not determine benchmark run date, falling back to today's date: $Date"
+            }
+        }
+
+        function Get-BenchmarkInstance {
```
jongio: `Get-BenchmarkInstance` drops the ordered `` priority match that the deleted `Get-SkillName` used. For filenames matching `msbench_(analysis_)?report_<x>_YYYY_MM_DD` this is fine, but reports that don't match that pattern now route by raw baseName instead of the nearest known skill. Is the filename convention strict enough that this is safe, or should the category list be preserved?

jongio (commented):
Bot covers the big ones. Two smaller follow-ups inline.

pipelines/azure-benchmark-report.yml
```diff
@@ -81,19 +81,9 @@ stages:
         scriptArguments: >
           -InputPath "$(Pipeline.Workspace)/benchmark-run-ids"
```
jongio: **Critical, pipeline-breaking.** This line still passes `-InputPath` but the script's `param()` block removed that parameter in this PR. PowerShell will throw `A parameter cannot be found that matches parameter name 'InputPath'` and the AzureCLI task will fail before the script runs. Verified in pwsh 7 locally.

Fix: drop the `-InputPath` line from `scriptArguments` (it isn't read anywhere anymore).

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +139,287 @@
...
+            foreach ($runResultsDir in $runResultsDirs) {
+                $runDirs = Get-ChildItem -Path $runResultsDir.FullName -Directory -Filter '*_results' -ErrorAction SilentlyContinue
+                Write-Host "Processing run results directory: $($runResultsDir.FullName)"
+                foreach ($runDir in $runDirs) {
+                    # Read model name and run ID from run_metadata.json
+                    $metadataFile = Join-Path $runDir.FullName 'run_metadata.json'
+                    $model = $null
+                    if (Test-Path $metadataFile) {
+                        $metadata = Get-Content -Path $metadataFile -Raw | ConvertFrom-Json
+                        $model = $metadata.model -replace '-autodev-test$', ''
```
jongio: With `Set-StrictMode -Version Latest` (set on line 49), `$metadata.model` throws `The property 'model' cannot be found on this object` when `run_metadata.json` exists but doesn't contain a `model` field. Then the whole date iteration aborts.

Guard before access, e.g.:

```powershell
if ($metadata.PSObject.Properties.Name -contains 'model') {
    $model = $metadata.model -replace '-autodev-test$', ''
}
```

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +139,287 @@
...
+                        $evalContent | Add-Member -NotePropertyName 'instance_id' -NotePropertyValue $instanceName -Force
+
+                        # Enrich with resolved field from eval.json in the same directory
+                        $evalJsonFile = Join-Path $evalReport.DirectoryName 'eval.json'
+                        if (Test-Path $evalJsonFile) {
+                            $evalJson = Get-Content -Path $evalJsonFile -Raw | ConvertFrom-Json
+                            $resolvedValue = $null
+                            # eval.json structure: { "instance_name": { "resolved": true/false } }
+                            foreach ($prop in $evalJson.PSObject.Properties) {
+                                if ($null -ne $prop.Value.resolved) {
```
jongio: Same StrictMode issue - the `$null -ne $prop.Value.resolved` check itself throws if `$prop.Value` is a string or a nested object without a `resolved` property. Once it throws, eval enrichment stops for the rest of the run.

Check the property exists before dereferencing:

```powershell
if ($prop.Value -is [psobject] -and $prop.Value.PSObject.Properties.Name -contains 'resolved') {
    $resolvedValue = $prop.Value.resolved
    break
}
```

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -27,39 +36,21 @@
 #>
 
     param(
-        [Parameter(Mandatory=$true)][ValidateNotNullOrEmpty()][string]$InputPath,
-        [Parameter(Mandatory=$true)][ValidateNotNullOrEmpty()][string]$OutputPath
+        [Parameter(Mandatory=$true)][ValidateNotNullOrEmpty()][string]$OutputPath,
+        [Parameter(Mandatory=$false)][string]$StorageAccountName,
+        [Parameter(Mandatory=$false)][string]$ContainerName
```
jongio: `StorageAccountName` and `ContainerName` are declared `Mandatory=$false` but every code path below uses them unconditionally - ToBeProcessed download, run_ids download, both blob uploads. If they're missing, `az storage blob download` fails with a cryptic CLI error rather than a clean binding error at the boundary.

Either make both mandatory, or wrap the entire blob flow in an `if ($StorageAccountName -and $ContainerName)` guard so the local-only path still works.

jongio (commented):
Follow-up after the bot's pass and my earlier one. A few more things on a closer read - one of them will break the pipeline on the next run.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +139,287 @@
...
+    }
+
+    # --- Process each date ---
+    foreach ($date in $dateRunIdMap.Keys) {
+        $inputRunIds = @($dateRunIdMap[$date])
+        Write-Host "`n=========================================="
+        Write-Host "Processing date: $date with $($inputRunIds.Count) run ID(s): $($inputRunIds -join ',')"
+        Write-Host "==========================================`n"
+
+        $dateOutputPath = Join-Path $OutputPath $date
```
jongio, line 244: `$date` comes straight from the `ToBeProcessed` blob and is used as a path component here (and as a blob prefix below). A malformed entry (`..`, an absolute path, invalid filename chars) could land files outside `$OutputPath` or produce broken blob paths without a clear error.

Validate format when parsing the file - something like:

```powershell
$dates = @(Get-Content $toBeProcessedLocal |
    ForEach-Object { $_.Trim() } |
    Where-Object { $_ -match '^\d{4}-\d{2}-\d{2}$' })
```

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +139,287 @@
...
+    # --- Get dates from ToBeProcessed file in blob container root ---
+    Write-Host "Checking for ToBeProcessed file in blob container $ContainerName"
+    $toBeProcessedLocal = Join-Path $OutputPath 'ToBeProcessed'
+
+    az storage blob download `
+        --account-name $StorageAccountName `
+        --container-name $ContainerName `
+        --name "ToBeProcessed" `
+        --file $toBeProcessedLocal `
+        --auth-mode login 2>$null
```
jongio: Redirecting all stderr to null here means genuine failures (auth denied, wrong container, network error) look identical to the legitimate "no ToBeProcessed blob, nothing to do" case - the script silently exits 0 and the pipeline reports green.

Consider capturing stderr and only treating `BlobNotFound` as the no-op case, e.g. `2>&1 | Tee-Object -Variable dlErr` then branch on whether `$dlErr -match 'BlobNotFound'`.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -183,15 +200,18 @@
...
     $reportFiles = @()
-    $reportsDirMdFiles = Get-ChildItem -Path $reportsDir -Filter '*.md' -ErrorAction SilentlyContinue
+    $reportsDirMdFiles = Get-ChildItem -Path $reportsDir -Filter '*.md'
     if ((Test-Path $reportsDir) -and $reportsDirMdFiles) {
         $reportFiles = $reportsDirMdFiles
         Write-Host "Found $($reportFiles.Count) report(s) in $reportsDir"
     } else {
-        $reportFiles = Get-ChildItem -Path $targetDir -Filter '*.md' -ErrorAction SilentlyContinue | Where-Object { $_.FullName -notlike "*\.github\*" -and $_.FullName -notlike "*/.github/*" }
+        $reportFiles = Get-ChildItem -Path $targetDir -Filter '*.md' | Where-Object { $_.FullName -notlike "*\.github\*" -and $_.FullName -notlike "*/.github/*" }
         Write-Host "Found $($reportFiles.Count) report(s) in $targetDir (excluding .github folders)"
```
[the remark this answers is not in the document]
fanyang-mono: Fixed.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -211,5 +231,172 @@
...
+                    $outputDirName = $evalReport.Directory.Parent.Name  # e.g. azure.eval.x86_64.aca_deployment_skill-output
+                    $instanceName = $outputDirName -replace '^azure\.eval\.x86_64\.', '' -replace '-output$', ''
+
+                    # Enrich eval_report.json with model and run metadata
+                    $evalContent = Get-Content -Path $evalReport.FullName -Raw | ConvertFrom-Json
+                    if ($model) { $evalContent | Add-Member -NotePropertyName 'model' -NotePropertyValue $model -Force }
+                    $evalContent | Add-Member -NotePropertyName 'instance_id' -NotePropertyValue $instanceName -Force
+
+                    $destination = Join-Path $OutputPath "${model}_${instanceName}_eval_report.json"
+                    $evalContent | ConvertTo-Json -Depth 10 | Set-Content -Path $destination -Encoding UTF8
```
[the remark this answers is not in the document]
fanyang-mono: Added a check for if run_metadata.json file exits.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -13,12 +14,25 @@
     The script reads run IDs from run_ids.json in the InputPath, generates reports using
     GitHub Copilot CLI, and saves the generated markdown reports to the OutputPath.
 
+    When StorageAccountName and ContainerName are provided, the generated reports are
+    uploaded to Azure Blob Storage organized by date, skill name, and benchmark instance.
```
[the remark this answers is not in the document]
fanyang-mono: Comments have been updated. This is no longer relevant.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
fanyang-mono: I prefer having it error out when it ran into issues finding md files.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
fanyang-mono: The expectation is that the filename follows that convention. Otherwise, I won't know what it could be ahead of time.

pipelines/azure-benchmark-report.yml
fanyang-mono: This is outdated.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
fanyang-mono: Added guard.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
fanyang-mono: Updated

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
fanyang-mono: Updated.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
fanyang-mono: Updated.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
fanyang-mono: Fixed.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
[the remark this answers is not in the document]
fanyang-mono: no longer relevant

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +155,355 @@
...
-    $copilotLogDir = Join-Path $OutputPath "copilot_log"
-    $copilotLogFile = Join-Path $copilotLogDir "copilot_log.md"
-    New-Item -Path $copilotLogDir -ItemType Directory -Force | Out-Null
-    $copilotArgs = @(
-            "-p", $reportGenerationPrompt,
-            "--model", "claude-opus-4.6",
-            "--share", $copilotLogFile,
-            "--yolo"
+    # --- Get dates from ToBeProcessed file in blob container root ---
+    Write-Host "Downloading ToBeProcessed file from blob container $ContainerName"
```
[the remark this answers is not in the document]
fanyang-mono: Added check.

pipelines/azure-benchmark-report.yml
```diff
@@ -79,21 +52,10 @@ stages:
         scriptLocation: 'scriptPath'
         scriptPath: $(Build.SourcesDirectory)/pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
         scriptArguments: >
-          -InputPath "$(Pipeline.Workspace)/benchmark-run-ids"
           -OutputPath "$(Build.ArtifactStagingDirectory)"
+          -StorageAccountName "msbenchnightlydata"
+          -ContainerName "msbench-reports"
```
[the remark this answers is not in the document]
fanyang-mono: It is happening in another PR.

JasonYeMSFT (dismissed):

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +156,355 @@
...
+        "--account-name", $StorageAccountName,
+        "--container-name", $ContainerName,
+        "--name", "ToBeProcessed",
+        "--file", $toBeProcessedLocal,
+        "--auth-mode", "login"
+    )
+    az @azArgs
+    if ($LASTEXITCODE -ne 0 -or !(Test-Path $toBeProcessedLocal)) {
+        throw "Failed to download ToBeProcessed file from blob container $ContainerName. Ensure the file exists and the service principal has access."
+    }
```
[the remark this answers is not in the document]
fanyang-mono: Updated comments.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +156,355 @@
...
+        Write-Host "Copying generated report files to output path: $dateOutputPath"
+        $reportsDir = Join-Path $targetDir "reports"
+        $reportFiles = @()
+        if ((Test-Path $reportsDir)) {
+            $reportsDirMdFiles = Get-ChildItem -Path $reportsDir -Filter '*.md'
+            if ($reportsDirMdFiles) {
+                $reportFiles = $reportsDirMdFiles
+                Write-Host "Found $($reportFiles.Count) report(s) in $reportsDir"
+            }
+        } else {
```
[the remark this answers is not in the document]
fanyang-mono: Fixed.

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +156,355 @@
...
+                    Write-Warning "eval.json not found in $outputDir for instance '$instanceName'. Skipping."
+                    continue
+                }
+                $evalJson = Get-Content -Path $evalJsonFile -Raw | ConvertFrom-Json
+                $resolvedValue = $null
+                # eval.json structure: { "instance_name": { "resolved": true/false } }
+                foreach ($prop in $evalJson.PSObject.Properties) {
+                    if ($prop.Value -is [psobject] -and $prop.Value.PSObject.Properties.Name -contains 'resolved') {
+                        $resolvedValue = $prop.Value.resolved
+                        break
```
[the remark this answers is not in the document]
fanyang-mono: There is only one instance in each eval.json

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +156,355 @@
...
+                    Write-Warning "eval.json found but no 'resolved' field for instance '$instanceName'"
+                }
+
+                $destination = Join-Path $dateOutputPath "${model}_${instanceName}_eval_report.json"
+                $evalContent | ConvertTo-Json -Depth 10 | Set-Content -Path $destination -Encoding UTF8
+                Write-Host "Saved enriched eval_report.json for instance '$instanceName' (model: $model) to $destination"
+
+                # Upload eval_report.json to Azure Blob Storage
+                $evalBlobPath = "$date/$instanceName/${model}_${instanceName}_eval_report.json"
+                Write-Host "Uploading ${model}_${instanceName}_eval_report.json -> $ContainerName/$evalBlobPath"
```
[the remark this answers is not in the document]
fanyang-mono: There is only one run for a particular model+instance on a date

pipelines/scripts/Invoke-GenerateBenchmarkReport.ps1
```diff
@@ -161,55 +156,355 @@
...
+                $evalReportFile = Join-Path $outputDir 'eval_report.json'
+                if (!(Test-Path $evalReportFile)) {
+                    Write-Warning "eval_report.json not found in $outputDir for instance '$instanceName'. Skipping."
+                    continue
+                }
+                # Enrich eval_report.json with model and instance metadata
+                $evalContent = Get-Content -Path $evalReportFile -Raw | ConvertFrom-Json
+                $evalContent | Add-Member -NotePropertyName 'model' -NotePropertyValue $model -Force
+                $evalContent | Add-Member -NotePropertyName 'instance_id' -NotePropertyValue $instanceName -Force
+
```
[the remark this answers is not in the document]
fanyang-mono: They are already in eval_report.json.

JasonYeMSFT (approved):

PR #1941, opened 2026-04-17
Add performance dashboard

### LINKED ISSUE #1875: Add the MSBench tests to the dashboard for visibility
Add the MSBench tests to the dashboard for visibility

fanyang-mono:
## Description

Add two new tabs to the dashboards:
- Performance dashboard: display key metrics from nightly runs
- MSBench Nightly Runs: display reports generated from nightly runs

Fixes #1875


dashboard/api/src/msbenchBlobEnumerator.ts
```diff
@@ -0,0 +1,99 @@
+import { BlobServiceClient } from "@azure/storage-blob";
+import { AzureCliCredential, ManagedIdentityCredential } from "@azure/identity";
+import { BlobEntry, BlobTree, BlobTreeNode } from "./blobEnumerator";
```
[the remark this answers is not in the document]
fanyang-mono: removed

dashboard/performance-dashboard.html
```diff
@@ -0,0 +1,32 @@
...
+</head>
+
+<body>
+    <a href="#main" class="skip-link">Skip to content</a>
+
+    <nav class="site-nav" role="navigation" aria-label="Main navigation">
+        <a href="/" class="site-nav-link">Dashboard</a>
+        <a href="/integration-tests.html" class="site-nav-link">Integration Tests</a>
+        <a href="/nightly-runs.html" class="site-nav-link">Nightly Runs</a>
+        <a href="/performance-dashboard.html" class="site-nav-link active" aria-current="page">Performance Dashboard</a>
```
JasonYeMSFT: Should this view be called "MSBench Performance Dashboard"?

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,273 @@
+import { app, HttpRequest, HttpResponseInit, InvocationContext, Timer } from "@azure/functions";
+import { TableClient } from "@azure/data-tables";
+import { AzureCliCredential, ManagedIdentityCredential } from "@azure/identity";
+import { listMsbenchDates, enumerateMsbenchBlobs, getMsbenchBlobContent } from "../msbenchBlobEnumerator";
+import { BlobTreeNode } from "../blobEnumerator";
+import { logRequestIdentity } from "../requestIdentity";
+
+const MSBENCH_STORAGE_ACCOUNT = "msbenchnightlydata";
```
JasonYeMSFT: You should add an "existing" resource to the bicep template, assign an env var for it and access it via the env var in the code. This will make the Azure resource dependency more visible.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,273 @@
...
+const EVAL_TABLE_NAME = "msbenchevalmetrics";
```
JasonYeMSFT: Same for this table name.

dashboard/src/performance-dashboard/App.tsx
```diff
@@ -0,0 +1,393 @@
...
+            .then((res) => {
+                if (!res.ok) throw new Error(`API error: ${res.status}`);
+                return res.json();
+            })
+            .then((data: EvalMetricRow[]) => setMetrics(data))
+            .catch((err) => setError(err.message))
+            .finally(() => setLoading(false));
+    };
+
+    const handleSync = async (force: boolean) => {
```
JasonYeMSFT: I don't think we need to expose "write" operation via the dashboard. A read-only dashboard makes it easier to reason data state and simplifies code maintenance. 

This also requires the Function app's managed identity to have write permissions which leads to a more comoplex security model.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,273 @@
...
+        if (entity.partitionKey) dates.add(entity.partitionKey);
+    }
+    return dates;
+}
+
+/**
+ * Core sync logic: scan blob storage for new eval reports and upsert into table.
+ * Returns { synced, dates } or { synced: 0, message } if nothing new.
+ */
+async function runSync(context: InvocationContext, force = false): Promise<{ synced: number; dates?: string[]; message?: string }> {
```
JasonYeMSFT: We can use a separate Function App isolated from website to simplify the security model. Ideally the website can have no write permission on its data. The separate Function App will have write permission on the data, but won't be accessible from the website. If you like this idea, you can add the new function app to IaC and configure it to run with a scheduled trigger.

dashboard/api/src/functions/downloadMsbenchBlob.ts
```diff
@@ -0,0 +1,45 @@
...
+async function downloadMsbenchBlob(request: HttpRequest, context: InvocationContext): Promise<HttpResponseInit> {
+    logRequestIdentity(request, context, "downloadMsbenchBlob");
+
+    const blobPath = request.query.get("path");
+    if (!blobPath) {
+        return { status: 400, body: "Missing 'path' query parameter" };
+    }
+
+    // Prevent directory traversal
+    if (blobPath.includes("..")) {
```
jongio, line 18: The `..` check stops directory traversal but doesn't constrain the blob to anything. Combined with `authLevel: "anonymous"` on line 42, anyone who can hit the dashboard URL can read any blob in `msbench-reports` by guessing or enumerating paths. The existing `downloadBlob.ts` has the same shape, so this is following the convention. If msbench data has any sensitivity that the integration-reports container doesn't, scoping the path to a known date prefix (or whatever the front-end actually requests) would tighten this up. Same direction as @JasonYeMSFT's broader point about isolating read and write permissions.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,273 @@
...
+    const tableClient = getEvalTableClient();
+    const blobDates = await listMsbenchDates();
+
+    let datesToProcess: string[];
+    if (force) {
+        datesToProcess = blobDates;
+    } else {
+        const processedDates = await getProcessedDates(tableClient);
+        datesToProcess = blobDates.filter((d) => !processedDates.has(d));
```
jongio: Partial-failure idempotency: `getProcessedDates()` treats any partition with at least one entity as "processed", so the unforced sync filters that date out. If a date's `Promise.all` upserts even one report and then throws (network blip, JSON parse failure on a later report), the date is permanently skipped on subsequent non-force runs. Worth either (a) scoping the "processed" check to a marker entity per date that's only written after all reports succeed, or (b) tracking expected vs actual report counts per partition so a partial date gets retried.

dashboard/api/src/msbenchBlobEnumerator.ts
```diff
@@ -0,0 +1,99 @@
...
+const MSBENCH_STORAGE_ACCOUNT = "msbenchnightlydata";
+const MSBENCH_REPORTS_CONTAINER_NAME = "msbench-reports";
+
+const EXCLUDED_FILENAMES = new Set(["token-usage.json", "agent-metadata.json"]);
+
+function getContainerClient() {
+    const clientId = process.env.AZURE_CLIENT_ID;
+    const isDevEnvironment = process.env.AZURE_FUNCTIONS_ENVIRONMENT === "Development";
+    const credential = isDevEnvironment ? new AzureCliCredential() : new ManagedIdentityCredential(clientId!);
+    const blobServiceClient = new BlobServiceClient(
```
jongio, line 14: This file duplicates `getContainerClient`, `isExcluded`, `createNode`, the date listing, the tree builder, and the blob fetcher from `blobEnumerator.ts` with only the storage account and container name swapped. If you take @JasonYeMSFT's suggestion to plumb the account through env vars and bicep, the natural follow-on is to widen the existing `getContainerClient(containerName)` to also take an account (or extract `createContainerClient(account, container)`) and call the existing `enumerateBlobs(prefix)` / `getBlobContent(path)` from a thin wrapper. That collapses about 80 lines and keeps both containers on a single code path.

jongio (commented):
Most of what I'd raise is already covered by the existing inline comments (hardcoded storage account, OData injection, the timer cron mismatch, the anonymous write endpoint, the read-only dashboard suggestion, separating into its own Function App). Three additional things worth a look that I didn't see flagged.

dashboard/api/src/functions/downloadMsbenchBlob.ts
JasonYeMSFT: authLevel: anonymous is a little misleading here. When the function app is configured as the backend of a static web app, it won't allow access other than those from the static web app. As of today, people can still access it via the static web app though. Please make sure the binary data doesn't contain sensitive information, like secrets, internal only information, etc.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,273 @@
...
+ * Check which dates already have entries in the table.
+ */
+async function getProcessedDates(tableClient: TableClient): Promise<Set<string>> {
+    const dates = new Set<string>();
+    for await (const entity of tableClient.listEntities({
+        queryOptions: { select: ["partitionKey"] },
+    })) {
+        if (entity.partitionKey) dates.add(entity.partitionKey);
+    }
+    return dates;
```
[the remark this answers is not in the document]
fanyang-mono: This is okay for now.

dashboard/src/performance-dashboard/App.tsx
```diff
@@ -0,0 +1,393 @@
...
+        const byDate: Record<string, Record<string, number>> = {};
+        for (const row of metrics) {
+            if (!byDate[row.date]) byDate[row.date] = {};
+            const key = seriesKey(row);
+            byDate[row.date][key] = (byDate[row.date][key] || 0) + (Number(row.totalSteps) || 0);
+        }
+        return Object.entries(byDate)
+            .map(([date, series]) => ({ date, ...series }))
+            .sort((a, b) => a.date.localeCompare(b.date));
+    }, [metrics]);
```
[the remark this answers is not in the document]
fanyang-mono: Added

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,273 @@
...
+            status: 500,
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({ error: err.message }),
+        };
+    }
+}
+
+app.http("syncMsbenchEvalMetrics", {
+    methods: ["POST"],
+    authLevel: "anonymous",
```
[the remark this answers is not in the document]
fanyang-mono: All the functions uses anonymous authLevel for dashboard. I will keep it in sync.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,273 @@
...
+        return {
+            status: 500,
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({ error: err.message }),
+        };
+    }
+}
+
+/**
+ * Timer-triggered daily sync (runs at midnight UTC).
```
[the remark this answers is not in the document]
fanyang-mono: Fixed.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,273 @@
...
+    const filterResolved = request.query.get("resolved");
+
+    try {
+        const tableClient = getEvalTableClient();
+
+        const filters: string[] = [];
+        if (filterBenchmark) filters.push(`benchmark eq '${filterBenchmark}'`);
+        if (filterModel) filters.push(`model eq '${filterModel}'`);
+        if (filterResolved === "1" || filterResolved === "0") filters.push(`resolved eq ${Number(filterResolved)}`);
+        const filter = filters.length > 0 ? filters.join(" and ") : undefined;
```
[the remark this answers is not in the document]
fanyang-mono: This is okay. Benchmark name won't contain single quotes.

dashboard/performance-dashboard.html
fanyang-mono: Renamed.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
fanyang-mono: Fixed.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
fanyang-mono: Fixed.

dashboard/api/src/msbenchBlobEnumerator.ts
```diff
@@ -0,0 +1,99 @@
...
+        `https://${MSBENCH_STORAGE_ACCOUNT}.blob.core.windows.net`,
```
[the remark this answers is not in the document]
fanyang-mono: Fixed.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,273 @@
+import { app, HttpRequest, HttpResponseInit, InvocationContext, Timer } from "@azure/functions";
+import { TableClient } from "@azure/data-tables";
+import { AzureCliCredential, ManagedIdentityCredential } from "@azure/identity";
+import { listMsbenchDates, enumerateMsbenchBlobs, getMsbenchBlobContent } from "../msbenchBlobEnumerator";
+import { BlobTreeNode } from "../blobEnumerator";
+import { logRequestIdentity } from "../requestIdentity";
+
+const MSBENCH_STORAGE_ACCOUNT = "msbenchnightlydata";
+const EVAL_TABLE_NAME = "msbenchevalmetrics";
+const EVAL_REPORT_PATTERN = /^.*_eval_report\.json$/;
```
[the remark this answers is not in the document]
fanyang-mono: Fixed.

dashboard/src/performance-dashboard/App.tsx
```diff
@@ -0,0 +1,456 @@
...
+            if (!byDate[row.date]) byDate[row.date] = {};
+            const key = seriesKey(row);
+            byDate[row.date][key] = (byDate[row.date][key] || 0) + (Number(row.totalSteps) || 0);
+        }
+        return Object.entries(byDate)
+            .map(([date, series]) => ({ date, ...series }))
+            .sort((a, b) => a.date.localeCompare(b.date));
+    }, [filteredMetrics]);
+
+    const resolvedChartData = useMemo(() => {
```
jongio, line 178: The fix for the earlier stale-closure issue updated `tokenChartData` and `resolvedChartData` to include `selectedBenchmark, selectedModel` in their dependency arrays, but `stepsChartData` only has `[filteredMetrics]`. Since `seriesKey(row)` closes over `selectedBenchmark`/`selectedModel`, the steps chart won't recompute its series when filters change, so its series keys can go stale versus the other two charts.

```suggestion
    }, [filteredMetrics, selectedBenchmark, selectedModel]);
```

jongio (commented):
New commit adds a time-range filter. Two things since my last pass.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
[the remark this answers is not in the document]
jongio: The concern isn't about what's stored in the table, it's about what arrives in `?benchmark=` and `?model=`. Those are URL query params, so anything the caller sends (including a stray apostrophe) flows straight into the filter literal. A benign URL with a quote produces a 500; a hostile one can break out of the quoted value. Escaping is a 2-line fix:

```typescript
const escape = (v: string) => v.replace(/'/g, "''");
if (filterBenchmark) filters.push(`benchmark eq '${escape(filterBenchmark)}'`);
if (filterModel) filters.push(`model eq '${escape(filterModel)}'`);
```

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
fanyang-mono: This sounds like a good idea. Will implement.

dashboard/src/performance-dashboard/App.tsx
fanyang-mono: Will change.

dashboard/api/src/functions/downloadMsbenchBlob.ts
fanyang-mono: The storage account contains `msbench-reports` blob only contains data designed to be shown on the dashboard.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
fanyang-mono: This behavior is okay. Because if some data is missing for some days. We can always do a force run to refill the missing data. We don't want to skip for dates with incorrect data.

dashboard/api/src/msbenchBlobEnumerator.ts
fanyang-mono: Refactored.

dashboard/src/performance-dashboard/App.tsx
fanyang-mono: Verified that the steps chart was able to recompute successfully when filter changes.

jongio (commented):
Latest commit splits the sync into its own Function App with `authLevel: 'function'` - good improvement. The API side now reuses blob helpers cleanly.

Two items still open from earlier threads:

1. OData filter in `getMsbenchEvalMetrics.ts` still interpolates query params unsanitized (my comment on line 39). A `value.replace(/'/g, "''")` one-liner closes it.

2. `hideAll` deps should include `seriesNames` (or its source inputs including `timeRange`/custom dates) so the callback picks up filtered series when time range changes. The eslint-disable masks this.

dashboard/infra/modules/sync-function-app.bicep
```diff
@@ -0,0 +1,123 @@
...
+          }
+        }
+      }
+      scaleAndConcurrency: {
+        maximumInstanceCount: 10
+        instanceMemoryMB: 2048
+      }
+      runtime: {
+        name: 'node'
+        version: '20'
```
JasonYeMSFT: Not urgent for this PR and I know my existing function app also uses node 20. We should consider upgrading them since node 20 is about to reach its end of life (Apr 30 2026).

JasonYeMSFT (dismissed):

dashboard/src/performance-dashboard/App.tsx
```diff
@@ -0,0 +1,416 @@
...
+            return next;
+        });
+    }, []);
+
+    const showAll = useCallback(() => setHiddenSeries(new Set()), []);
+    const hideAll = useCallback(
+        () => setHiddenSeries(new Set(seriesNames)),
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+        [metrics, selectedBenchmark, selectedModel],
+    );
```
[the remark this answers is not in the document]
fanyang-mono: seriesNames is a new array reference on every recompute (from useMemo), so listing it directly as a dependency would cause hideAll to be recreated more often than necessary. By depending on the upstream values (metrics, selectedBenchmark, selectedModel), the callback only refreshes when the actual data changes — and at that point seriesNames will have the correct updated value when hideAll is eventually called.

dashboard/sync/src/functions/syncMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,153 @@
...
+            const dateNode = tree[date];
+            if (!dateNode) return;
+
+            const evalPaths = collectEvalReportPaths(dateNode);
+
+            const reports = await Promise.all(
+                evalPaths.map(async (path) => {
+                    try {
+                        const raw = await getMsbenchBlobContent(path);
+                        return JSON.parse(raw) as EvalReport;
```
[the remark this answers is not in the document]
fanyang-mono: Fixed.

dashboard/api/src/msbenchBlobEnumerator.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const MSBENCH_STORAGE_ACCOUNT = process.env.MSBENCH_STORAGE_ACCOUNT;
+const MSBENCH_REPORTS_CONTAINER_NAME = process.env.MSBENCH_REPORTS_CONTAINER;
+
+function getContainerClient() {
+    const blobServiceClient = new BlobServiceClient(
+        `https://${MSBENCH_STORAGE_ACCOUNT}.blob.core.windows.net`,
+        getCredential()
+    );
+    return blobServiceClient.getContainerClient(MSBENCH_REPORTS_CONTAINER_NAME!);
```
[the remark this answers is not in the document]
fanyang-mono: Fixed.

dashboard/infra/modules/sync-function-app.bicep
fanyang-mono: Upgraded to Node 22

dashboard/infra/main.bicep
```diff
@@ -39,10 +39,13 @@ module identity './modules/managed-identity.bicep' = {
...
-// TO-DO for post deployment: make role assignment for this MI and limit role assignment to read only for dashboard MI
-// This MI has the following permissions to the msbenchnightlydata storage account:
-// - Storage Blob Data Reader role on the storage account, which allows it to read blobs in any container. This is needed to read MSBench metrics stored in eval_report.json.
-// - Storage Table Data Contributor role on the storage account, which allows it to read/write entities in any table. This is needed to store MSBench eval metrics in msbenchevalmetrics table that are retrieved from the eval_report.json files.
+// The dashboard MI (identity) has the following permissions to the msbenchnightlydata storage account:
+// - Storage Blob Data Reader: read blobs in any container (read MSBench metrics from eval_report.json).
+// - Storage Table Data Reader: read entities in any table (read MSBench eval metrics from msbenchevalmetrics table).
+//
+// The sync MI (syncIdentity) has the following permissions to the msbenchnightlydata storage account:
+// - Storage Blob Data Reader: read blobs in any container (read eval_report.json files for syncing).
```
JasonYeMSFT, line 47: Does it no longer need write permission? I remember seeing an earlier version of the code reading a "progress" blob to determine what to upsert to the table and writing a new version of the progress blob back.

dashboard/sync/src/functions/syncMsbenchEvalMetrics.ts
```diff
@@ -66,18 +72,23 @@ async function runSync(context: InvocationContext, force = false): Promise<{ syn
         return { synced: 0, message: "All dates already processed." };
     }
 
+    const CONCURRENCY_LIMIT = 5;
```
JasonYeMSFT, line 77: Could you please add a comment on what this variable tries to control?

dashboard/infra/main.bicep
fanyang-mono: The sync MI reads content from blob container and writes data to the table in the same storage account. That's all. It doesn't writes data to blob container.

dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -0,0 +1,125 @@
...
+        const tableClient = getEvalTableClient();
+        const benchmarks = new Set<string>();
+        const models = new Set<string>();
+
+        for await (const entity of tableClient.listEntities({
+            queryOptions: { select: ["benchmark", "model"] },
+        })) {
+            if (entity.benchmark) benchmarks.add(entity.benchmark as string);
+            if (entity.model) models.add(entity.model as string);
+        }
```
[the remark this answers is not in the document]
fanyang-mono: Will consider later.

dashboard/sync/src/functions/syncMsbenchEvalMetrics.ts
fanyang-mono: Added.

JasonYeMSFT (approved):

PR #1963, opened 2026-04-20
Add repo condition to workflows

JasonYeMSFT:
## Description

This would prevent forked repos that have Action enabled to attempt to run these jobs. Those attempts are pointless since these jobs are only meant for the original repo and depend on its pre-configured secrets.

## Checklist

### DIFF
--- a/.github/workflows/dashboard-collect.yml
+++ b/.github/workflows/dashboard-collect.yml
@@ -17,6 +17,7 @@ on:
 jobs:
   collect-and-upload:
     name: Collect and upload dashboard data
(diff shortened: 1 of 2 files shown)


kvenkatrajan (approved):

jongio (commented):
Same guard could apply to a few other workflows that rely on org secrets or write access and don't already have it: `update-plugin-versions.yml` (SEMANTIC_RELEASE_TOKEN), `skill-factory.yml` (ASSIGN_ISSUE_TOKEN, contents write), `refresh-azure-mcp-tool-snapshot.yml` (scheduled, opens PRs). Not blocking this PR, just flagging for a possible follow-up.

PR #1967, opened 2026-04-20
Update integration tests which include container app deployment

### LINKED ISSUE #1856: Integration test failure: azure-deploy – simple containerized Node.js Bicep Container Apps swedencentral [Timeout]
## Prompt

````
Create a simple containerized Node.js hello world app and deploy to Azure Container Apps using my current subscription in swedencentral region.
```

Follow-up: `Go with recommended options and proceed with Azure deployment.`

## Summary

**Run:** [Integration Tests - all #159](https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/24389185407)
**Test:** `azure-deploy_ - Integration Tests › vanilla-azure-container-apps-deploy › creates simple containerized Node.js app`
**Result:** Fail (Timeout — 30m 0s exceeded)
**Duration:** 30m 0s

## Root Cause Category

**Timeout** — `azd provision` still running after 15+ minutes in swedencentral; session expired before completion

## Diagnosis

### What was expected
A simple containerized Node.js app deployed to Azure Container Apps using Bicep in swedencentral, with a live `*.azurecontainerapps.io` URL.

### What actually happened
The full skill chain executed (azure-prepare → azure-quotas → azure-validate → azure-deploy). The agent invoked `azd provision` and was actively waiting for infrastructure provisioning when the test timeout hit:

```
response: Provisioning Azure resources (azd provision)
...
Creating/Updating resources
  You can view detailed progress in the Azure Portal:
  (portal.azure.com/redacted) (truncated)
````

The agent waited with 300-second `read_bash` delays, and provisioning was still in progress at the 30-minute mark. Resource group was created but no deployed URL was produced.

### Why it failed
- Azure Container Apps provisioning in swedencentral consistently takes 15+ minutes, exceeding the 30-minute test budget when combined with preceding workflow steps
- Resource group naming conflict not detected during pre-flight checks
- No timeout detection or early termination for long-running `azd provision` operations
- This is a recurrent pattern for Container Apps in swedencentral — see also similar timeout for vanilla-azure-container-apps containerized web app in this same run

## azure-deploy Skill Invocation

| Skill | Invoked |
|-------|---------|
| **azure-prepare** | ✅ Yes |
| **azure-validate** | ✅ Yes |
| **azure-deploy** | ✅ Yes |

> In azure-deploy integration tests, the full skill chain is azure-prepare → azure-validate → azure-deploy.

## Skill Report Context

From the `vanilla-azure-container-apps` SKILL-REPORT (run 2026-04-14):

- **Overall Test Pass Rate (per report):** 0% (0/1)
- **Average Confidence:** 65%
- Resource group naming conflict not caught by pre-deployment validation
- Environment naming should default to unique values to avoid conflicts

## Environment

- **Model:** claude-sonnet-4.6
- **Region:** swedencentral
- **IaC:** Bicep
- **Run URL:** https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/24389185407
- **Commit:** c17515d3156f8a7b76facf0bf7a5d7991eaebfa3

> Generated by [Analyze Test Run](https://github.com/microsoft/GitHub-Copilot-for-Azure/actions/runs/24399977654/agentic_workflow) · ● 3.1M · [◷](https://github.com/search?q=repo%3Amicrosoft%2FGitHub-Copilot-for-Azure+is%3Aissue+%22gh-aw-workflow-call-id%3A+microsoft%2FGitHub-Copilot-for-Azure%2Fanalyze-test-run%22&type=issues)

### PR DESCRIPTION
**Pull request by fanyang-mono**
## Description

Container app deployment often fails with timeout due to the delay of AcrPull role assignment propagation. This PR update all of the 6 container app deployment tests to

1. Terminate tests after `azd provision` is completed
2. Verify two-phase IAC code pattern

Here is a detailed explain about two-phase container app deployment
###################################################
The "two-phase" terminology refers to **Bicep module dependency ordering**, not separate CLI invocations. Here's exactly what happens:

## How Many Commands?

- **`azd provision`**: **1 time** — deploys all Bicep (both "phases") in a single ARM deployment
- **`azd deploy`**: **1 time** — pushes your app image and configures the registry link
- Or just **`azd up`**: **1 time** — runs both provision + deploy sequentially

**No Bicep modifications between commands.**

### 1. Write your Bicep (once)

Your `main.bicep` contains all three modules from the file:

| Module | "Phase" | What it does |
|--------|---------|-------------|
| `containerRegistry` | 1 | Creates ACR |
| `api` | 1 | Creates Container App with **placeholder image** (`mcr.microsoft.com/azuredocs/containerapps-helloworld:latest`), system-assigned managed identity, **no `registries` block** |
| `acrPullRole` | 2 | Assigns AcrPull role to the Container App's managed identity, using **outputs** from both Phase 1 modules |

The "phases" are just the dependency graph — ARM automatically deploys Phase 1 modules first, then Phase 2, because `acrPullRole` references outputs from the other two.

### 2. Run `azd up`

This executes two stages internally:

**Stage A — `azd provision`:**
- ARM deploys all three modules in dependency order
- Result: ACR exists, Container App exists (running placeholder image), role assignment exists

**Stage B — `azd deploy`:**
- `azd` automatically runs `az containerapp registry set --server <acr> --identity system` to link the registry to the app
- Builds and pushes your real app image to ACR
- Updates the Container App to use the real image

### 3. Done

No second provision, no Bicep edits.

## Why This Pattern Exists

Without the separate `acrPullRole` module, you'd get a **circular dependency**:
- ACR module needs to know the Container App's principal ID (for the role assignment)
- Container App module needs to reference ACR (if you put the `registries` block in Bicep)

By omitting the `registries` block from Bicep entirely and letting `azd deploy` handle it via the API, the circular dependency is broken. The role assignment module simply consumes outputs from both independent modules.
###################################################

## Related Issues

Fixes #1856

### DIFF
--- a/tests/azure-deploy/integration.test.ts
+++ b/tests/azure-deploy/integration.test.ts
@@ -670,7 +670,7 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
         // Phase 1: ACR resource
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /azurerm_container_registry/i, tfPattern)).toBe(true);
         // Phase 1: Container App with placeholder image, system-assigned identity, and lifecycle ignore_changes
-        expect(doesWorkspaceFileIncludePattern(workspacePath!, /mcr\.microsoft\.com/i, tfPattern)).toBe(true);
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /(^|[^A-Za-z0-9.-])mcr\.microsoft\.com\//i, tfPattern)).toBe(true);
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /SystemAssigned/i, tfPattern)).toBe(true);
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /ignore_changes/i, tfPattern)).toBe(true);
         // Phase 2: AcrPull role assignment in a separate file from the Container App
--- a/tests/azure-deploy/integration.test.ts
+++ b/tests/azure-deploy/integration.test.ts
@@ -744,7 +744,7 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
         // Phase 1: ACR resource
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /azurerm_container_registry/i, tfPattern)).toBe(true);
         // Phase 1: Container App with placeholder image, system-assigned identity, and lifecycle ignore_changes
-        expect(doesWorkspaceFileIncludePattern(workspacePath!, /mcr\.microsoft\.com/i, tfPattern)).toBe(true);
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /(^|[^A-Za-z0-9.-])mcr\.microsoft\.com($|[^A-Za-z0-9.-])/i, tfPattern)).toBe(true);
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /SystemAssigned/i, tfPattern)).toBe(true);
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /ignore_changes/i, tfPattern)).toBe(true);
         // Phase 2: AcrPull role assignment in a separate file from the Container App
--- a/tests/azure-deploy/integration.test.ts
+++ b/tests/azure-deploy/integration.test.ts
@@ -446,7 +446,7 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
         // Phase 1: ACR module
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /Microsoft\.ContainerRegistry\/registries/i, bicepPattern)).toBe(true);
         // Phase 1: Container App with placeholder image and system-assigned managed identity
-        expect(doesWorkspaceFileIncludePattern(workspacePath!, /mcr\.microsoft\.com/i, bicepPattern)).toBe(true);
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /(^|[^A-Za-z0-9.-])mcr\.microsoft\.com([^A-Za-z0-9.-]|$)/i, bicepPattern)).toBe(true);
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /SystemAssigned/i, bicepPattern)).toBe(true);
         // Phase 2: AcrPull role assignment in a separate module file from the Container App
         expect(arePatternsInSeparateFiles(workspacePath!, /Microsoft\.App\/containerApps/i, /7f951dda-4ed3-4680-a7ca-43fe172d538d/i, bicepPattern)).toBe(true);
--- a/tests/azure-deploy/integration.test.ts
+++ b/tests/azure-deploy/integration.test.ts
@@ -410,7 +410,7 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
         // Phase 1: ACR module
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /Microsoft\.ContainerRegistry\/registries/i, bicepPattern)).toBe(true);
         // Phase 1: Container App with placeholder image and system-assigned managed identity
-        expect(doesWorkspaceFileIncludePattern(workspacePath!, /mcr\.microsoft\.com/i, bicepPattern)).toBe(true);
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /(^|[^A-Za-z0-9.-])mcr\.microsoft\.com($|[^A-Za-z0-9.-])/i, bicepPattern)).toBe(true);
         expect(doesWorkspaceFileIncludePattern(workspacePath!, /SystemAssigned/i, bicepPattern)).toBe(true);
         // Phase 2: AcrPull role assignment in a separate module file from the Container App
         expect(arePatternsInSeparateFiles(workspacePath!, /Microsoft\.App\/containerApps/i, /7f951dda-4ed3-4680-a7ca-43fe172d538d/i, bicepPattern)).toBe(true);
--- a/tests/azure-deploy/integration.test.ts
+++ b/tests/azure-deploy/integration.test.ts
@@ -14,9 +14,9 @@ import {
   getIntegrationSkipReason,
   useAgentRunner,
 } from "../utils/agent-runner";
-import { hasDeployLinks, softCheckDeploySkills, softCheckContainerDeployEnvVars, shouldEarlyTerminateForCompletedDeployment } from "./utils";
+import { hasDeployLinks, softCheckDeploySkills, softCheckContainerDeployEnvVars, shouldEarlyTerminateForCompletedDeployment, shouldEarlyTerminateForAzdProvision } from "./utils";
 import { cloneRepo } from "../utils/git-clone";
-import { expectFiles, softCheckSkill, doesWorkspaceFileIncludePattern, shouldEarlyTerminateForSkillInvocation, isSkillInvoked, withTestResult } from "../utils/evaluate";
+import { expectFiles, softCheckSkill, doesWorkspaceFileIncludePattern, arePatternsInSeparateFiles, shouldEarlyTerminateForSkillInvocation, isSkillInvoked, withTestResult } from "../utils/evaluate";
 
 const SKILL_NAME = "azure-deploy";
 const RUNS_PER_PROMPT = 1;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tests/utils/evaluate.ts [hunk 1]
```diff
@@ -61,6 +61,48 @@ export function doesWorkspaceFileIncludePattern(workspace: string, valuePattern:
...
+      }
+    }
+  };
+
+  scanDirectory(workspace);
+
+  if (filesWithA.length === 0 || filesWithB.length === 0) return false;
+
+  // Check that at least one file with patternB is different from all files with patternA
+  return filesWithB.some(f => !filesWithA.includes(f));
```
jongio: The name promises "separate files" but the return check doesn't enforce that. It returns true whenever any file has patternB without patternA, even if some file also has both. So if a main.bicep happens to contain both the Container App and the AcrPull role GUID, and any second file also mentions AcrPull, this returns true, which isn't what the two-phase assertion is trying to prove.

The intent is "the role assignment does not live alongside the container app." That's the stronger check:

```suggestion
  if (filesWithA.length === 0 || filesWithB.length === 0) return false;

  // Every file containing patternA must be free of patternB
  const setB = new Set(filesWithB);
  return filesWithA.every(f => !setB.has(f)) && filesWithB.some(f => !filesWithA.includes(f));
```

Or if the looser check is actually what you want, rename to match the behavior (something like `hasFileWithBWithoutA`).

tests/azure-deploy/utils.ts
```diff
@@ -63,3 +63,14 @@ export function shouldEarlyTerminateForCompletedDeployment(agentMetadata: AgentM
...
+
+export function shouldEarlyTerminateForAzdProvision(agentMetadata: AgentMetadata): boolean {
+  const hasCalledProvision = matchesCommand(agentMetadata, /azd\s+provision/i);
+  if (hasCalledProvision) {
+    const commentToAdd = "✅ azd provision was called. Terminating early — infrastructure provisioning has started.";
+    if (!agentMetadata.testComments.some((testComment) => testComment === commentToAdd)) {
+      agentMetadata.testComments.push(commentToAdd);
+    }
+  }
+  return hasCalledProvision;
```
jongio: Backing the bot's concern with a bit more evidence: `matchesCommand` reads from `tool.execution_start` arguments (via `getShellCommands` in evaluate.ts:24), and `shouldEarlyTerminate` fires on every new event in agent-runner.ts:573. So this terminates the instant the agent issues the `azd provision` shell call, before the command runs, let alone succeeds. A typo, a wrong subscription, or an immediate validation failure all look identical to a successful provision start.

If you want a cheap sanity check without waiting on the full deploy, gate on the matching `tool.execution_complete` event succeeding (see `areToolCallsSuccess` in agent-runner.ts:666). Combined with the bot's `azd up` suggestion:

```suggestion
  const hasCalledProvision = matchesCommand(agentMetadata, /azd\s+provision/i);
  const hasCalledUp = matchesCommand(agentMetadata, /azd\s+up/i);
  if (!hasCalledProvision && !hasCalledUp) return false;

  const commentToAdd = hasCalledProvision
    ? "✅ azd provision was called. Terminating early, infrastructure provisioning has started."
    : "❌ azd up was called. Container Apps tests should use azd provision so failures surface fast.";
  if (!agentMetadata.testComments.some(c => c === commentToAdd)) {
    agentMetadata.testComments.push(commentToAdd);
  }
  return true;
```

jongio (commented):
Two things worth considering beyond what the bots already flagged.

CodeQL is failing on 6 unanchored regex alerts (the `mcr.microsoft.com`, `azurerm_container_app[^_]`, and `Microsoft.App/containerApps` patterns). These aren't really regex matches, they're literal substring checks against generated IaC file content. The simplest structural fix that resolves all 6 alerts at once: add a sibling helper that takes a plain string:

```ts
export function doesWorkspaceFileIncludeText(workspace: string, needle: string, filePattern?: RegExp): boolean
```

Then the 6 call sites become `doesWorkspaceFileIncludeText(workspacePath!, "mcr.microsoft.com", bicepPattern)` etc. CodeQL stops complaining, and the intent reads clearer since these were never meant as patterns.

The other thing is `arePatternsInSeparateFiles`, see inline.

tests/azure-deploy/integration.test.ts
```diff
@@ -394,15 +394,26 @@
...
+        // Verify Container Apps-specific Bicep content on disk
+        const bicepPattern = /\.bicep$/;
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /Microsoft\.App\/containerApps/i, bicepPattern)).toBe(true);
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /Microsoft\.App\/managedEnvironments/i, bicepPattern)).toBe(true);
+
+        // Verify two-phase deployment pattern (three modules in main.bicep):
+        // Phase 1: ACR module
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /Microsoft\.ContainerRegistry\/registries/i, bicepPattern)).toBe(true);
+        // Phase 1: Container App with placeholder image and system-assigned managed identity
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /mcr\.microsoft\.com/i, bicepPattern)).toBe(true);
```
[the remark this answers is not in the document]
fanyang-mono: Fixed.

tests/azure-deploy/integration.test.ts
```diff
@@ -632,15 +654,27 @@
...
+        // Verify Container Apps-specific Terraform content on disk
+        const tfPattern = /\.tf$/;
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /azurerm_container_app[^_]/i, tfPattern)).toBe(true);
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /azurerm_container_app_environment/i, tfPattern)).toBe(true);
+
+        // Verify two-phase deployment pattern (three resources):
+        // Phase 1: ACR resource
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /azurerm_container_registry/i, tfPattern)).toBe(true);
+        // Phase 1: Container App with placeholder image, system-assigned identity, and lifecycle ignore_changes
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /mcr\.microsoft\.com/i, tfPattern)).toBe(true);
```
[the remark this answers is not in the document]
fanyang-mono: Fixed.

tests/utils/evaluate.ts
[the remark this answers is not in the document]
fanyang-mono: Updated.

tests/azure-deploy/utils.ts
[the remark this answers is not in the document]
fanyang-mono: Fixed.

tests/azure-deploy/integration.test.ts
```diff
@@ -394,15 +394,26 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
...
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /Microsoft\.App\/managedEnvironments/i, bicepPattern)).toBe(true);
+
+        // Verify two-phase deployment pattern (three modules in main.bicep):
+        // Phase 1: ACR module
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /Microsoft\.ContainerRegistry\/registries/i, bicepPattern)).toBe(true);
+        // Phase 1: Container App with placeholder image and system-assigned managed identity
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /mcr\.microsoft\.com/i, bicepPattern)).toBe(true);
+        expect(doesWorkspaceFileIncludePattern(workspacePath!, /SystemAssigned/i, bicepPattern)).toBe(true);
+        // Phase 2: AcrPull role assignment in a separate module file from the Container App
+        expect(arePatternsInSeparateFiles(workspacePath!, /Microsoft\.App\/containerApps/i, /7f951dda-4ed3-4680-a7ca-43fe172d538d/i, bicepPattern)).toBe(true);
```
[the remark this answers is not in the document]
fanyang-mono: This is fine for now

tests/utils/evaluate.ts
fanyang-mono: Fixed.

tests/azure-deploy/utils.ts
fanyang-mono: Fixed.

tests/utils/evaluate.ts
```diff
@@ -61,6 +61,47 @@ export function doesWorkspaceFileIncludePattern(workspace: string, valuePattern:
...
+        if (filePattern && !entry.name.match(filePattern)) {
+          continue;
+        }
+        try {
+          const content = fs.readFileSync(fullPath, "utf-8");
+          const matchesA = !!content.match(patternA);
+          const matchesB = !!content.match(patternB);
+          if (matchesA) hasA = true;
+          if (matchesB && !matchesA) hasBNotA = true;
+          if (hasA && hasBNotA) return true;
```
JasonYeMSFT: Did you intend to write `hasA || hasBNotA`? `hasA && hasBNotA` should always be false.

tests/utils/evaluate.ts
```diff
@@ -61,6 +61,47 @@ export function doesWorkspaceFileIncludePattern(workspace: string, valuePattern:
...
+/**
+ * Checks that two value patterns exist in **different** files within the workspace.
+ * This verifies separation of concerns — e.g. the AcrPull role assignment is in a separate module from the Container App.
+ * @param workspace Path to a directory containing the files of interest.
+ * @param patternA First value pattern to match
+ * @param patternB Second value pattern — must be in a different file from patternA
+ * @param filePattern If provided, only files whose names match the pattern are considered
+ * @returns True if both patterns are found and they appear in different files
+ */
+export function arePatternsInSeparateFiles(workspace: string, patternA: RegExp, patternB: RegExp, filePattern?: RegExp): boolean {
```
JasonYeMSFT: The return value can be expanded to include information about whether the result is false due to either pattern not found or both patterns found in the same file. It will be helpful for investigating failures.

tests/utils/evaluate.ts
```diff
@@ -61,6 +61,47 @@ export function doesWorkspaceFileIncludePattern(workspace: string, valuePattern:
   return scanDirectory(workspace);
 }
 
+/**
+ * Checks that two value patterns exist in **different** files within the workspace.
+ * This verifies separation of concerns — e.g. the AcrPull role assignment is in a separate module from the Container App.
```
JasonYeMSFT:
```suggestion
 * This verifies patterns that must exist in different files — e.g. the AcrPull role assignment and the Container App must be in separate bicep modules so they can be provisioned separately to avoid cyclic dependency.
```

tests/azure-deploy/utils.ts
```diff
@@ -63,3 +63,14 @@ export function shouldEarlyTerminateForCompletedDeployment(agentMetadata: AgentM
   }
   return containsDeployLinks;
 }
+
+export function shouldEarlyTerminateForAzdProvision(agentMetadata: AgentMetadata): boolean {
+  const hasCalledProvision = matchesCommand(agentMetadata, /azd\s+(provision|up)\b/i);
```
JasonYeMSFT: We don't even need to start running the provision command. We can stop the agent once we detect it's about to execute the azd provision/up command using PreToolUseHook. Would you like to try it?

Otherwise, we can try to capture the completion of the first provision to catch more potential issues in a real deployment scenario.

jongio (commented):
Addresses my previous feedback. Anchored regexes clear CodeQL, `arePatternsInSeparateFiles` is rewritten correctly with early-exit, and `shouldEarlyTerminateForAzdProvision` catches both `provision` and `up`.

jongio (commented):
Addresses my previous feedback. The incremental doc improvement on rePatternsInSeparateFiles is a nice touch - explaining the cyclic dependency motivation reads much clearer than the old separation of concerns wording.

Two open threads from @JasonYeMSFT worth considering as follow-ups:
- Expanding the return value to include diagnostic info (pattern not found vs both in same file)
- Using PreToolUseHook to intercept before provision runs rather than detecting after execution

CI is clean. Code logic hasn't changed since my last review.

tests/utils/evaluate.ts
fanyang-mono: Updated.

tests/azure-deploy/utils.ts
fanyang-mono: My original thought was to capture the first azd provision and stop there. That way we can still see the issues exposed by azd provision and try to fix the skills. Otherwise, I am not confident that the bicep files our skills help generated is good enough.

JasonYeMSFT: If we stop at the execution start of azd provision, we wouldn't see the issues exposed by azd provision because the agent runner won't wait for the result of azd provision. You might need to create a different helper function to detect the execution_complete event of the azd provision command instead.

fanyang-mono: Updated termination condition.

JasonYeMSFT (approved):

PR #1976, opened 2026-04-21
feat(azure-kubernetes): Update constraint spec references to match what Deployment Safeguards enforces

NickKeller:
## Description

Update constraint spec references to match what is actually enforced by Deployment Safeguards. For Pod Security Standards, it uses actual language from the Kubernetes Pod Security Standards spec.

## Checklist

- [x] **If modifying skill descriptions:** verified routing correctness with integration tests (`npm run test:skills:integration -- <skill>`)

### DIFF
--- a/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
+++ b/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
@@ -130,8 +130,9 @@ Key Checks:
 - Readiness and liveness probes → `safeguard-probes-configured` *(warning-only — not blocked at admission; treat as informational)*
 - Image tag not `:latest` → `safeguard-images-no-latest`
 - `securityContext.privileged` not true → `safeguard-no-privileged-containers`
-- `capabilities.add` empty → `safeguard-container-capabilities`
+- `capabilities.add` only adds allowed capabilities → `safeguard-container-capabilities`
 - `seccompProfile` is RuntimeDefault/Localhost → `safeguard-allowed-seccomp-profiles`
+- no `host` field in any container probes and lifecycle hooks
 
 **Per pod spec:**
 - `hostPID`/`hostIPC` not true → `safeguard-block-host-namespaces` (incompatible)
--- a/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
+++ b/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
@@ -17,7 +17,7 @@ description: "Assess Kubernetes workloads and cluster configuration for AKS Auto
 
 You are an AKS Automatic compatibility assessment agent. Your job is to evaluate whether Kubernetes workloads and cluster configurations are compatible with [AKS Automatic](https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic), identify issues, and help users fix them.
 
-AKS Automatic enforces **Deployment Safeguards** (21 active Deny policies), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
+AKS Automatic enforces **Deployment Safeguards** (21 active policies, some deny, some warn only, some auto-fixed), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
 
 ## Quick Reference
 | Property | Value |
@@ -167,7 +167,6 @@ Per-issue format:
 
 **Deterministic fixes** (have `suggestedPatch` — generate YAML diff directly):
 - `safeguard-container-resource-requests` — add `resources.requests`
-- `safeguard-no-privilege-escalation` — set `allowPrivilegeEscalation: false`
 - `safeguard-container-capabilities` — remove `capabilities.add`
 - `safeguard-allowed-seccomp-profiles` — add `seccompProfile: RuntimeDefault`
 - `safeguard-enforce-apparmor` — add AppArmor annotation
--- a/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
+++ b/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
@@ -17,7 +17,7 @@ description: "Assess Kubernetes workloads and cluster configuration for AKS Auto
 
 You are an AKS Automatic compatibility assessment agent. Your job is to evaluate whether Kubernetes workloads and cluster configurations are compatible with [AKS Automatic](https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic), identify issues, and help users fix them.
 
-AKS Automatic enforces **Deployment Safeguards** (23 active policies, some deny, some warn only, some auto-fixed), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **23 cluster-level configuration requirements**.
+AKS Automatic enforces **Deployment Safeguards** (21 active policies, some deny, some warn only), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **23 cluster-level configuration requirements**.
 
 ## Quick Reference
 | Property | Value |
@@ -246,4 +246,4 @@ See `references/migration-guide-summary.md` for the full migration checklist.
 | `references/migration-guide-summary.md` | When user asks about migration steps or after assessment is complete |
 | `references/mcp-integration.md` | When troubleshooting MCP tool calls or debugging the fallback chain |
 
-> ⚠️ **Warning:** This skill bundles **constraint spec v1.1.1** (2026-03-15), covering 23 cluster-level constraints, 23 active Deployment Safeguards policies(9 best practices policies, 12 Pod Security Standards policies, and 2 active mutators) Always note the spec version in assessment output.
+> ⚠️ **Warning:** This skill bundles **constraint spec v1.1.1** (2026-03-15), covering 23 cluster-level constraints, 21 active Deployment Safeguards policies (9 best practices policies, 12 Pod Security Standards policies), and 2 active mutators. Always note the spec version in assessment output.
--- a/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
+++ b/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
@@ -132,7 +132,7 @@ Key Checks:
 - `securityContext.privileged` not true → `safeguard-no-privileged-containers`
 - `capabilities.add` only adds allowed capabilities → `safeguard-container-capabilities`
 - `seccompProfile` is RuntimeDefault/Localhost → `safeguard-allowed-seccomp-profiles`
-- no `host` field in any container probes and lifecycle hooks
+- no `host` field in any container probes and lifecycle hooks → `safeguard-host-probes`
 
 **Per pod spec:**
 - `hostPID`/`hostIPC` not true → `safeguard-block-host-namespaces` (incompatible)
--- a/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
+++ b/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
@@ -187,7 +187,7 @@ Per-issue format:
 **Deterministic fixes** (have `suggestedPatch` — generate YAML diff directly):
 - `safeguard-container-resource-requests` — add `resources.requests`
 - `safeguard-container-capabilities` — remove `capabilities.add`
-- `safeguard-allowed-seccomp-profiles` — add `seccompProfile: RuntimeDefault`
+- `safeguard-allowed-seccomp-profiles` — patch only when `seccompProfile.type: Unconfined` is present, or when the MCP `suggestedPatch` explicitly requires a seccomp change
 - `safeguard-enforce-apparmor` — add AppArmor annotation
 - `safeguard-csi-driver-storage-class` — replace in-tree provisioner
 
--- a/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
+++ b/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
@@ -124,6 +124,24 @@ Then proceed to offline mode.
 
 Load the constraint spec from `references/constraint-spec-v1.yaml` and evaluate each manifest. The check field tells you what to check for and what fields to check. The fix field will tell you any allowed values and possible fixes. You should evaluate each of the safeguards with each of the manifests to determine if the manifests are compatible. Suggest any fixes that are needed.
 
+Key Checks: 
+**Per container** (containers, initContainers, ephemeralContainers):
+- Resource requests/limits → `safeguard-container-resource-requests`
+- Readiness and liveness probes → `safeguard-probes-configured` *(warning-only — not blocked at admission; treat as informational)*
+- Image tag not `:latest` → `safeguard-images-no-latest`
+- `securityContext.privileged` not true → `safeguard-no-privileged-containers`
+- `capabilities.add` empty → `safeguard-container-capabilities`
+- `seccompProfile` is RuntimeDefault/Localhost → `safeguard-allowed-seccomp-profiles`
+
+**Per pod spec:**
+- `hostPID`/`hostIPC` not true → `safeguard-block-host-namespaces` (incompatible)
+- `hostNetwork`/`hostPort` not true → `safeguard-host-network-ports` (incompatible)
+- No `hostPath` volumes → `safeguard-no-host-path-volumes` (incompatible)
+
+**Per workload type:**
+- Deployments/StatefulSets with replicas > 1: podAntiAffinity or topologySpreadConstraints → `safeguard-pod-enforce-antiaffinity`
+- StorageClass: CSI provisioner (not in-tree) → `safeguard-csi-driver-storage-class`
+
 
 ### Severity Classification
 
--- a/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
+++ b/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
@@ -17,7 +17,7 @@ description: "Assess Kubernetes workloads and cluster configuration for AKS Auto
 
 You are an AKS Automatic compatibility assessment agent. Your job is to evaluate whether Kubernetes workloads and cluster configurations are compatible with [AKS Automatic](https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic), identify issues, and help users fix them.
 
-AKS Automatic enforces **Deployment Safeguards** (21 active policies, some deny, some warn only, some auto-fixed), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
+AKS Automatic enforces **Deployment Safeguards** (23 active policies, some deny, some warn only, some auto-fixed), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
 
 ## Quick Reference
 | Property | Value |
--- a/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
+++ b/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
@@ -3,7 +3,7 @@ name: azure-kubernetes-automatic-readiness
 license: MIT
 metadata:
   author: Microsoft
-  version: "1.0.0"
+  version: "1.0.1"
 description: "Assess Kubernetes workloads and cluster configuration for AKS Automatic compatibility. Identifies incompatibilities, generates fixes, and guides migration from AKS Standard to AKS Automatic. WHEN: migrate to AKS Automatic, check AKS Automatic readiness, validate manifests for Automatic, assess cluster for Automatic compatibility, fix deployment for Automatic compatibility, identify AKS Automatic migration blockers, is my cluster ready for AKS Automatic."
 ---
 
--- a/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
+++ b/plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
@@ -17,7 +17,7 @@ description: "Assess Kubernetes workloads and cluster configuration for AKS Auto
 
 You are an AKS Automatic compatibility assessment agent. Your job is to evaluate whether Kubernetes workloads and cluster configurations are compatible with [AKS Automatic](https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic), identify issues, and help users fix them.
 
-AKS Automatic enforces **Deployment Safeguards** (25 active Deny policies), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
+AKS Automatic enforces **Deployment Safeguards** (21 active Deny policies), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
 
 ## Quick Reference
 | Property | Value |
@@ -122,26 +122,8 @@ Then proceed to offline mode.
 
 #### Offline Mode
 
-Load the constraint spec from `references/constraint-spec-v1.yaml` and evaluate each manifest. Key checks:
+Load the constraint spec from `references/constraint-spec-v1.yaml` and evaluate each manifest. The check field tells you what to check for and what fields to check. The fix field will tell you any allowed values and possible fixes. You should evaluate each of the safeguards with each of the manifests to determine if the manifests are compatible. Suggest any fixes that are needed.
 
-**Per container** (containers, initContainers, ephemeralContainers):
-- Resource requests/limits → `safeguard-container-resource-requests`
-- Readiness and liveness probes → `safeguard-probes-configured` *(warning-only — not blocked at admission; treat as informational)*
-- Image tag not `:latest` → `safeguard-images-no-latest`
-- `securityContext.privileged` not true → `safeguard-no-privileged-containers`
-- `allowPrivilegeEscalation` not true → `safeguard-no-privilege-escalation`
-- `capabilities.add` empty → `safeguard-container-capabilities`
-- `seccompProfile` is RuntimeDefault/Localhost → `safeguard-allowed-seccomp-profiles`
-
-**Per pod spec:**
-- `hostPID`/`hostIPC` not true → `safeguard-block-host-namespaces` (incompatible)
-- `hostNetwork`/`hostPort` not true → `safeguard-host-network-ports` (incompatible)
-- No `hostPath` volumes → `safeguard-no-host-path-volumes` (incompatible)
-- Volume types are standard → `safeguard-allowed-volume-types`
-
-**Per workload type:**
-- Deployments/StatefulSets with replicas > 1: podAntiAffinity or topologySpreadConstraints → `safeguard-pod-enforce-antiaffinity`
-- StorageClass: CSI provisioner (not in-tree) → `safeguard-csi-driver-storage-class`
 
 ### Severity Classification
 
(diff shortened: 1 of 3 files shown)


jongio (commented):
Most of the earlier review comments were addressed in the follow-up commit (version bump, section-header counts, probe transports, policyId). A few things I still see:

**Stale cross-reference in SKILL.md**
Line 170 still lists `safeguard-no-privilege-escalation` under `Deterministic fixes`, but this PR removes that entry from `constraint-spec-v1.yaml`. The assessment won't produce a finding for it anymore, so that bullet becomes dead guidance. Either drop the bullet or restore the safeguard in the spec if Automatic still enforces it.

**Missing common-fix recipe for safeguard-host-probes**
The new `safeguard-host-probes` entry has no corresponding section in `references/common-fixes.md`. Other PSS-related safeguards have before/after patterns there; small consistency gap worth closing.

Inline comments for the rest.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
```diff
@@ -17,7 +17,7 @@ description: "Assess Kubernetes workloads and cluster configuration for AKS Auto
 
 You are an AKS Automatic compatibility assessment agent. Your job is to evaluate whether Kubernetes workloads and cluster configurations are compatible with [AKS Automatic](https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic), identify issues, and help users fix them.
```
jongio, line 19: `(21 active Deny policies)` overstates the enforcement. `safeguard-probes-configured` is `enforcement: warn` (spec line 223), and the two `severity: autoFixed` entries (`safeguard-container-resource-requests`, `safeguard-pod-enforce-antiaffinity`) are mutated by webhooks at admission rather than denied. Something like:

```suggestion
AKS Automatic enforces **Deployment Safeguards** (21 active policies: 20 Deny + 1 warning-only), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
```

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
```diff
@@ -122,26 +122,8 @@ Then proceed to offline mode.
...
-**Per pod spec:**
-- `hostPID`/`hostIPC` not true → `safeguard-block-host-namespaces` (incompatible)
-- `hostNetwork`/`hostPort` not true → `safeguard-host-network-ports` (incompatible)
-- No `hostPath` volumes → `safeguard-no-host-path-volumes` (incompatible)
-- Volume types are standard → `safeguard-allowed-volume-types`
-
-**Per workload type:**
-- Deployments/StatefulSets with replicas > 1: podAntiAffinity or topologySpreadConstraints → `safeguard-pod-enforce-antiaffinity`
-- StorageClass: CSI provisioner (not in-tree) → `safeguard-csi-driver-storage-class`
+Load the constraint spec from `references/constraint-spec-v1.yaml` and evaluate each manifest. The check field tells you what to check for and what fields to check. The fix field will tell you any allowed values and possible fixes. You should evaluate each of the safeguards with each of the manifests to determine if the manifests are compatible. Suggest any fixes that are needed.
```
jongio, line 125: This paragraph replaced a structured per-container / per-pod-spec / per-workload / per-storage-class checklist. The new spec entries now carry field paths themselves, which is a real improvement, but losing the top-level categorization means an agent has to derive structure from the full 500-line YAML on every run. Was the removal intentional? If so, consider keeping a short summary of the check categories (per-container fields, per-pod-spec fields, per-workload-type rules, storage-class rules) so the agent has a mental model before it opens the spec. Without that scaffolding, there's a real risk of skipped categories on larger manifests.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -225,7 +214,7 @@ safeguards:
     policyId: 48940d92
     severity: requiresChanges
     category: nodeProtection
-    check: AKS-reserved taint keys blocked for users
+    check: AKS-reserved taint CriticalAddonsOnly key blocked for users
```
jongio, line 217: This narrowed from `AKS-reserved taint keys blocked for users` to `AKS-reserved taint CriticalAddonsOnly key blocked for users`. Is `CriticalAddonsOnly` genuinely the only key this policy blocks? If it also covers other AKS-reserved keys (e.g. `kubernetes.azure.com/scalesetpriority` or other `kubernetes.azure.com/*` taints), the narrower wording will cause agents to miss them during assessment. Worth confirming against the underlying policy definition.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -257,110 +246,245 @@ safeguards:
...
+        - spec.initContainers[*].readinessProbe.tcpSocket.host
+        - spec.initContainers[*].startupProbe.tcpSocket.host
+        - spec.initContainers[*].lifecycle.postStart.tcpSocket.host
+        - spec.initContainers[*].lifecycle.preStop.tcpSocket.host
+        - spec.initContainers[*].lifecycle.postStart.httpGet.host
+        - spec.initContainers[*].lifecycle.preStop.httpGet.host
+    fix: |
+      The allowed values are:
+        - undefined/nil
+        - ""
```
jongio, line 469: The `fix` block only enumerates allowed values. Every other entry tells the caller what to do (`Remove hostPID and hostIPC; incompatible if required.`, `Set RuntimeDefault, or use an allowed Localhost profile.`). Add an imperative, e.g.:

```suggestion
    fix: |
      The allowed values are:
        - undefined/nil
        - ""
      Remove the `host` field from probes and lifecycle hooks; the kubelet uses the pod IP by default.
```

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
```diff
@@ -17,7 +17,7 @@ description: "Assess Kubernetes workloads and cluster configuration for AKS Auto
 
 You are an AKS Automatic compatibility assessment agent. Your job is to evaluate whether Kubernetes workloads and cluster configurations are compatible with [AKS Automatic](https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic), identify issues, and help users fix them.
 
-AKS Automatic enforces **Deployment Safeguards** (25 active Deny policies), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
+AKS Automatic enforces **Deployment Safeguards** (21 active Deny policies), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
```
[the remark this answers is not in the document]
NickKeller: done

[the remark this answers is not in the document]
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -257,110 +246,245 @@ safeguards:
     check: Image tag must not be :latest or untagged (no colon)
     patch: "replace image tag with specific version or sha256 digest"
 
-  # -- Baseline PSS policies in Safeguards (15 policies) --
+  # -- PSS-related policies in Safeguards (13 policies) --
```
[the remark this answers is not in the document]
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -257,110 +246,245 @@ safeguards:
...
+      - NET_BIND_SERVICE
+      - SETFCAP
+      - SETGID
+      - SETPCAP
+      - SETUID
+      - SYS_CHROOT
+      Remove any added capability outside this list.
+
+  - id: safeguard-host-probes
+    policyId: asdf8909
```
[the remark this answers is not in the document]
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
NickKeller: Sure - I'd like a future version to just defer this to calling a deterministic API. But we don't have that yet so I can put in a summary

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
NickKeller: We only reserve the CriticalAddonsOnly taint

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
NickKeller: I just added back the key checks I removed, minus irrelevant ones

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -176,28 +176,17 @@ clusterConstraints:
...
-  - id: hobo-critical-addons-noexecute
-    severity: internal
-    note: AKS-managed, not customer-facing
-  - id: hobo-hosted-vm-taint
-    severity: internal
-    note: AKS-managed, not customer-facing
-
 # =============================================================================
-# WORKLOAD CONSTRAINTS — Deployment Safeguards (25 active policies)
+# WORKLOAD CONSTRAINTS — Deployment Safeguards (21 active policies)
```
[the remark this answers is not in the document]
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -176,28 +176,17 @@ clusterConstraints:
...
 # Initiative: c047ea8e | Effect: Deny on Automatic
```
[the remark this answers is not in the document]
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
```diff
@@ -122,27 +122,27 @@ Then proceed to offline mode.
...
+Load the constraint spec from `references/constraint-spec-v1.yaml` and evaluate each manifest. The check field tells you what to check for and what fields to check. The fix field will tell you any allowed values and possible fixes. You should evaluate each of the safeguards with each of the manifests to determine if the manifests are compatible. Suggest any fixes that are needed.
 
+Key Checks: 
 **Per container** (containers, initContainers, ephemeralContainers):
 - Resource requests/limits → `safeguard-container-resource-requests`
 - Readiness and liveness probes → `safeguard-probes-configured` *(warning-only — not blocked at admission; treat as informational)*
 - Image tag not `:latest` → `safeguard-images-no-latest`
 - `securityContext.privileged` not true → `safeguard-no-privileged-containers`
-- `allowPrivilegeEscalation` not true → `safeguard-no-privilege-escalation`
 - `capabilities.add` empty → `safeguard-container-capabilities`
```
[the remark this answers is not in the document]
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/common-fixes.md
```diff
@@ -32,21 +32,6 @@ containers:
...
-```yaml
-securityContext:
-  allowPrivilegeEscalation: false
-```
-
----
-
 ## `safeguard-container-capabilities` — Drop all capabilities
 
 **Before:**
```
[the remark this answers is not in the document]
NickKeller: removed

jongio (commented):
Follow-up on my earlier review. The PSS language updates and field path additions look clean.

A few remaining consistency items:

1. Cluster constraint count mismatch in the YAML header - see inline.

2. Stale numbers in the warning banner (SKILL.md line 249): still says "26 cluster-level constraints, 25 active Deployment Safeguards policies, ...5 Pod Security Baseline policies." With this PR those should be ~23 cluster constraints, 21 safeguard policies, and the standalone PSS section was removed entirely.

3. safeguard-host-probes missing ephemeralContainers field paths - see inline.

4. common-fixes.md host-probes After section still vague - see inline.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
     podSecurityRestricted: 42b8ef37-b724-4e24-bbc8-7a7708edfe00
 
 # =============================================================================
-# CLUSTER CONSTRAINTS (26 total, 3 are internal/HOBO)
+# CLUSTER CONSTRAINTS (21 total)
```
jongio: Header says `21 total` but I count 23 `- id:` entries under `clusterConstraints:`. The original had `26 total, 3 are internal/HOBO` - removing the 3 HOBO entries gives 23, not 21.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
```diff
@@ -17,7 +17,7 @@ description: "Assess Kubernetes workloads and cluster configuration for AKS Auto
 
 You are an AKS Automatic compatibility assessment agent. Your job is to evaluate whether Kubernetes workloads and cluster configurations are compatible with [AKS Automatic](https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic), identify issues, and help users fix them.
 
-AKS Automatic enforces **Deployment Safeguards** (25 active Deny policies), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
+AKS Automatic enforces **Deployment Safeguards** (21 active policies, some deny, some warn only, some auto-fixed), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
```
jongio: This line was updated for safeguard count (good), but `26 cluster-level configuration requirements` wasn't adjusted. The constraint spec now has 23 cluster entries after HOBO removal. Should this say 23?

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -257,110 +246,246 @@ safeguards:
...
+        - spec.initContainers[*].livenessProbe.httpGet.host
+        - spec.initContainers[*].readinessProbe.httpGet.host
+        - spec.initContainers[*].startupProbe.httpGet.host
+        - spec.initContainers[*].livenessProbe.tcpSocket.host
+        - spec.initContainers[*].readinessProbe.tcpSocket.host
+        - spec.initContainers[*].startupProbe.tcpSocket.host
+        - spec.initContainers[*].lifecycle.postStart.tcpSocket.host
+        - spec.initContainers[*].lifecycle.preStop.tcpSocket.host
+        - spec.initContainers[*].lifecycle.postStart.httpGet.host
+        - spec.initContainers[*].lifecycle.preStop.httpGet.host
```
jongio, line 465: Every other PSS safeguard in this spec includes `ephemeralContainers[*]` field paths alongside `containers[*]` and `initContainers[*]` (see `safeguard-no-privileged-containers`, `safeguard-container-capabilities`, `safeguard-enforce-apparmor`, etc). This entry only covers `containers` and `initContainers`. Are ephemeral containers exempt from the host-probes policy, or should their paths be added here for consistency?

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/common-fixes.md
```diff
@@ -170,6 +154,35 @@ readinessProbe:
...
+  containers:
+  - name: my-container
+    image: nginx:v1.2.3
+    livenessProbe:
+      httpGet:
+        host: "some host"
+```
+
+**After:**
+Suggest a fix similar to the previously mentioned `safeguard-probes-configured` fix
```
jongio: The After section still points to the probes-configured fix rather than showing the actual fix. For host-probes, the fix is removing the `host` field, not rewriting the entire probe. Every other entry in this file shows concrete before/after YAML. Consider something like:

```yaml
spec:
  containers:
  - name: my-container
    image: nginx:v1.2.3
    livenessProbe:
      httpGet:
        path: /healthz
        port: 8080
```

That way the `After` shows exactly what changes (`host` field gone, probe still works).

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
NickKeller: yes, ephemeralContainers are exempt since they don't have probes. This is directly from the pss spec 
https://kubernetes.io/docs/concepts/security/pod-security-standards/#baseline
image

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/common-fixes.md
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -234,14 +223,14 @@ safeguards:
     enforcement: warn  # Warning-only — deployments are admitted with a kubectl warning, not denied
     category: reliability
     check: Every container should have readinessProbe + livenessProbe (recommended best practice)
-    fix: Add probes (app-specific — HTTP, TCP, or exec) — recommended, not required for migration
+    fix: Add probes (app-specific — HTTP/TCP/exec/gRPC) — recommended, not required for migration
 
   - id: safeguard-csi-driver-storage-class
     policyId: 4f3823b6
     severity: requiresChanges
```
sabbour, line 230: Should this not be required then?

NickKeller: I don't think I fully understand your question - it's marked as requiresChanges, so it's required to be changed in order to work?

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
```diff
@@ -17,7 +17,7 @@ description: "Assess Kubernetes workloads and cluster configuration for AKS Auto
 
 You are an AKS Automatic compatibility assessment agent. Your job is to evaluate whether Kubernetes workloads and cluster configurations are compatible with [AKS Automatic](https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic), identify issues, and help users fix them.
 
-AKS Automatic enforces **Deployment Safeguards** (25 active Deny policies), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
+AKS Automatic enforces **Deployment Safeguards** (23 active policies, some deny, some warn only, some auto-fixed), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
```
jongio: This was flagged in my previous review - the constraint-spec header was updated from 26 to 23, but this line still says `26 cluster-level configuration requirements`. Should be 23 to match.

jongio (commented):
Incremental review of the 2 new commits. Count fix (21 -> 23), host-probes example, gRPC capitalization, and safeguard-host-probes cross-reference all look good. One item from my previous review is still open.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
jongio: I think @sabbour is asking: if `enforcement: warn` means deployments still succeed, should `severity` stay at `requiresChanges`? The SKILL.md already calls this one 'informational' (line ~33), so having `requiresChanges` here sends a conflicting signal to the agent. It'll tell users they must add probes to migrate, but probes aren't actually required for deployment.

NickKeller: This is for the csi drivers not the probes? It's required to be changed if the manifest is using the not-allowed provisioner.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
NickKeller: done

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
```diff
@@ -17,7 +17,7 @@ description: "Assess Kubernetes workloads and cluster configuration for AKS Auto
 
 You are an AKS Automatic compatibility assessment agent. Your job is to evaluate whether Kubernetes workloads and cluster configurations are compatible with [AKS Automatic](https://learn.microsoft.com/en-us/azure/aks/intro-aks-automatic), identify issues, and help users fix them.
 
-AKS Automatic enforces **Deployment Safeguards** (25 active Deny policies), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **26 cluster-level configuration requirements**.
+AKS Automatic enforces **Deployment Safeguards** (23 active policies, some deny, some warn only, some auto-fixed), **Pod Security Standards** (Baseline mandatory, Restricted optional), **2 active webhook mutators** that auto-fix certain fields at admission (resource-requests defaults and anti-affinity/topology-spread), and **23 cluster-level configuration requirements**.
```
jongio: If the 23 here includes the 2 mutators (line 249 breaks it down as 9 + 12 + 2), then "2 active webhook mutators" later in this sentence double-counts them. Either say "21 active safeguard policies" here and keep the separate mutator callout, or keep 23 and drop the separate mention.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
```diff
@@ -246,4 +246,4 @@ See `references/migration-guide-summary.md` for the full migration checklist.
 | `references/migration-guide-summary.md` | When user asks about migration steps or after assessment is complete |
 | `references/mcp-integration.md` | When troubleshooting MCP tool calls or debugging the fallback chain |
 
-> ⚠️ **Warning:** This skill bundles **constraint spec v1.1.1** (2026-03-15), covering 26 cluster-level constraints, 25 active Deployment Safeguards policies, 2 active webhook mutators, and 5 Pod Security Baseline policies. Always note the spec version in assessment output.
+> ⚠️ **Warning:** This skill bundles **constraint spec v1.1.1** (2026-03-15), covering 23 cluster-level constraints, 23 active Deployment Safeguards policies(9 best practices policies, 12 Pod Security Standards policies, and 2 active mutators) Always note the spec version in assessment output.
```
jongio: Two things:
1. Missing space before `(` - should be `policies (9 best practices...`
2. Missing period after `)` before "Always" - should be `...mutators). Always note...`

Also, "23 active Deployment Safeguards policies" here doesn't match the constraint-spec header which says "21 active policies" (line 180). The 2 mutators are in a separate section there. Aligning these would prevent confusion when the LLM agent cross-references the files.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
```diff
@@ -176,28 +176,17 @@ clusterConstraints:
...
-    severity: internal
-    note: AKS-managed, not customer-facing
-  - id: hobo-hosted-vm-taint
-    severity: internal
-    note: AKS-managed, not customer-facing
-
 # =============================================================================
-# WORKLOAD CONSTRAINTS — Deployment Safeguards (25 active policies)
-# Initiative: c047ea8e | Effect: Deny on Automatic
+# WORKLOAD CONSTRAINTS — Deployment Safeguards (21 active policies)
```
jongio, line 180: This says 21 active policies. SKILL.md now says 23, counting the 2 mutators from the mutations section below. If the umbrella count should include mutators, this header needs to be 23 (or "21 safeguards + 2 mutators"). If it should stay at 21, SKILL.md needs updating.

jongio (commented):
Latest commit addresses my count feedback (26->23 for cluster constraints). There's still a consistency gap between the safeguard counts in SKILL.md vs the constraint-spec - see inline comments.

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/constraint-spec-v1.yaml
NickKeller: How about I just remove all references to counts? That way we just tell the llm to look at everything here. 

I don't want to have to keep updating a comment in a markdown file every time we add a new constraint

NickKeller: Also mutators has it's own section, so it shouldn't be included in this section

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
NickKeller: fixed

jongio (commented):
Addresses my count feedback - 21 safeguard policies, 23 cluster constraints, and 2 mutators are now consistent between SKILL.md and the constraint spec. 9 best practices + 12 PSS = 21 checks out. Formatting fix (space, period) is clean too.

One thread from @sabbour about safeguard-host-probes severity vs enforcement is still open (line 230 of the constraint spec).

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/references/common-fixes.md
```diff
@@ -61,7 +46,6 @@ securityContext:
 securityContext:
   capabilities:
     drop: ["ALL"]
-  allowPrivilegeEscalation: false
 ```
 
 > ⚠️ **Warning:** If the app genuinely requires `NET_ADMIN` or similar, it is **incompatible** with AKS Automatic. Do not silently drop — explain the incompatibility and suggest redesign.
```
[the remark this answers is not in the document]
NickKeller: Fine for a suggested fix

plugin/skills/azure-kubernetes/azure-kubernetes-automatic-readiness/SKILL.md
```diff
@@ -185,9 +186,8 @@ Per-issue format:
 
 **Deterministic fixes** (have `suggestedPatch` — generate YAML diff directly):
 - `safeguard-container-resource-requests` — add `resources.requests`
-- `safeguard-no-privilege-escalation` — set `allowPrivilegeEscalation: false`
 - `safeguard-container-capabilities` — remove `capabilities.add`
-- `safeguard-allowed-seccomp-profiles` — add `seccompProfile: RuntimeDefault`
+- `safeguard-allowed-seccomp-profiles` — patch only when `seccompProfile.type: Unconfined` is present, or when the MCP `suggestedPatch` explicitly requires a seccomp change
```
jongio, line 190: Good change - the old text would make the agent add seccomp profiles to pods that already comply (`undefined/nil` is allowed). The two conditions (`Unconfined` present, or MCP `suggestedPatch` directs it) cover the right cases.

Follow-up to consider: `common-fixes.md` still shows a 'Before' with no seccomp at all. Adding a second 'Before' example showing `seccompProfile.type: Unconfined` would make the fix template match the primary offline trigger condition. Not blocking.

jongio (commented):
Seccomp trigger fix is correct - `undefined/nil` is compliant per the constraint spec, so only patching on `Unconfined` (or when MCP directs it) makes sense. One cross-reference note inline.

@sabbour's thread on constraint-spec line 230 is still open - worth resolving before merge.

jongio (approved):
Addresses my feedback from 5 review cycles. Counts are consistent (21 safeguards = 9 best practices + 12 PSS, 2 mutators called out separately, 23 cluster constraints). Seccomp fix guidance won't patch pods that already comply. Common-fixes examples look solid with proper spec context.

JasonYeMSFT:
@chandraneel  @gambtho Could you please review the change as code owner?

gambtho (approved):
Lgtm, Thanks Nick!

PR #1986, opened 2026-04-21
Move Azure Messaging Troubleshooting Info To New Directory

kashifkhan:
Move troubleshooting information inside diagnostics folder


plugin/skills/azure-diagnostics/troubleshooting/messaging/auth-best-practices.md
```diff
@@ -0,0 +1,128 @@
+# Azure Authentication Best Practices
```
jongio, line 1: This is byte-for-byte identical to `azure-messaging/references/auth-best-practices.md` (same SHA 6938d5f). The old copy isn't deleted in this PR, so both copies exist. If either gets updated, they'll silently diverge.

Should this be a proper move (delete the original) with the azure-messaging SKILL.md reference updated? Or is keeping two copies intentional?

plugin/skills/azure-messaging/SKILL.md
```diff
@@ -50,13 +50,16 @@ metadata:
...
-See [Service Troubleshooting Guide](references/service-troubleshooting.md) for ports, WebSocket fallback, IP firewall, private endpoints, and service tags.
+See the Service Troubleshooting Guide in the azure-diagnostics skill (`troubleshooting/messaging/service-troubleshooting.md`) for ports, WebSocket fallback, IP firewall, private endpoints, and service tags.
 
 ## SDK Troubleshooting Guides
 
-- **Event Hubs**: [Python](references/sdk/azure-eventhubs-py.md) | [Java](references/sdk/azure-eventhubs-java.md) | [JS](references/sdk/azure-eventhubs-js.md) | [.NET](references/sdk/azure-eventhubs-dotnet.md)
-- **Service Bus**: [Python](references/sdk/azure-servicebus-py.md) | [Java](references/sdk/azure-servicebus-java.md) | [JS](references/sdk/azure-servicebus-js.md) | [.NET](references/sdk/azure-servicebus-dotnet.md)
+SDK-specific troubleshooting guides are located in the azure-diagnostics skill under `troubleshooting/messaging/`:
+
+- **Event Hubs**: Python (`azure-eventhubs-py.md`) | Java (`azure-eventhubs-java.md`) | JS (`azure-eventhubs-js.md`) | .NET (`azure-eventhubs-dotnet.md`)
```
jongio: These changed from clickable markdown links to backticked filenames. An agent reading this SKILL.md can't follow backticked text to the actual docs.

If cross-skill relative links (like `../azure-diagnostics/troubleshooting/messaging/azure-eventhubs-py.md`) aren't supported by the skill loader, consider dropping the per-language filenames entirely. The routing sentence on line 57 already directs agents to the right skill and directory - the backticked names after each language add visual clutter without being navigable.

tests/azure-messaging/unit.test.ts
```diff
@@ -62,9 +62,9 @@ describe(`${SKILL_NAME} - Unit Tests`, () => {
       expect(skill.content).toContain("SDK Troubleshooting Guides");
     });
 
-    test("links to language-specific SDK references", () => {
-      expect(skill.content).toContain("references/sdk/azure-eventhubs-py.md");
-      expect(skill.content).toContain("references/sdk/azure-servicebus-py.md");
+    test("points to azure-diagnostics for SDK references", () => {
+      expect(skill.content).toContain("azure-diagnostics");
```
jongio: These assertions went from exact paths (`references/sdk/azure-eventhubs-py.md`) to generic substrings (`azure-diagnostics`, `troubleshooting/messaging/`). This would pass even if the actual reference text is malformed. Consider asserting a more specific expected pattern that matches what the SKILL.md actually contains now.

jongio (commented):
Three issues with this content move:

1. **Duplicate `auth-best-practices.md`** - identical file now exists at both `azure-messaging/references/` and `azure-diagnostics/troubleshooting/messaging/` (same git SHA 6938d5f). If either copy is updated independently, they'll diverge silently.

2. **Non-navigable SDK references** - `azure-messaging/SKILL.md` changed clickable markdown links to backticked filenames that agents can't follow. See inline comment.

3. **Loosened test assertions** - tests check generic substrings instead of verifiable paths. See inline comment.

jongio (approved):
Content move looks clean. All three issues from my earlier review are addressed: auth-best-practices.md is now a proper rename (not a copy), SDK references are simplified to a routing instruction, and tests validate file existence with existsSync.

One minor suggestion: the new troubleshooting/messaging/README.md doesn't surface auth-best-practices.md in its routing table or references. It's discoverable via the individual SDK guides, but listing it in the README would help agents find auth guidance without needing to load a language-specific doc first.

plugin/skills/azure-diagnostics/troubleshooting/messaging/auth-best-practices.md
jongio: This is now a proper rename - the old copy is deleted. Issue resolved.

kvenkatrajan (approved):

PR #2048, opened 2026-04-25
Cleanup preserve workspace

saikoumudi:
`preserveWorkspace: true` flag seems to be useful for local debugging and has no impact on test correctness since all assertions run inside test body before `afterEach` cleanup


JasonYeMSFT:
I agree that cleaning them up is the right thing to do right now. The option was added and these test cases had it set to true because we originally wanted to publish the artifacts. However, that attempt wasn't implemented due to concerns of leaking unwanted data since all the data we publish back then were via GitHub Action artifacts. We may revisit publishing those files later because we can publish data to private access only storage now.

JasonYeMSFT (dismissed):

saikoumudi:
@copilot resolve the merge conflicts in this pull request

tests/README.md
```diff
@@ -214,6 +214,14 @@ Test cases are grouped under the `describe` groups. It's commonly useful to use
 
 To learn more about how the CLI options work, check out `tests/scripts/run-tests.js`.
 
+### Local Workspace Cleanup
+
+Each integration test creates a temporary workspace under `os.tmpdir()` (e.g., `%TEMP%\skill-test-*` on Windows). By default, these are **automatically deleted** in `afterEach` after the test completes (pass or fail).
+
+To preserve workspaces for debugging, set `preserveWorkspace: true` in the `agent.run()` call.
+
+> **Note:** If a test times out and Jest kills the worker, `afterEach` may not run, leaving orphaned `skill-test-*` folders in your temp directory. Clean these up manually if needed.
```
JasonYeMSFT: Is this true? I feel I have seen the opposite, meaning jest executes the afterEach hook for test cases that time out.

saikoumudi: interesting! On my machine, they were just left out and manually deleted some of the folders after the test timed out.

saikoumudi: sometime the folder maybe empty but the folder lives

saikoumudi: removed the folder for now until we know for certain how the behavior is

JasonYeMSFT (approved):

PR #2078, opened 2026-04-28
fix: update toolbox sample link

XiaofuHuang:
## Description

Update toolbox sample link

## Checklist


jongio (commented):
Verified all 10 updated links - they all resolve correctly. The old `bring-your-own/responses/toolbox/main.py` path was genuinely broken (404), and the new `bring-your-own/responses/bring-your-own-toolbox/main.py` path works. The table restructure (sample-path-based instead of pattern-based) is cleaner and easier to navigate.

Two minor items:

1. **Version bump**: the PR checklist item "Version bumped in skill frontmatter (if skill files changed)" is unchecked. `microsoft-foundry/SKILL.md` is at 1.0.11 - worth a patch bump since this updates reference content.

2. **PR description**: the body is just the template with no description filled in. The commit message ("fix: update toolbox sample link") covers it, but a one-liner in the description helps reviewers who scan the PR list.

Neither is blocking. The link fixes and sample additions look correct.

JasonYeMSFT:
> Version bump: the PR checklist item "Version bumped in skill frontmatter (if skill files changed)" is unchecked. microsoft-foundry/SKILL.md is at 1.0.11 - worth a patch bump since this updates reference content.

We no longer bump skill versions manually, please leave it as is. I have a PR to remove that item from the PR template. 

> PR description: the body is just the template with no description filled in. The commit message ("fix: update toolbox sample link") covers it, but a one-liner in the description helps reviewers who scan the PR list.

None of the old links look obviously wrong. If you anticipate frequent changes like the one that broke any of the old links, would you like to pin each link to a specific commit instead of pointing them to the main branch?

JasonYeMSFT (approved):

jongio:
Thanks for the context on version bumps - I'll drop that suggestion going forward.

On link pinning: the tradeoff is stability vs maintenance cost. Pinning to a commit means links won't break when the samples repo reorganizes (which is exactly what this PR fixes), but it also means someone has to update the pinned SHAs whenever the samples get meaningful updates. Given that foundry-samples is actively evolving and these links are reference material for the skill (not user-facing docs), pointing to `main` seems like the right call - a broken link gets caught and fixed quickly (like this PR), while stale pinned links silently point to outdated code. I'd stick with `main` unless link breakage becomes a recurring problem.

kvenkatrajan (approved):

kvenkatrajan:
CC: @tendau - please review

tendau (approved):

PR #2114, opened 2026-04-30
Fix entra-agent-id test structure

JasonYeMSFT:
## Description

As title describes.

## Checklist

## Related Issues

#2113


JasonYeMSFT:
@jongio Could you please help review this PR? For some reason, I couldn't assign this PR to the other code owner for review in the GitHub UI.

jongio (approved):
Clean structural fix - moves the content-assertion describe block out of invalid nesting inside a test() callback to proper sibling scope. Variable scoping is correct (agent accessible at parent scope, each test creates its own agentMetadata). Tests that were silently not running will now execute.

PR #2116, opened 2026-04-30
fix: sanitize untrusted artifact fields in GEPA quality score comment workflow 

spboyer:
## Summary

Fixes the unauthenticated artifact injection / Markdown table spoofing reported in `pipelines/gepa-quality-score-comment.yml`.

cc @kvenkatrajan

## What Changed

`pipelines/gepa-quality-score-comment.yml` — comment-step script now treats the artifact as untrusted:

- New `sanitizeMd()` helper strips `\r`, `\n`, `|`, `` ` ``, `<`, `>` from any string field and clamps to 100 chars.
- New `finiteNumber()` helper validates `quality_score` / `trigger_accuracy` as actual finite numbers; type confusion falls back to `N/A` and a `❓` icon.
- Caps results to 200 rows (oversized-comment / DoS guard) and notes the truncation in the body.
- Filters out non-object items defensively.
- `JSON.parse` is wrapped in `try/catch` so a malformed artifact logs a warning and exits cleanly instead of failing the privileged step in an undefined state.

No behavioral change for well-formed artifacts produced by the legitimate scoring workflow.

## Testing

Replayed the MSRC PoC payload locally against the new logic:

```json
[{"skill": "GenuineSkill\n| Malicious-Code | ✅ 1.00 | 1.00 | TIU |", "quality_score": 0.01, ...}]
```

Before: produced a second row showing `Malicious-Code` with `✅ 1.00`.
After: produces a single row, the injected newline/pipes are collapsed into the skill cell, and the actual `❌ 0.01` score is reported.

Also exercised: string `quality_score` (type confusion), backtick injection, CR-only injection, non-string `skill`, 500-char `skill`. All neutralized; row count always equals input item count.

## Follow-up (not in this PR)

`.github/workflows/pr-comment.yml` exhibits the same trust-boundary pattern — it reads `combined-report.md` from a PR-controlled artifact and posts the markdown verbatim. Worth reviewing whether to render that under a `<details>` block with a header that disclaims its origin, or to validate the report before posting. Happy to file a follow-up issue if useful.

---

### DIFF
--- a/pipelines/gepa-quality-score-comment.yml
+++ b/pipelines/gepa-quality-score-comment.yml
@@ -52,8 +52,38 @@ jobs:
         with:
           script: |
             const fs = require('fs');
-            const results = JSON.parse(fs.readFileSync('score-results.json', 'utf8'));
-            const items = Array.isArray(results) ? results : [results];
+
+            // The score-results.json artifact originates from a workflow that ran on a
+            // pull_request (potentially from a fork) and is therefore UNTRUSTED.
+            // Every field must be sanitized before being interpolated into Markdown
+            // to prevent table-injection / comment-spoofing (MSRC 114952).
+            const MAX_ITEMS = 200;
+            const MAX_FIELD_LEN = 100;
+
+            function sanitizeMd(input) {
+              if (typeof input !== 'string') return 'N/A';
+              // Strip table-breaking, code-fence, and HTML-ish characters; clamp length.
+              const cleaned = input.replace(/[\r\n|`<>]/g, ' ').trim().slice(0, MAX_FIELD_LEN);
+              return cleaned.length > 0 ? cleaned : 'N/A';
+            }
+
+            function finiteNumber(v) {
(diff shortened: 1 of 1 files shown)


pipelines/gepa-quality-score-comment.yml
```diff
@@ -52,29 +52,66 @@ jobs:
         with:
           script: |
             const fs = require('fs');
-            const results = JSON.parse(fs.readFileSync('score-results.json', 'utf8'));
-            const items = Array.isArray(results) ? results : [results];
+
+            // The score-results.json artifact originates from a workflow that ran on a
+            // pull_request (potentially from a fork) and is therefore UNTRUSTED.
+            // Every field must be sanitized before being interpolated into Markdown
+            // to prevent table-injection / comment-spoofing (MSRC 114952).
```
JasonYeMSFT, line 59: Let's not expose details of the specific case.

spboyer:
Closing — re-opening with a clean branch name and content (no internal advisory references). New PR will follow.

PR #2117, opened 2026-04-30
fix: sanitize untrusted artifact fields in GEPA quality score comment workflow

spboyer:
## Summary

Fixes an unauthenticated artifact injection / Markdown table spoofing in `pipelines/gepa-quality-score-comment.yml`. The `workflow_run`-triggered comment job consumes `score-results.json` produced by a `pull_request` workflow (which can run from a fork) and previously interpolated its fields directly into a Markdown comment posted to the PR. Because the `skill` field in that artifact ultimately reflects PR-author-controlled content (a file path / SKILL.md identifier from the changed skill), an attacker could inject newline + pipe characters into it to fabricate table rows and spoof a passing quality score.

cc @kvenkatrajan @JasonYeMSFT

## What Changed

`pipelines/gepa-quality-score-comment.yml` — the comment-step script now treats the artifact as untrusted:

- New `sanitizeMd()` helper strips `\r`, `\n`, `|`, `` ` ``, `<`, `>` from any string field, inserts a zero-width space after `@` so a value like `@org/team` can't ping users, and backslash-escapes `!`, `[`, `]`, `(`, `)` so a value can't render as a clickable link or embedded image. Output is clamped to 100 chars.
- New `finiteNumber()` helper validates `quality_score` / `trigger_accuracy` as actual finite numbers; type confusion falls back to `N/A` and a `❓` icon.
- Caps results to 200 rows (`shouldTruncate` / `truncatedItems`); truncated runs explicitly note `<displayed>/<MAX_ITEMS>` and the omitted-row count in the summary line.
- Filters out non-object items defensively.
- `JSON.parse` is wrapped in `try/catch` so a malformed artifact logs a warning and exits cleanly instead of failing the privileged step in an undefined state.

No behavioral change for well-formed artifacts produced by the legitimate scoring workflow.

## Testing

Replayed the proof-of-concept payload locally against the new logic:

```json
[{"skill": "GenuineSkill\n| Malicious-Code | ✅ 1.00 | 1.00 | TIU |", "quality_score": 0.01, ...}]
```

Before: produced a second row showing `Malicious-Code` with `✅ 1.00`.
After: produces a single row, the injected newline/pipes are collapsed into the skill cell, and the actual `❌ 0.01` score is reported.

Also exercised: string `quality_score` (type confusion), backtick injection, CR-only injection, non-string `skill`, 500-char `skill`, `@org/team` mention, `[click](https://evil)` link injection, `img` image injection. All neutralized; row count always equals input item count.

### DIFF
--- a/pipelines/gepa-quality-score-comment.yml
+++ b/pipelines/gepa-quality-score-comment.yml
@@ -52,8 +52,38 @@ jobs:
         with:
           script: |
             const fs = require('fs');
-            const results = JSON.parse(fs.readFileSync('score-results.json', 'utf8'));
-            const items = Array.isArray(results) ? results : [results];
+
+            // The score-results.json artifact originates from a workflow that ran on a
+            // pull_request (potentially from a fork) and is therefore UNTRUSTED.
+            // Every field must be sanitized before being interpolated into Markdown
+            // to prevent table-injection / comment-spoofing.
+            const MAX_ITEMS = 200;
+            const MAX_FIELD_LEN = 100;
+
+            function sanitizeMd(input) {
+              if (typeof input !== 'string') return 'N/A';
+              // Strip table-breaking, code-fence, and HTML-ish characters; clamp length.
+              const cleaned = input.replace(/[\r\n|`<>]/g, ' ').trim().slice(0, MAX_FIELD_LEN);
+              return cleaned.length > 0 ? cleaned : 'N/A';
+            }
+
+            function finiteNumber(v) {
+              return typeof v === 'number' && Number.isFinite(v) ? v : null;
+            }
+
+            let raw;
+            try {
+              raw = JSON.parse(fs.readFileSync('score-results.json', 'utf8'));
+            } catch (e) {
+              core.warning(`Could not parse score-results.json: ${e.message}`);
+              return;
+            }
+
+            const items = (Array.isArray(raw) ? raw : [raw]).filter(
+              r => r && typeof r === 'object' && !Array.isArray(r)
+            );
+            const truncated = items.length > MAX_ITEMS;
+            const limited = truncated ? items.slice(0, MAX_ITEMS) : items;
             const prNumber = parseInt('${{ steps.pr.outputs.number }}', 10);
 
             const marker = '<!-- gepa-quality-score -->';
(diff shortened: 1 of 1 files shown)


pipelines/gepa-quality-score-comment.yml
```diff
@@ -52,29 +52,66 @@ jobs:
...
+              raw = JSON.parse(fs.readFileSync('score-results.json', 'utf8'));
+            } catch (e) {
+              core.warning(`Could not parse score-results.json: ${e.message}`);
+              return;
+            }
+
+            const items = (Array.isArray(raw) ? raw : [raw]).filter(
+              r => r && typeof r === 'object' && !Array.isArray(r)
+            );
+            const truncated = items.length > MAX_ITEMS;
```
JasonYeMSFT:
```suggestion
            const shouldTruncate = items.length > MAX_ITEMS;
```

pipelines/gepa-quality-score-comment.yml
```diff
@@ -52,29 +52,66 @@ jobs:
...
+            const limited = truncated ? items.slice(0, MAX_ITEMS) : items;
```
JasonYeMSFT:
```suggestion
            const truncatedItems = truncated ? items.slice(0, MAX_ITEMS) : items;
```

spboyer: Renamed in 2707bc9 (cascaded into the for-loop and summary line as well).

pipelines/gepa-quality-score-comment.yml
```diff
@@ -52,29 +52,66 @@ jobs:
...
+              const icon = qs === null ? '❓' : qs >= 0.8 ? '✅' : qs >= 0.5 ? '⚠️' : '❌';
+              if (qs !== null && qs >= 0.8) passing++;
+              body += `| ${skill} | ${icon} ${q} | ${t} | ${tests} |\n`;
             }
 
-            const passing = items.filter(r => (r.quality_score || 0) >= 0.8).length;
-            body += `\n**${passing}/${items.length}** skills at quality ≥ 0.80\n`;
+            body += `\n**${passing}/${limited.length}** skills at quality ≥ 0.80\n`;
+            if (truncated) {
+              body += `\n_Note: ${items.length - MAX_ITEMS} additional row(s) omitted._\n`;
```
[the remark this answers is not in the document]
spboyer: Adopted in 2707bc9 — split the summary so the truncated path now reads `X/Y displayed skills (first 200 results) at quality ≥ 0.80` followed by the omitted-row note, while the non-truncated path keeps the original phrasing.

pipelines/gepa-quality-score-comment.yml
spboyer: Renamed in 2707bc9. Agreed — `truncated` reads like past tense; `shouldTruncate` is clearer.

spboyer:
Pushed 2707bc9 addressing the open review threads. The Copilot bot sanitizer thread did not accept an inline reply (API hung), so summarizing here: sanitizeMd now also inserts a zero-width space after @ and backslash-escapes the link/image markdown chars (! [ ] ( )) so a malicious skill field cannot ping users or render clickable links. The other three threads (truncation summary, shouldTruncate rename, truncatedItems rename) were replied to inline.

JasonYeMSFT (approved):

spboyer:
@JasonYeMSFT good point — you are right. The GEPA case is exposed because `r.skill` is a free-form string that ultimately reflects PR-author content (a SKILL.md identifier from a changed skill), and it was being interpolated raw. `pr-comment.yml` does not have that property: the report body is computed numbers wrapped in a static table; the only PR-author-controlled value is the file path, and that is wrapped in backticks. Walking back the speculative follow-up — I have removed it from the PR description so it does not mislead future readers.

PR #2122, opened 2026-05-01
update the path in script to support claude marketplace file refs

saikoumudi:
Tested hooks for Azure Plugin in official Claude Marketplace. The file reference path needs to be updated to emit file reference name telemetry. Update powershell script for consistency and claude only refers to bash file.


jongio (commented):
Both scripts correctly extend Claude cache path matching to support `claude-plugins-official` alongside `azure-skills`. The PS1 regex alternation and the new bash glob in `is_azure_skills_path()` are consistent, and the downstream path extraction still works for both directories. Added documentation is accurate.

JasonYeMSFT (approved):

PR #2132, opened 2026-05-01
Always set deploy test follow up timeout

JasonYeMSFT:
## Description

When I increased the timeout in https://github.com/microsoft/GitHub-Copilot-for-Azure/pull/2115, I forgot to pass these follow up timeout parameters. In deploy tests, the actual time consuming deployment happens after the follow up. Without setting them explicitly, they fallback to the 30 minute default. By passing a longer timeout for the deploy test follow up, there will be even less chance for an agent to time out if it's still making progress.

## Checklist


jongio (approved):
Reviewed 1 changed file. Nothing to flag - straightforward fix ensuring all deploy test follow-ups use the explicit 40-minute timeout instead of the 30-minute default.

PR #2154, opened 2026-05-04
fix: update foundry-agent invoke skill for invocations protocol

Kishore712:
## Summary

Update the foundry-agent invoke skill to clearly document how the invocations protocol works — the platform is pure pass-through, input and output are raw bytes defined by the container developer.

### `invoke.md`
- Updated **Protocols** section: explicitly states invocations is bytes in, bytes out with no platform transformation
- Updated **Step 3**: added OpenAPI spec discovery as the preferred method to learn expected input format
- Updated **Step 4**: clarified `conversationId` and `stream` do not apply to invocations
- Added **schema mismatch** error to Error Handling table

### New `references/invocations-protocol.md`
- Full I/O contract comparison table (responses vs invocations)
- 3-step schema discovery workflow: fetch OpenAPI spec → inspect source code → ask user
- Documents the `GET .../invocations/docs/openapi.json` endpoint
- Concrete examples for both protocols
- Common use cases and error handling

## Motivation

The invoke skill previously gave vague guidance for the invocations protocol ("payload shape depends on what the hosted agent code expects"). This caused confusion during local testing — agents need clear instructions that invocations is raw bytes pass-through and how to discover the expected input format.

## Validation

- ✅ `npm run build` — passes
- ✅ `npm run tokens check` — within limits
- ✅ `npm run references` — no broken/orphaned references
- ✅ All 32 invoke skill tests pass (unit + triggers)


plugin/skills/microsoft-foundry/foundry-agent/invoke/invoke.md
```diff
@@ -1,145 +1,125 @@
...
-#### Session Support for Hosted Agents
-
-Hosted agents accept a 25 character alphanumeric `sessionId` parameter. Sessions are **sticky**: they route the request to the same underlying compute resource, so the agent can reuse state stored on that compute across multiple turns.
+## Workflow
 
-Rules:
-1. You MUST generate a unique `sessionId` before making the first `agent_invoke` call.
-2. If you have a session ID, you MUST include it in every subsequent `agent_invoke` call for that conversation.
-3. When the user explicitly requests a new session, create a new `sessionId` and use it for the rest of the `agent_invoke` calls.
+### Step 1: Verify Agent Readiness
```
jongio, line 55: Step 3 explains invocations is ytes in, bytes out and points to the references doc for details - but doesn't include a minimal inline example of calling gent_invoke with protocol: 'invocations'. Consider adding a 2-line example here (mirroring what's in 
eferences/invocations-protocol.md) so users following the workflow sequentially can see the shape without jumping to a reference. Not blocking - just reduces friction.

jongio (commented):
Solid documentation restructuring. The invoke skill now clearly separates the two protocols and properly delegates session management and file operations to reference docs. The rename from `create.md` to `create-hosted.md` is handled cleanly with all cross-references updated.

One observation the bot review didn't catch: the old `az rest` fallback for invocations is removed, but the inline workflow (Step 3) doesn't show a concrete `agent_invoke` call with `protocol: 'invocations'`. The example lives in `references/invocations-protocol.md`, which is fine for progressive disclosure - but a one-liner example or snippet in Step 3 itself (where users will first encounter invocations guidance) would reduce the back-and-forth for someone following the workflow linearly.

The Quick Reference conversation row inconsistency (already flagged) is the only real issue - the rest reads well.

plugin/skills/microsoft-foundry/foundry-agent/create/create-hosted.md
```diff
@@ -279,6 +279,21 @@ Apply these to both greenfield and brownfield projects:
 
 6. **Tool integration** — Hosted agents access tools through [Foundry Toolbox](references/toolbox.md), NOT by wiring tools directly. If the user needs tools (web search, AI search, code execution, MCP servers, etc.), follow the toolbox integration guide. The toolbox provides a single MCP-compatible endpoint that handles credential injection and tool discovery.
 
+7. **Reserved environment variables** — The Foundry platform injects environment variables into every hosted agent container at startup (per [container-image-spec §3](https://github.com/rapida_microsoft/foundrysdk_specs/blob/rapida/hosted-agent-container-spec/specs/hosted-agents/container-spec/docs/container-image-spec.md)). You MUST NOT generate, suggest, or configure any of these in `.env` files, `agent.yaml` `environment_variables`, or application code:
```
ankitbko: The spec repo is not public. Mention all the env we inject here

ankitbko (changes requested):

plugin/skills/microsoft-foundry/foundry-agent/invoke/invoke.md
Kishore712: Addressed this and added a minimal inline example  after the schema discovery steps

plugin/skills/microsoft-foundry/foundry-agent/create/create-hosted.md
Kishore712: Removed the private spec link and listed the env variables injected by platform

jongio (dismissed):
Addresses all prior feedback cleanly. The inline invocations example in Step 3, clarified Quick Reference row, and documented FOUNDRY_* variables all look correct. The test assertion is now precise. No new issues.

ankitbko (dismissed):

ankitbko (approved):

XiaofuHuang (approved):

PR #2197, opened 2026-05-08
Support publishing adhoc integration test results

JasonYeMSFT:
## Description

This would allow people to run adhoc integration tests and collect the test artifacts.

## Checklist

### DIFF
--- a/.github/workflows/test-all-integration.yml
+++ b/.github/workflows/test-all-integration.yml
@@ -46,6 +46,11 @@ on:
         required: false
         type: boolean
         default: false
+      ReportStorageAccount:
+        description: 'Optional: Azure Storage account name to publish reports to (only used for manual runs)'
+        required: false
+        type: string
+        default: ''
 
 # SCHEDULE CONFIGURATION
 # Customize the inputs for each scheduled run by modifying these env vars.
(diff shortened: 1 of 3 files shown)


.github/workflows/test-all-integration.yml
```diff
@@ -46,6 +46,11 @@ on:
         required: false
         type: boolean
         default: false
+      ReportStorageAccount:
+        description: 'Optional: Azure Storage account name to publish reports to (only used for manual runs)'
```
tmeschter: Do we want an entirely different storage account, or a different folder in the same storage account?

.github/workflows/test-all-integration.yml
```diff
@@ -345,3 +350,26 @@ jobs:
             --source reports/test-run-all-integration/ \
             --auth-mode login \
             --overwrite
+
+      # Publish reports to a user-provided storage account for manual runs.
+      # Note: The managed identity must have write permission to the target storage account.
+      - name: Publish report to Azure Storage (manual run)
+        if: always() && github.event_name == 'workflow_dispatch' && inputs.ReportStorageAccount != ''
```
tmeschter: What does the `always()` do here?

tmeschter: Do we already have logic that prevents manual runs from being stored in our primary storage account, or do we want them stored in both?

JasonYeMSFT: always() is needed so even if prior steps fail, this step will still be evaluated and run if its condition pass. When a prior step failed, we still want to publish the available data.

.github/workflows/test-all-integration.yml
JasonYeMSFT: I think you made a good point. I'll change to use a different container in the same storage account. I'll add an environment variable to make viewing data from a different container possible in the dashboard app.

The storage layout already supports storing multiple results of a test case so having multiple runs write data for the same test case shouldn't be an issue. However, the dashboard app is built with the assumption that there is up to one result per test case per day. We can update the application code to support the new scenario if it becomes an issue.

.github/workflows/test-all-integration.yml
JasonYeMSFT: It's only the condition `github.event_name == 'workflow_dispatch'` that prevents manual runs from publishing data to the primary storage account for now. I changed the code to store both auto and manual runs data in the same storage account, but in different containers.

.github/workflows/test-all-integration.yml
```diff
@@ -46,6 +46,11 @@ on:
         required: false
         type: boolean
         default: false
+      ReportStorageAccount:
+        description: 'Optional: Azure Storage account name to publish reports to (only used for manual runs)'
+        required: false
+        type: string
+        default: ''
```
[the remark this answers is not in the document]
JasonYeMSFT: This new parameter is removed.

tmeschter (approved):

PR #2270, opened 2026-05-15
fix azure-upgrade: improve Java Azure SDK BOM migration reliability

XiaofeiCao:
## Summary

- Tighten Java Azure SDK BOM migration guidance so agents must resolve the latest stable `azure-sdk-bom` from the Azure SDK source of truth during planning and validation.
- Update `upgrade_bom.py` to resolve the latest BOM internally, reject explicit version pins, enforce the supported stable-version floor, and keep the `latest` token only as a compatibility alias.
- Refresh Java migration guidance for Maven, Gradle, TOML catalogs, and validation so the resolved `TARGET_AZURE_SDK_BOM_VERSION` is carried through the full migration workflow.
- Refactor Java migration integration tests to use remote sample projects from `java-update-examples` via sparse checkout instead of storing local fixtures.

## Benchmark impact

The benchmark pass rate improved from 60%+ to 94%+:

- Before: https://msbenchapp.azurewebsites.net/run-analysis/25540122197
- After: https://msbenchapp.azurewebsites.net/run-analysis/25898537401

## Validation

- `cd scripts && npm run references`
- `git --no-pager diff --check`

Note: `cd scripts && npm run frontmatter` currently reports existing placeholder-version validation errors for plugin skills that use `metadata.version: "0.0.0-placeholder"`; this is unrelated to the changes in this PR.

integration tests pass:
image


weidongxu-microsoft (commented):
agent comment
```
 2. _parse_stable_semver return type hint — Uses tuple[int, int, int] (Python
  3.9+ syntax). The docstring says Python 3.8+ is required. Should use Tuple[int, int, int] from typing for 3.8 compat, or from __future__ import 
annotations.
 3. _resolve_latest_bom_version type hint — Uses str | None (Python
  3.10+), same 3.8 compat issue. Should use Optional[str]
 4. Network dependency in script — upgrade_bom.py now always fetches from GitHub on every run. No caching or timeout handling visible in the diff. If the
fetch fails mid-migration, the error path should be clear.
```

tests/azure-upgrade/integration.test.ts
```diff
@@ -48,6 +49,7 @@ const SKILL_NAME = "azure-upgrade";
 // generous timeout as other migration integration tests.
 const javaMigrationTimeoutMs = 2700000;
 const FOLLOW_UP_PROMPT = ["Continue with recommended options until complete."];
+const JAVA_UPDATE_EXAMPLES_REPO = "https://github.com/weidongxu-microsoft/java-update-examples.git";
```
weidongxu-microsoft, line 53: you may want to pin to SHA of it

XiaofeiCao: done

XiaofeiCao:
> agent comment
> 
> ```
>  2. _parse_stable_semver return type hint — Uses tuple[int, int, int] (Python
>   3.9+ syntax). The docstring says Python 3.8+ is required. Should use Tuple[int, int, int] from typing for 3.8 compat, or from __future__ import 
> annotations.
>  3. _resolve_latest_bom_version type hint — Uses str | None (Python
>   3.10+), same 3.8 compat issue. Should use Optional[str]
>  4. Network dependency in script — upgrade_bom.py now always fetches from GitHub on every run. No caching or timeout handling visible in the diff. If the
> fetch fails mid-migration, the error path should be clear.
> ```

Thanks. 
- updated compatibility to 3.10+, to align with SKILL.md
- cached HTTP result
- explicit instructions on script failure

tests/azure-upgrade/integration.test.ts
```diff
@@ -48,6 +50,8 @@ const SKILL_NAME = "azure-upgrade";
 // generous timeout as other migration integration tests.
 const javaMigrationTimeoutMs = 2700000;
```
jongio, line 51: **[MEDIUM]** Tests now depend on `weidongxu-microsoft/java-update-examples.git` - a personal GitHub repo. If that account is deactivated, the repo goes private, or the commit gets force-pushed away, all three Java migration integration tests break with no fallback.

Worth moving these fixtures to an org-owned repo (e.g. under `Azure-Samples/` or `microsoft/`) or documenting an owner/continuity plan for this dependency.

jongio (commented):
The BOM-resolution hardening and the switch from explicit version pins to auto-resolve look good - cleaner UX for agents and prevents stale versions. One concern about test resilience below.

tests/azure-upgrade/integration.test.ts
```diff
@@ -233,21 +237,33 @@ describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
...
-            workspace,
-            { recursive: true }
-          );
+          await cloneRepo({
+            repoUrl: JAVA_UPDATE_EXAMPLES_REPO,
+            targetDir: workspace,
+            depth: 1,
+            sparseCheckoutPath,
+          });
+          await simpleGit({ baseDir: workspace, binary: "git" }).checkout(commitId);
```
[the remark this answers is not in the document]
jongio: Seconding this. The `cloneRepo` utility passes both `--depth 1` and `--filter=tree:0` when `sparseCheckoutPath` is set. After the initial sparse checkout of HEAD, the subsequent `simpleGit.checkout(commitId)` won't find the pinned commit in the shallow history unless it happens to be the current tip. Once the upstream repo gets a new commit, this test starts failing.

Easiest fix: drop `depth: 1` from the `cloneRepo` call (the treeless filter already saves bandwidth), or fetch the specific commit before checking it out.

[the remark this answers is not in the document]
XiaofeiCao: Thanks, this is helpful. It happened to work due to it being the latest commit ID.

tests/azure-upgrade/integration.test.ts
XiaofeiCao: It's samples either collected from public GitHub repository, or written to reflect edge cases. It may not be a good candidate for `Azure-Samples` or `microsoft` org...

fanyang-mono (approved):

fanyang-mono (changes requested):
This PR needs to be signed off by @XiaofeiCao's team.

tests/azure-upgrade/integration.test.ts
weidongxu-microsoft: Thinking about whether azure-sdk org would be a better location. Or maybe just put to azure org (there is already quite some strangely named repo there...). Let's ask around.

XiaofeiCao: Thanks, issue logged: https://github.com/Azure/azure-sdk-for-java-pr/issues/1484

Let's have this PR merged first.

weidongxu-microsoft (approved):
SDK dev has started moving the test data to e.g. Azure org, maybe `azure-sdk-for-java-update-test-data`. It likely going to take a few more days, as we discuss the org and repo name.

fanyang-mono (approved):

XiaofeiCao:
@jongio Would you help approve the PR? Thanks!

JasonYeMSFT (approved):

JasonYeMSFT:
@MadhuraBharadwaj-MSFT @saikoumudi Please review as code owner.

saikoumudi (approved):

PR #2297, opened 2026-05-18
feat(azure-compute): add VM creation workflow with approval gate before deploy

deankroker:
## What this does

Before this PR, asking Copilot for Azure to "create a VM" jumped straight to generating Bicep or running `az vm create`. The user had no chance to review what was about to be deployed, and for VM scale sets the model often missed the intent entirely.

This PR adds a guided workflow so the same request now:

1. Asks 2–3 short questions to figure out what the user actually needs (region, size, OS, network exposure)
2. Shows a **Plan Card** — a 14–21 row summary of every parameter the deployment will use, with each value labeled by where it came from (user input, default, recommendation)
3. Waits for the user to approve, edit, or change format — in a single picker, not three sequential prompts
4. Only then generates Bicep / Terraform / `az` CLI / MCP calls

It also recognizes VMSS intent without the user having to say "VMSS" (e.g. "fleet that autoscales" → Flexible orchestration scale set), and refuses to deploy templates that would leave SSH open to the internet.

## What's in the PR

**Workflow + supporting docs**
- `57789bd` — new `vm-creator` workflow under `plugin/skills/azure-compute/`, with depth-probe references (beginner / cost / networking / security / spec) so the question set adapts to user expertise. Output adapters for `az` CLI, Bicep, Terraform, and Azure MCP. Trims `vm-recommender.md` and pulls verbose `web_fetch` rules into a reference file.
- `a7c7954` — Plan Card approval is now one batched picker (approve / edit / change format) instead of three sequential questions.

**Plumbing**
- `e729b2f` — switches the Azure MCP CLI to `--mode single`, collapsing ~26 per-service tool schemas (~30K tokens) behind one routing tool (~3K tokens). Same APIs, same auth — the schemas just don't load until the tool is called. This is what makes the workflow fit in context without truncating.

**Eval coverage**
- `4592b1f` — `evals/azure-compute/eval.yaml` with 16 stimuli covering router accuracy, depth-probe branches, VMSS recognition, Plan Card gate, and output adapter coverage. Smoke tier (5 × 3) was 3/5 flaky before; this PR brings it to 5/5 passing 3-for-3.

**Follow-ups during review**
- `710e3f7` — fixes Skill Structure CI: a few cross-file references pointed at paths that didn't exist after the workflow split.
- `a590be1` — rewrites the skill frontmatter `description:` from a `>-` folded scalar to an inline double-quoted string. `skills.sh` only accepts the inline form.
- `ef08203` — removes the `'*'` default on `sshSourceAddressPrefix` (Bicep) and `ssh_source_address_prefix` (Terraform). The templates now require the caller to pass their own IP or a trusted CIDR; opening port 22 to the entire internet has to be an explicit, accepted choice. Resolves the open Copilot security suggestions.

## Why it matters

The visible change: a user asking for a VM gets to see and edit what's about to be deployed, instead of finding out after the fact. The eval suite locks the behavior in so it doesn't regress.

The invisible change (`e729b2f`): the whole workflow used to run out of context room around turn 3 because the Azure MCP server was loading every service's tool schema up front. With `--mode single`, schemas load on demand and the workflow holds end-to-end.

## Test plan

- [x] `vally run evals/azure-compute/eval.yaml --tier smoke` — 5/5 stimuli × 3 trials
- [x] Manual: "I need a small Ubuntu VM to test something" → beginner depth probe → 2 questions → Plan Card → batched picker. **PASS** (5 turns, $0.42)
- [x] Manual: "set up an autoscaling web tier for ~500 rps" → VMSS recognized without user saying VMSS → Flexible orchestration recommended. **PASS** (4 turns, $0.39)
- [x] Manual: "apply via MCP" end-to-end — verify the single MCP routing tool stays responsive past turn 3. **PASS** through 16 turns (quota validation caught zero D2s_v5 quota, model auto-switched to B1s). Apply step was cancelled at the permission prompt, which is orthogonal to the schema-cost fix being tested.
- [x] No regressions in `vm-recommender` standalone (recommend-only, no hand-off to creator). **PASS** (1 turn, $0.30)
- [x] All 11 CI checks pass on the latest commit (`ef08203`)

Manual items run via the `claude-copilot` proxy on `claude-sonnet-4.6` with the plugin loaded from a working tree clone.


deankroker:
Pushed `ef08203` to address the open Copilot review suggestions. Audited all 6; 3 had real underlying problems that needed code changes, 3 were already correct as written.

**Fixed (`ef08203`):**

| File | Change |
|---|---|
| `examples/bicep/main.bicep` | `sshSourceAddressPrefix` no longer defaults to `'*'` — now required; description explains the tradeoff |
| `examples/terraform/variables.tf` | `ssh_source_address_prefix` no longer defaults to `"*"` — now required; description explains the tradeoff |
| `examples/terraform/README.md` | Quickstart derives `MY_IP=$(curl -s ifconfig.me)/32` and passes it explicitly to plan/apply; variables table marks the var required; Notes call out why; Cleanup command updated to match |

Templates now refuse to deploy without an explicit source prefix — the open-to-internet path requires the user to pass `"*"` and accept the risk.

**Audited, no change needed:**

| Comment | Why no change |
|---|---|
| `plan-card.md` "4 options" vs 6 listed | File already says "6 options" on the relevant line; suggestion was based on stale context |
| `mcp-apply.md` `familyPrefix` → `family` | The doc uses `familyPrefix` for `compute_vm_list-skus` (correct param for that command) and `family` for `compute_vm_check-quota` (correct param for *that* command) — confirmed against `references/mcp-tools.md` |
| `depth-probe/beginner.md` SSH/RDP defaults to `*` | Beginner path already defaults to the user's current public IP via `curl -s ifconfig.me`; `*` is only the fallback if detection fails, and is always flagged with ⚠ |

Watching CI; will flag if anything regresses.

evals/azure-compute/eval.yaml
```diff
@@ -0,0 +1,800 @@
...
+    graders:
+      # Must engage vm-creator. The workflow has multiple legitimate surface
+      # phrasings: it may say "Plan Card" / "provision" / "create the VM",
+      # or it may enter the Beginner fast-path ("Single VM, fast-path", "I'll
+      # default silently") and only later render the Plan Card.
+      - type: output-matches
+        config:
+          pattern: "(?i)plan card|vm.?creator|provision|create.*vm|single.?vm|fast.?path|default.*silently|beginner"
+      # Must NOT route to azure-prepare (that's for apps, not bare VMs)
+      - type: output-not-matches
```
JasonYeMSFT: If you want to validate skill routing you should use the skill-invocation grader.
https://literate-engine-r3wnl4v.pages.github.io/reference/graders/skill-invocation/

evals/azure-compute/eval.yaml
```diff
@@ -0,0 +1,800 @@
...
+        config:
+          pattern: "(?i)0\\.0\\.0\\.0/0.*22"
+
+  # ════════════════════════════════════════════════════════════════════
+  # FULL TIER — WORKFLOW COMPLIANCE
+  # ════════════════════════════════════════════════════════════════════
+
+  # ── plan-card-gate-001 ──
+  # Mirrors azure-prepare's plan-first-001. User explicitly asks the skill
+  # to skip planning. The vm-creator Step 4 GATE forbids this — Plan Card
```
JasonYeMSFT: What's the reason behind forbidding the agent to do what the user explicitly asked it to? In general, we only want apply our recommendations or best practices when the user doesn't explicitly issue commands like this test prompt. It would be better for us to change the test prompt for such a fuzzy scenario.

evals/azure-compute/eval.yaml
```diff
@@ -0,0 +1,800 @@
...
+      - type: output-matches
+        config:
+          pattern: "(?i)plan|gather|step|render|propos|question|detail|default|design|configure|recommend|fast.?path|scale.?set|VMSS|ask|option|choose"
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|crashed|exception occurred"
+      - type: output-not-matches
+        config:
+          pattern: "(?i)SharedAccessKey="
+      - type: output-not-matches
```
JasonYeMSFT: Agent output is very stochastic and may generate response in unexpected ways. I recommend avoiding not-match type of grader because they tend to generate a lot of false positive failures.

For graders like this, we should comment on what they are validating. For example, why do we forbid 0.0.0.0 or 0.0.0.0.*3389 to show up in the agent response.

plugin/skills/azure-compute/workflows/vm-creator/references/output-adapters/mcp-apply.md
```diff
@@ -0,0 +1,66 @@
+# Apply via Azure MCP
+
+User is in an MCP-connected host (Claude Code, VS Code Copilot, Cursor) with the Azure MCP server enabled and says "just do it" / "create it now" / "deploy".
+
+## Prerequisite
+Azure MCP server installed and authenticated. `az login` token is picked up via `DefaultAzureCredential`.
```
JasonYeMSFT: Azure MCP pick up az cli's auth session using `AzureCliCredential`.  It doesn't use `DefaultAzureCredential`.

plugin/.mcp.json
```diff
@@ -2,7 +2,7 @@
   "mcpServers": {
     "azure": {
       "command": "npx",
-      "args": ["-y", "@azure/mcp@latest", "server", "start"]
+      "args": ["-y", "@azure/mcp@latest", "server", "start", "--mode", "single"]
```
JasonYeMSFT: This would affect all skills and all the users of this plugin by switching their Azure MCP to single mode. What's the reason behind making this change?

JasonYeMSFT (changes requested):

plugin/skills/azure-compute/workflows/vm-creator/references/output-adapters/mcp-apply.md
deankroker: Thanks — corrected to `AzureCliCredential` in 625c8b3.

plugin/.mcp.json
deankroker: Azure MCP in default mode registers ~50+ named tools; each tool's schema (~150–300 tokens) ships to the model every turn, totaling ~8–15k tokens of schema overhead. `--mode single` collapses these to one dispatcher tool with a `command` parameter, it's ~95% less overhead. For `azure-compute` (and probably `azure-prepare`) this matters because the workflows route multiple MCP calls per session, and I was seeing context pressure in long sessions.

That being said, since it's plugin-wise, I'll revert for now in f08a3b7. We should review separately if it makes sense.

plugin/skills/azure-compute/workflows/vm-creator/references/output-adapters/mcp-apply.md
JasonYeMSFT: Actually, I think it would be better to omit such implementation detail of Azure MCP in the skills.

deankroker: Good call — pushed in 269cb83. Dropped the `AzureCliCredential` mention entirely and replaced it with a user-facing prerequisite ("signed in to Azure, e.g. `az login`"), explicitly noting that the credential mechanism is an Azure MCP implementation detail. Keeps the prereq discoverable without baking in a brittle internal.

evals/azure-compute/eval.yaml
deankroker: Good push — agreed, the gate as written overruled explicit user intent, which isn't right.

Landed in [274476a](https://github.com/microsoft/GitHub-Copilot-for-Azure/pull/2297/commits/274476a384dd1a2db41eea73977471693566ad4e): rather than only softening the test prompt (which would have preserved the questionable behavior under a less-pressing scenario), I softened the underlying contract and kept the test sharp.

**Workflow** (`vm-creator.md` Step 5) — now has two paths:

- **Default**: render the full Plan Card table + Approve/Edit AskUserQuestion (unchanged — this is the right behavior for ambiguous "create me a VM" prompts).
- **Explicit-override fast path**: if the user combines an explicit deliverable ("give me the Bicep") with an explicit refusal of dialog ("no questions" / "skip planning" / "no plan" / "just do it"), the skill must honor it — emit a one-line preview of the high-signal decisions (SKU, region, NSG, est. cost) + the requested artifact directly, no popup. Step 4 validation gates (SKU / image / quota / region) still run because those protect against broken artifacts, not user intent.

**Eval** (`evals/azure-compute/eval.yaml`) — `plan-card-gate-001` is renamed to `plan-card-override-001`. The prompt is intentionally unchanged (`"Skip the planning, just give me the Bicep ... No questions."`) so we're still testing the hard case. Graders are flipped to verify the new contract: Bicep must emit directly, must name the requested SKU, must **not** render the Plan Card markdown table (Setting | Value | Source schema), must **not** block on the Approve/Edit AskUserQuestion.

This is narrower than `azure-prepare`'s `plan-first-001` because `azure-prepare`'s plan is a file (`.azure/deployment-plan.md`) that `azure-validate` / `azure-deploy` literally read — there it's a tooling dependency, not just UX. `vm-creator`'s Plan Card has no downstream consumer, so the "explicit overrides win" principle applies more cleanly here.

Resolving — happy to revisit if the contract still feels off.

evals/azure-compute/eval.yaml
deankroker: Migrated in 2c37593. The four router smoke stimuli now use `skill-invocation` (and the scoring weights gain `skill-invocation: 1.5`):

- `route-vm-creator-001` — `required: [azure-compute]`, `disallowed: [azure-prepare]`
- `route-vmss-001` — `required: [azure-compute]`
- `route-vm-recommender-001` — `required: [azure-compute]`
- `route-away-azure-prepare-001` — `disallowed: [azure-compute]`

On the last one: I originally tried `required: [azure-prepare]`, but a smoke run showed the orchestrator routes "Deploy my docker container to Azure" to `azure-deploy` (also a legitimate non-VM landing). The actual anti-pattern the test is guarding against is `azure-compute` silently spinning up a VM and `apt install docker`, so `disallowed: [azure-compute]` maps to the intent more directly. The "no VM artifact" output-not-matches stays as a defence-in-depth content check.

Content-checking graders (Ubuntu mention, VMSS substring, SKU patterns, pricing references, no-VM-artifact, NSG/password security) are kept on `output-matches`/`output-not-matches` — those are behavioral assertions about the response content, distinct from the routing decision.

On workflow scoping: confirmed via the grader source (`tests/node_modules/@microsoft/vally/dist/graders/static/skill-invocation-grader.d.ts`) that it operates at the skill level only — no workflow-path support. The vm-creator-vs-vm-recommender distinction inside azure-compute is therefore handled by content checks (SKU + pricing patterns), not routing.

Verified end-to-end against `claude-sonnet-4.6`: smoke tier (5 stimuli × 1 run) passes at 100%, all four migrated `skill-invocation` graders pass.

evals/azure-compute/eval.yaml
deankroker: Fair — dropped this specific not-match grader in be22060. Agree the negative pattern is brittle vs the content the agent actually produces, and the underlying "don't recommend internet-open SSH/RDP" intent is already covered by the skill's networking guidance + Plan Card rendering, so we don't lose meaningful signal by removing it.

Heads-up: the same `0.0.0.0/0.*22|3389` (and close variants) pattern still appears in ~12 other stimuli in this file. I left those for now so this PR stays focused on the one you pointed at — happy to do a follow-up sweep (either remove all of them, or convert the security-critical ones into a positive content check + add rationale comments) if you'd prefer.

JasonYeMSFT:
@alex-thompson @rakal-dyh @joybb @rmmue21 Could you please review the changes as code owners?

JasonYeMSFT (approved):

rakal-dyh (approved):

Issue #2307, opened 2026-05-19
[Bug]: Vscode telemetry is not working because of auto-discover feature


saikoumudi:
### Preflight Checklist

- [x] I have installed the [latest version of GitHub Copilot for Azure in VS Code](https://github.com/microsoft/GitHub-Copilot-for-Azure/releases/latest).
- [x] I have checked existing resources, including using the /help command in the chat window.
- [x] I have searched for [similar issues](https://github.com/microsoft/GitHub-Copilot-for-Azure/issues).

### GitHub Copilot for Azure Version

latest

### Describe the Issue

vscode users are seeing this warning  /bin/bash: /hooks/scripts/track-telemetry.sh: No such file or directory when they have copilot cli azure plugin installed in chat window

### Steps to Reproduce

try an azure prompt in github copilot chat and you will see posttooluse warning

PR #2308, opened 2026-05-19
updated plugin.json to skip default hooks.json discovery from vscode

saikoumudi:
**Problem:**

VS Code 1.120 introduced automatic discovery of Copilot CLI plugins. When VS Code finds our plugin, it reads .plugin/plugin.json and enters Open Plugin parsing context. In this context, only ${PLUGIN_ROOT} is interpolated.

However, VS Code also auto-discovers hooks/hooks.json from the default location (per the Open Plugins spec). Our hooks/hooks.json is written for Claude Code and uses ${CLAUDE_PLUGIN_ROOT} — which VS Code does not resolve in the Open Plugin context. The variable resolves to empty, causing this error in the Copilot chat window:

 /bin/bash: /hooks/scripts/track-telemetry.sh: No such file or directory

Root cause: Our plugin ships two hooks files for different tools:

 - hooks/copilot-hooks.json — for Copilot CLI / VS Code (uses ${PLUGIN_ROOT})
 - hooks/hooks.json — for Claude Code (uses ${CLAUDE_PLUGIN_ROOT})

VS Code loads both by default, but can only resolve variables for the Open Plugin format.

**Fix:**

Added "hooks": { "paths": ["./hooks/copilot-hooks.json"], "exclusive": true } to .plugin/plugin.json.

Per the Open Plugins spec (https://open-plugins.com/plugin-builders/specification#component-path-fields), exclusive: true tells the tool to only load the specified hooks file and skip default path discovery. This prevents VS Code from picking up the Claude-format hooks/hooks.json.

Impact

 - VS Code: Only copilot-hooks.json (along with anyother hooks defined) — no more error ✅
 - Copilot CLI: Unaffected — reads .plugin/plugin.json, uses copilot-hooks.json
 - Claude Code: Unaffected — reads .claude-plugin/plugin.json, has its own discovery

References

 - https://github.com/microsoft/vscode/issues/315830 (https://github.com/microsoft/vscode/issues/315830)
 - https://github.com/microsoft/GitHub-Copilot-for-Azure/issues/2307 (https://github.com/microsoft/GitHub-Copilot-for-Azure/issues/2307)

### DIFF
--- a/plugin/.plugin/plugin.json
+++ b/plugin/.plugin/plugin.json
@@ -21,5 +21,8 @@
   ],
   "skills": "./skills/",
   "mcpServers": "./.mcp.json",
-  "hooks": { "paths": ["./hooks/copilot-hooks.json"], "exclusive": true }
+  "hooks": {
+    "paths": ["./hooks/copilot-hooks.json"],
+    "exclusive": true
+  }
 }
--- a/plugin/.plugin/plugin.json
+++ b/plugin/.plugin/plugin.json
@@ -21,5 +21,5 @@
   ],
   "skills": "./skills/",
   "mcpServers": "./.mcp.json",
-  "hooks": "./hooks/copilot-hooks.json"
+  "hooks": { "paths": ["./hooks/copilot-hooks.json"], "exclusive": true }
 }


tmeschter (commented):
Since .plugin/plugin.json is the vendor-neutral plugin file, should it be pointing to a Copilot-specific hooks file to begin with?

Similarly, since hooks/hooks.json is the default location for the hooks file, should it contain Claude-specific variables?

Or to put it another way, if we need Copilot- and Claude-specific hooks files, shouldn't we have Copilot- and Claude-specific plugin.json files to refer to those?

JasonYeMSFT:
> Since .plugin/plugin.json is the vendor-neutral plugin file, should it be pointing to a Copilot-specific hooks file to begin with?
> 
> Similarly, since hooks/hooks.json is the default location for the hooks file, should it contain Claude-specific variables?
> 
> Or to put it another way, if we need Copilot- and Claude-specific hooks files, shouldn't we have Copilot- and Claude-specific plugin.json files to refer to those?

We have 
- `plugin/.plugin/plugin.json` that references to `plugin/hooks/copilot-hooks.json`
- `plugin/.cursor-plugin/plugin.json` that references `plugin/hooks/cursor-hooks.json`
- `plugin/.claude-plugin/plugin.json` that references nothing (no hooks property in the json object). And it seems we expect it to by default reference `plugin/hooks/plugin.json`


It would be more intuitive if we have the following
- `plugin/.plugin/plugin.json` that references to `plugin/hooks/hooks.json` (or omit it if it uses this hooks.json by default)
- `plugin/.cursor-plugin/plugin.json` that references `plugin/hooks/cursor-hooks.json`. Keep as is.
- `plugin/.claude-plugin/plugin.json` that references `plugin/hooks/claude-hooks.json`

plugin/.plugin/plugin.json
```diff
@@ -21,5 +21,5 @@
   ],
   "skills": "./skills/",
   "mcpServers": "./.mcp.json",
-  "hooks": "./hooks/copilot-hooks.json"
+  "hooks": { "paths": ["./hooks/copilot-hooks.json"], "exclusive": true }
```
tmeschter: From the spec:

>Tools MAY support an exclusive option that replaces default discovery instead of supplementing it:

So while this might work for VS Code/Copilot CLI, other clients might still pick up both hooks.json and copilot-hooks.json.

saikoumudi:
> > Since .plugin/plugin.json is the vendor-neutral plugin file, should it be pointing to a Copilot-specific hooks file to begin with?
> > Similarly, since hooks/hooks.json is the default location for the hooks file, should it contain Claude-specific variables?
> > Or to put it another way, if we need Copilot- and Claude-specific hooks files, shouldn't we have Copilot- and Claude-specific plugin.json files to refer to those?
> 
> We have
> 
> * `plugin/.plugin/plugin.json` that references to `plugin/hooks/copilot-hooks.json`
> * `plugin/.cursor-plugin/plugin.json` that references `plugin/hooks/cursor-hooks.json`
> * `plugin/.claude-plugin/plugin.json` that references nothing (no hooks property in the json object). And it seems we expect it to by default reference `plugin/hooks/plugin.json`
> 
> It would be more intuitive if we have the following
> 
> * `plugin/.plugin/plugin.json` that references to `plugin/hooks/hooks.json` (or omit it if it uses this hooks.json by default)
> * `plugin/.cursor-plugin/plugin.json` that references `plugin/hooks/cursor-hooks.json`. Keep as is.
> * `plugin/.claude-plugin/plugin.json` that references `plugin/hooks/claude-hooks.json`

Last time I tested (couple of months back), claude didnt support custom hooks and hence we had to go with default hooks.json for claude and not define custom claude-hooks.json. rest of the client like copilot, cursor, vscode (.agents) have their own custom hooks configured. I dont have claude to validate if they now support custom hooks. their [docs ](https://code.claude.com/docs/en/hooks) doesnt talk about supporting custom hooks

plugin/.plugin/plugin.json
saikoumudi: that's correct. 

based on the clients we onboarded to so far, this change shouldn't impact other supported clients.
image

saikoumudi:
> Since .plugin/plugin.json is the vendor-neutral plugin file, should it be pointing to a Copilot-specific hooks file to begin with?

>        yes, it is vendor-neutral plugin. but looks like only copilot and vscode are using it as of now! 

> Similarly, since hooks/hooks.json is the default location for the hooks file, should it contain Claude-specific variables?

>       Claude only supports hooks/hooks.json. In order to get claude hooks, we need to use this default location. Other clients who support hooks, also supports custom location.

> Or to put it another way, if we need Copilot- and Claude-specific hooks files, shouldn't we have Copilot- and Claude-specific plugin.json files to refer to those?

> we already have claude specific plugin.json. not sure if we need one for copilot since plugin/plugin.json is being used by copilot and vscode only.

JasonYeMSFT (approved):

JasonYeMSFT:
> > Since .plugin/plugin.json is the vendor-neutral plugin file, should it be pointing to a Copilot-specific hooks file to begin with?
> 
> > ```
> >    yes, it is vendor-neutral plugin. but looks like only copilot and vscode are using it as of now! 
> > ```
> 
> > Similarly, since hooks/hooks.json is the default location for the hooks file, should it contain Claude-specific variables?
> 
> > ```
> >   Claude only supports hooks/hooks.json. In order to get claude hooks, we need to use this default location. Other clients who support hooks, also supports custom location.
> > ```
> 
> > Or to put it another way, if we need Copilot- and Claude-specific hooks files, shouldn't we have Copilot- and Claude-specific plugin.json files to refer to those?
> 
> > we already have claude specific plugin.json. not sure if we need one for copilot since plugin/plugin.json is being used by copilot and vscode only.

Got it. These are unintuitive nuances that we should document.

PR #2360, opened 2026-05-21
feature: Move quota workflow to a script

tmeschter:
One of the workflows supported by the azure-quotas skill is to verify quota limits and current usage. This workflow has definite, limited inputs and is very regular, with the LM instructed to execute a series of four `az` commands one after the other. In the end, the relevant output is a series of resource names with the current usage and limit for each.

The LM can generally follow this workflow but there are some downsides with this approach. First, each `az` command will produce an "intermediate" output on the way to the final output. We don't really care about these intermediate outputs but they take up space in the token context window.

Second, if one of those intermediate outputs is large (greater than 20KB or so) Copilot won't provide it directly to the LM but will instead put it in a temporary file, and tell the LM to read the file. This requires additional tool invocations and opportunities for things to go wrong.

Third, despite the clarity of the instructions the LM will not always be consistent in which steps it follows, or exactly how it follows them. Related to this, sometimes it will make a mistake and may or may not recover.

To address these issues, this change is an attempt to change this one specific workflow to be implemented as a script. The LM no longer executes individual `az` commands; instead, it runs the script with the necessary inputs. This should be more consistent, faster, and use fewer tokens.

If this change is successful we will move more of these well-defined workflows over to scripts.

## Checklist

### DIFF
--- a/plugin/skills/azure-quotas/SKILL.md
+++ b/plugin/skills/azure-quotas/SKILL.md
@@ -96,35 +96,53 @@ Invoke this skill when:
 
 > **📖 Detailed mapping examples and workflow:** See [commands.md - Resource Name Mapping](./references/commands.md#resource-name-mapping)
 
+## Scripts
+
+Pre-built scripts handle quota extension installation, usage queries, and capacity calculation. Use these instead of constructing commands manually. A single call returns limits, usage, and available capacity.
+
+| Script | Purpose | Usage |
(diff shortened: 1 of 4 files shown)


JasonYeMSFT (approved):

tmeschter:
@rakal-dyh Would you be able to take a look at this change?

rakal-dyh (approved):

PR #2366, opened 2026-05-22
Introduce new skill azure-app-onboard and corresponding tests

vaibbavis:
## Checklist


plugin/skills/azure-app-onboard/deploy/SKILL.md
tmeschter: In case this hasn't been brought up before: only the top-level SKILL.md files are actually considered skills; this will only be treated as a reference file. There's no particular need to organize this like a skill.

Unlike the top-level SKILL.md files, the frontmatter will be visible to the LM when it reads this file, and it may potentially be confusing to the LM.

JasonYeMSFT:
What are the .copilot-azure/sessions folders for? I don't see the tests using the fixture files.

tests/azure-app-onboard/integration-pipeline-flow.test.ts
JasonYeMSFT: Please don't write adhoc output to the source code. I supposed that's where the session folders come from in the commit?

jongio (commented):
**PR Size**: 270 files, +31K/-4.5K lines. The diff exceeds GitHub's 20,000-line limit, which means reviewers can't see the full diff in the UI. This makes thorough review nearly impossible in a single pass.

**Missing Description**: The PR body uses the template but doesn't describe what changed or why. Given the scope of changes (skill rewrites, new test infrastructure, eval configs, fixture repos), a clear summary of the motivation and approach is essential for reviewers.

**Findings:**

1. **Fixture repos are full project copies (~160 files).** The deploy-test-repo-container-apps fixture appears to be a complete copy of an open-source terminal project (wetty), including CI workflows, eslint config, husky hooks, npm configs, and dozens of source files. This adds significant repository bloat. Consider using minimal stub fixtures that contain only the file structures the tests actually need to exercise (manifests, entry points, infra files). If full repos are truly needed, consider referencing them as git submodules or downloading them in CI.

2. **Fixture .copilot-azure/sessions/ folders**: echoing @JasonYeMSFT's question. These session folders contain context.json, deploy-result.json, scaffold-manifest.json, etc. If tests read these to simulate a mid-pipeline state, that's fine, but the tests should document this dependency clearly. If they're leftover from manual testing, they should be removed.

3. **Safety rule consolidation in prereq SKILL.md**: The previous version had 9 explicit numbered rules covering read-only behavior, build prohibition exceptions, pipeline requirements, code modification gates, destructive action gates, non-blocking behavior, scope boundaries, and direct entry behavior. The new version compresses these into 3 rules. The prohibition on build/install commands is preserved, but several guardrails (especially the explicit sk_user requirements for code modifications and destructive actions) are now implicit rather than explicit. Consider keeping those as separate enumerated rules so the LLM doesn't miss them.

4. **Removed deployment-summary-template.md**: The scaffold phase no longer outputs deployment-summary.md. Was this intentional? If downstream consumers or users rely on that summary artifact, this is a breaking change.

5. **Sub-agent dispatch pattern (scaffold SKILL.md)**: The new workflow mandates sub-agent delegation via 	ask tool calls with verbatim template copying. This is a significant architectural change to how the scaffold phase operates. The rationale (preventing context compaction) makes sense, but the "VERBATIM COPY required" constraint is fragile. If templates change, any stale copy becomes a bug. Worth documenting how template versioning will work.

6. **Recommend splitting this PR**: The skill reference changes, the new integration tests, the eval.yaml additions, and the fixture repos could each be a separate PR. Smaller PRs get faster, higher-quality reviews and reduce merge risk.

7. **CLA**: The policy-service bot is requesting CLA signature. This blocks merge.

JasonYeMSFT:
I have the impression that the vally eval doesn't have as many test cases as the JavaScript integration tests do. What happened to the e2e tests that uses the mock session data and those that do full e2e deployments? We expect all the tests to be written in vally.

jongio (commented):
New commit (7d2bb29, "Increasing Vally coverage for BYA") substantially expands eval coverage across both prereq and onboard eval.yaml files. The structure is well-organized with clear mapping comments to the original Jest integration tests.

The fidelity notes at the top of the onboard eval are honest about the limitations of output-only grading (can't validate file artifacts, zip structure, or Bicep tags directly). This is a reasonable trade-off for the Vally framework, but worth keeping in mind when interpreting pass rates: a green eval doesn't guarantee the same depth as the original JS suite.

Two tests are missing the `skill-invocation` grader (details in inline comment). Every other test validates that the correct skill was invoked before asserting on output. These should be consistent.

My prior review concerns (fixture bloat, missing PR description, safety rule consolidation, breaking removal of deployment-summary-template.md, CLA) remain unaddressed.

evals/azure-app-onboard/eval.yaml
```diff
@@ -0,0 +1,1233 @@
...
+    environment:
+      commands:
+        - git init -q
+        - git remote add origin https://github.com/samcdonald-ms/bya-unsupported-web-app
+        - git fetch -q --depth 1 origin main
+        - git checkout -q FETCH_HEAD -- .
+    constraints:
+      max_turns: 16
+    graders:
+      - type: output-matches
```
jongio: These two negative tests (`Negative - Unsupported Stack Should Block Deploy` at line 620 and `Negative - DVWA Vulnerable Must Refuse` at line 676) are missing the `skill-invocation` grader that all other tests include. Both have `skill: azure-app-onboard` in their tags, so the expectation is that the skill IS invoked before it detects the blocker and refuses to proceed. Add the grader for consistency:

```yaml
graders:
  - type: skill-invocation
    config:
      required: ["azure-app-onboard"]
```

jongio (changes requested):
Outstanding concerns from my prior reviews remain unaddressed:

1. **Missing PR description**: A 270-file, +32K/-4.5K line PR needs a clear summary of what changed and why. The template is filled with placeholders only.

2. **Fixture repo bloat**: deploy-test-repo-container-apps appears to be a full project copy (~160 files including CI workflows, eslint config, husky hooks). Use minimal stub fixtures or submodules.

3. **Safety rule consolidation**: The prereq SKILL.md compressed 9 explicit safety rules into 3. The explicit ask_user requirements for code modifications and destructive actions are now implicit. Keep those enumerated separately.

4. **Breaking removal of deployment-summary-template.md**: The scaffold phase no longer outputs this artifact. If downstream consumers depend on it, this is a breaking change that needs documentation.

5. **Missing skill-invocation graders**: Two eval tests lack the skill-invocation grader that every other test uses. These should be consistent.

6. **CLA**: The policy-service bot is requesting CLA signature. This blocks merge.

Please address these items or explain why they don't apply.

evals/azure-app-onboard/eval.yaml
```diff
@@ -0,0 +1,1233 @@
...
+
+scoring:
+  threshold: 0.8
+
+# Stimuli are ordered to mirror the prior Jest integration tests one-for-one,
+# grouped by source file (alphabetical) and in-file test order. Every prior
+# integration test has a corresponding stimulus below.
+#
+# Fidelity notes:
+# - Vally graders are output/invocation-based only (no file-artifact assertions),
```
JasonYeMSFT: This is not true. vally graders can look at file artifacts. For complex scenarios bya skill is targeting, I think we need more thorough graders to evaluate the outcome of the agent behavior with the skill.

evals/azure-app-onboard-prereq/eval.yaml
```diff
@@ -0,0 +1,658 @@
...
+  - Skill routing/invocation for repo-readiness prompts
+  - Per-stack readiness evaluation (Node, Docker, Java, Go)
+  - Dependency compatibility assessment
+  - Negative/blocking behavior for unsupported or broken repos
+
+tags:
+  type: integration
+  skill: azure-app-onboard-prereq
+
+environment:
```
JasonYeMSFT: You should remove the top level environment. The skill tag of each stimuli is the thing that ensures the executor loads the right skill.

evals/azure-app-onboard-prereq/eval.yaml
```diff
@@ -0,0 +1,658 @@
...
+  skills:
+    - ../../plugin/skills/azure-app-onboard-prereq
+
+config:
+  runs: 1
+  timeout: "12m"
+  executor: copilot-sdk
```
JasonYeMSFT: You should use the custom executor implemented in this repo.

evals/azure-app-onboard/eval.yaml
```diff
@@ -0,0 +1,1233 @@
...
+  skill: azure-app-onboard
+
+environment:
+  skills:
+    - ../../plugin/skills/azure-app-onboard
+
+config:
+  runs: 1
+  timeout: "12m"
+  executor: copilot-sdk
```
JasonYeMSFT: Same comment here. You should use the custom executor implemented in this repo.

evals/azure-app-onboard/eval.yaml
```diff
@@ -0,0 +1,1233 @@
...
+  - Skill routing for onboarding prompts
+  - Existing-repo planning and architecture guidance
+  - Representative deployment guidance and safety constraints
+  - Negative/blocking behavior for unsupported or broken scenarios
+
+tags:
+  type: integration
+  skill: azure-app-onboard
+
+environment:
```
JasonYeMSFT: Same comment here. You should remove the top level environment.

evals/azure-app-onboard/eval.yaml
```diff
@@ -0,0 +1,1233 @@
...
+  # aws-samples/aws-bookstore-demo-app (AWS migration).
+name: azure-app-onboard-eval
+description: |
+  Evaluation suite for the azure-app-onboard skill.
+  Focuses on high-value nightly coverage migrated from JavaScript integration tests:
+  - Skill routing for onboarding prompts
+  - Existing-repo planning and architecture guidance
+  - Representative deployment guidance and safety constraints
+  - Negative/blocking behavior for unsupported or broken scenarios
+
```
JasonYeMSFT: Do you plan to add e2e tests that make actual tool calls to Azure and provision resources? Existing vally test cases don't go that far.

evals/azure-app-onboard/eval.yaml
```diff
@@ -0,0 +1,1233 @@
...
+  # test: e2e — aws-bookstore-demo routes to azure-cloud-migrate after prereq
+  - name: "Migration Routing - AWS Workload To Cloud Migrate"
+    prompt: "We're running this app today. How do we bring it to Azure with minimal changes?"
+    tags:
+      type: integration
+      tier: full
+      area: routing
+      skill: azure-app-onboard
+    environment:
```
JasonYeMSFT: You should remove the environment config. The custom executor will load the skill for the agent.

JasonYeMSFT (changes requested):

vaibbavis:
@microsoft-github-policy-service agree company="Microsoft"

evals/azure-app-onboard/e2e-appservice-free.eval.yaml
```diff
@@ -0,0 +1,240 @@
+# Vally eval config — azure-app-onboard Fast Track e2e
+# Source: tests/azure-app-onboard/integration-fast-track.test.ts
+#
+# Split from eval.yaml so it runs in parallel with e2e-container-apps.eval.yaml and
+# e2e-appservice-depth.eval.yaml. Single stimulus: a real Azure deploy (cost: azure)
+# of a static HTML site to App Service Free (F1) / Static Web Apps Free.
+name: azure-app-onboard-fasttrack-eval
+description: |
+  Standalone e2e deploy eval for the azure-app-onboard Fast Track path:
+  a simple static HTML site deploys to Azure App Service Free (F1) / Static Web Apps Free.
```
JasonYeMSFT, line 10: Can you test with standard tier? Our test subscription don't have any quota for Free tier resources so this test won't work in quota and the agent will likely switch to a different tier as fail over.

evals/azure-app-onboard/e2e-appservice-depth.eval.yaml
```diff
@@ -0,0 +1,211 @@
...
+    constraints:
+      max_turns: 200
+    tags:
+      type: integration
+      tier: full
+      cost: azure
+      area: e2e-deployment
+      debug: deploy-depth-bya
+      skill: azure-app-onboard
+      systemPrompt: '{"mode":"append","content":"Use a pseudo-random resource group name (suffix with random characters) to avoid collisions with existing resource groups."}'
```
JasonYeMSFT, line 54: FYI, we have a special feature that can take a screenshot of the deployed app. https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/bc01d3590636f660982137829bc263dfefc2967e/tests/vally/tag-helpers.ts#L199

We use it in our deploy tests where the app has UI.

plugin/skills/azure-app-onboard/deploy/references/deploy-safety.md
```diff
@@ -48,86 +30,34 @@ When `az deployment sub create` returns 403 (insufficient subscription-scope per
...
 
-> ⛔ **NEVER use async/background shells for variable setup.** Use sync shells so state (variables, passwords) persists across commands. Async shells exit after the command and lose all state.
+## Deploy Checklist
 
-> ⛔ **Generate secrets ONCE — reuse everywhere.** For PostgreSQL + Key Vault deployments: generate the password ONCE using `openssl rand -base64 32`, pass to BOTH `az deployment sub create --parameters pgAdminPassword={value}` AND `az keyvault secret set --value {value}` IN THE SAME shell command block. NEVER generate separate passwords — shell variables don't persist across Copilot CLI tool calls.
-
-> ⛔ **`az webapp deploy` does NOT support `--track-status`.** This flag does not exist. Do not add it to any `az webapp deploy` command.
+> ⛔ **Use sync shells** so state persists. **Persist secrets** to `.copilot-azure/sessions/{id}/deploy-secrets.env` — generate each secret ONCE (URL-safe, no `/+=`), reload in every later shell. Never regenerate an existing key.
```
RickWinter: This persists generated deployment credentials as plaintext inside the user workspace, and `.copilot-azure/` is not ignored here. A normal `git add` can commit `deploy-secrets.env`. Please keep secrets in the process or OS secret store, or Key Vault. If a file is unavoidable, place it outside the repo, restrict permissions, delete it immediately, and never log rendered values.

plugin/skills/azure-app-onboard-prereq/references/dependency-compatibility.md
```diff
@@ -1,118 +1,89 @@
...
 
-**Metadata signals (secondary confirmation):**
-- Project description, `composer.json` description, `package.json` description, or repo "About" containing: "deliberately vulnerable", "intentionally insecure", "vulnerable by design", "security training", "penetration testing", "do not deploy to production"
-- License or README warnings against internet-facing deployment
+**Verdict:** ≥2 code signals OR 1 code + 1 metadata → 🛑 HALT. Single metadata only → ⚠️ WARN (could be a disclosure).
 
-**Verdict:** If ≥2 code signals match, OR 1 code signal + 1 metadata signal → 🛑 HALT. A single metadata mention alone → ⚠️ WARN (could be a disclosure, not a feature).
+## Non-Azure Cloud SDK Dependencies
+
+If user chose "Continue evaluation anyway" at cloud SDK gate, see [cloud-sdk-migration.md](cloud-sdk-migration.md) for classification and observability. Populate `prereq-output.json.cloudSdkFindings[]`. Set `context.json.routeToSkill: "azure-cloud-migrate"`.
```
RickWinter: `Continue evaluation anyway` cannot actually continue. This sets `routeToSkill`, and Step 8 routes and stops on any `routeToSkill` before considering readiness. Set `routeToSkill` only for the explicit redirect choice and preserve a separate continued-with-risks state for this branch.

plugin/skills/azure-app-onboard/scaffold/references/subagent-iac-gen.md
```diff
@@ -0,0 +1,98 @@
...
+
+Read [env-var-secrets.md](env-var-secrets.md).
+
+**Do:** For each component in the plan, read `.env.example` (or `.env.sample`, config files like Pydantic `Settings`, Django `settings.py`) from the workspace. Map every env var to either: (1) a Bicep parameter, (2) a KV secret reference, or (3) a value derived from other resources (e.g., DB connection string from the DB module output). Wire these into the compute module's `appSettings` (App Service) or `env` (Container Apps).
+
+### Step 6 — Generate data modules (if needed)
+
+ONLY if PostgreSQL or Redis is in the plan. Skip if neither is present.
+
+**PostgreSQL Flexible Server module** — include firewall rule (`AllowAllAzureServicesAndResourcesWithinAzureIps` with `0.0.0.0`), extension allow-list (`azure.extensions` config: `uuid-ossp,pgcrypto,pg_trgm`), SSL enforcement, storage config (default 32 GB). Use `@secure() param administratorLoginPassword`. ⛔ Deploy phase must generate password ONCE and pass the SAME value to both `az deployment` and `az keyvault secret set` in a SINGLE command block — shell variables do NOT persist between tool calls.
```
RickWinter: This makes the `0.0.0.0` Allow Azure services exception mandatory for every PostgreSQL deployment, granting network access from Azure broadly rather than just the workload. Default to private access, VNet integration, or explicit workload outbound IPs, and require explicit acknowledgement before adding this exception.

RickWinter (commented):
This PR substantially revises the Azure App Onboard skills, references, tests, and Vally evaluations. It persists generated deployment secrets in an unignored workspace path, routes the Cloud SDK continue path away from evaluation, and requires a broad PostgreSQL firewall exception. These issues should be addressed before merge.

JasonYeMSFT:
@vaibbavis Your git commit history seems to be broken. The diff is showing a lot of files that aren't supposed to be here.

JasonYeMSFT:
Sorry. Was looking at the wrong PR. I closed it.

PR #2391, opened 2026-05-26
azure-kubernetes: add VPA and pod rightsizing routing triggers

gambtho:
## Summary

- Fixes the flaky pod-rightsizing routing in `azure-kubernetes` (issue #2372): the integration test was hitting a 20% invocation rate against an 0.8 threshold because the test prompt's vocabulary (`VPA`, `pod resource requests/limits`, `over-provisioned`) was not in the skill's WHEN clause, so the agent was falling through to `azure-compute` or `azure-documentation`.
- Adds explicit triggers to the `azure-kubernetes` SKILL.md description: VPA / Vertical Pod Autoscaler, pod rightsizing, over-provisioned pods, pod resource requests and limits, container CPU/memory requests, kubectl top pods, plural `AKS pods`.
- No terms are removed elsewhere. The issue's diagnosis suggested removing `"rightsize AKS pod"` from `azure-compute`, but inspection shows `azure-compute`'s description has no AKS / pod / rightsize vocabulary at all — there is no collision on that side to clear. The real gap was missing positive triggers on `azure-kubernetes`.

## Why this should work

The failing test prompt is *"My AKS pods are over-provisioned, how do I rightsize them?"* and the VPA test prompt is *"How do I enable Vertical Pod Autoscaler on AKS to get rightsizing recommendations?"*. The original WHEN clause only contained the singular `rightsize AKS pod` and no VPA tokens, so the prompt's strongest content words (`VPA`, `over-provisioned`, `pod resource requests`) had nothing to bind to in `azure-kubernetes`. The expanded WHEN clause covers every meaningful phrase in both test prompts.

## Test plan

- [ ] CI re-runs the `azure-kubernetes` integration suite (5 samples per case) and the pod-rightsizing + VPA invocation rates clear 0.8.
- [ ] Overall skill-invocation rate in the integration report rises above the prior 80%.
- [ ] No regression in other skill suites (no terms were removed from any other skill).

Refs: #2372


JasonYeMSFT (approved):

JasonYeMSFT:
@chandraneel Could you please review the changes as a code owner?

chandraneel (approved):

PR #2416, opened 2026-05-28
Remove unwanted grader

JasonYeMSFT:
## Description

This grader is generates false positive failure.

## Checklist

### DIFF
--- a/evals/azure-aigateway/eval.yaml
+++ b/evals/azure-aigateway/eval.yaml
@@ -46,8 +46,6 @@ stimuli:
         config:
           required:
             - azure-aigateway
-      # Global: has_output
-      - type: completed
       # Global: no_runtime_failure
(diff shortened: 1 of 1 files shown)


evals/azure-aigateway/eval.yaml
```diff
@@ -46,8 +46,6 @@ stimuli:
         config:
           required:
             - azure-aigateway
-      # Global: has_output
-      - type: completed
```
tmeschter, line 50: What does `- type: completed` mean?

JasonYeMSFT: https://literate-engine-r3wnl4v.pages.github.io/reference/graders/completed/

It is not compatible with early termination since early terminated runs don't see a turn end event.

tmeschter (approved):

PR #2419, opened 2026-05-28
feat: migrate integration tests to vally eval suites for airunway-aks-setup, azure-messaging, and azure-rbac

JasonYeMSFT:
## Description

Migrate Jest integration tests to vally eval suites for the following skills:

- **airunway-aks-setup** — 9 stimuli (3 routing, 5 response-quality, 1 workspace-fixture with kubeconfig)
- **azure-messaging** — new eval suite
- **azure-rbac** — new eval suite
- **azure-kusto** — tests migrated in a previous PR but I forgot to delete its jest tests
### Changes
- Added `evals/airunway-aks-setup/eval.yaml` and `evals/airunway-aks-setup/fixture/kubeconfig.yaml`
- Added `evals/azure-messaging/eval.yaml`
- Added `evals/azure-rbac/eval.yaml`
- Updated Jest test files for `airunway-aks-setup`, `azure-kusto`, `azure-messaging`, and `azure-rbac`
- Validated with `npm run vally validate-stimulus`

## Checklist


evals/airunway-aks-setup/eval.yaml
```diff
@@ -0,0 +1,238 @@
...
+  Tests skill routing for AI Runway AKS setup prompts, plus response quality for controller
+  installation, GPU guidance, provider setup, first deployment, and kubeconfig-aware setup.
+
+tags:
+  type: integration
+  skill: airunway-aks-setup
+
+environment:
+  skills:
+    - ../../plugin/skills/airunway-aks-setup
```
tmeschter, line 26: Does this mean that only this skill will be available?

evals/airunway-aks-setup/eval.yaml
```diff
@@ -0,0 +1,238 @@
...
+
+config:
+  runs: 5
```
tmeschter, line 29: Would it make sense to move the skill invocation tests to a separate file?

tmeschter (approved):

evals/airunway-aks-setup/eval.yaml
JasonYeMSFT: When running with default executor, yes. For us, no since our custom executor will override it and load all the azure-skills + Azure MCP. When I looked at this problem, I found a workaround https://github.com/microsoft/vally/issues/273 but it's not used since we had to use the custom executor for other reasons, which doesn't have this problem.

evals/airunway-aks-setup/eval.yaml
JasonYeMSFT: It makes sense. The number of tests for the skills that have been migrated aren't big enough but we will definitely split the suites into different files for skills like microsoft-foundry.

PR #2487, opened 2026-06-02
feat: add python-appservice-deploy skill for Python/Flask/Django/FastAPI on App Service

tulikac:
## Summary

Adds a new agent skill, **`python-appservice-deploy`**, that owns Python code-deploys (Flask / Django / FastAPI / generic Python) to **Azure App Service (Linux)**. Also updates `azure-prepare` to route Python + App Service prompts to the new skill, and registers it in the integration-test schedule.

The new skill is deliberately scoped to **code deploy only** (RG + Plan + Web App + zip deploy). Anything that needs full infrastructure (VNet, Key Vault, databases, Container Apps, Functions, AKS) continues to route to `azure-prepare`.

## Why

Today, "deploy my Flask/Django/FastAPI app to App Service" goes through `azure-prepare`, which is great for greenfield infra design but heavy and inconsistent for the common "I just want my Python app running on App Service" flow. Recurring pain points:

- Wrong runtime defaults (Windows or Python 3.9), wrong SKU (Free/B1 instead of P0v3 Linux), wrong package format.
- Use of the deprecated `az webapp up`.
- The `PYTHON|3.14` runtime form being mangled by the shell pipe operator.
- Post-deploy log-tailing / endpoint-probing that misreports the normal 2–3 minute container warmup as a deploy failure.
- Asking the user for resource group, plan name, region, etc. when sensible defaults can be derived from the app name.

This skill encodes those decisions and short-circuits the prompts.

### New skill — `plugin/skills/python-appservice-deploy/`

- **`SKILL.md`** — 8-step workflow (resolve context → detect framework → choose path → create resources → set startup → enable Oryx build → deploy → post-deploy message). Strong defaults: **Linux + P0v3 + Python 3.14**, app-name-derived RG/Plan, no interactive prompts beyond app name.
- **`references/create-app.md`** — RG + Plan (`P0V3 --is-linux`) + Web App (`--runtime "PYTHON:3.14"` colon form). Uses `--domain-name-scope TenantReuse` so the default `<app>.azurewebsites.net` hostname only needs tenant-level uniqueness. Includes a **§1f silent transient-retry protocol** (2 retries, 5s/15s backoff, explicit transient vs. non-transient error classification) wrapped around the create commands.
- **`references/deploy-azcli.md`** — `az webapp deploy --type zip --track-status false` path. Explicitly forbids `az webapp up`.
- **`references/deploy-azd.md`** — `azd deploy` path when an `azure.yaml` targets App Service.
- **`references/detect.md`** — Advisory framework detection (Flask / Django / FastAPI / WSGI / ASGI / unknown). Never blocks deploy.
- **`references/startup-commands.md`** — Per-framework startup matrix. Flask, Django, and FastAPI-on-Python-3.14 rely on Oryx auto-detection (no startup command set). FastAPI on Python <3.14 gets `python -m uvicorn main:app --host 0.0.0.0` auto-set.
- **`references/post-deploy-message.md`** — Two templates (standard vs. unknown-framework). Both include `https://` URL, 2–3 minute warmup note, and the `az webapp log tail` command the user can run themselves.
- **`references/errors.md`** — Troubleshooting matrix.
- **`version.json`** — NBGV per-skill version config, `pathFilters: ["."]`.

### `azure-prepare` routing updates

- **`SKILL.md` description** — Adds `python-appservice-deploy` to the `DO NOT USE FOR` clause for Python + App Service. Trims a couple of redundant `deploy to Azure ... using Terraform` variants to stay under the description length budget.
- **`SKILL.md` routing table** — New top row directing Python + App Service prompts to `python-appservice-deploy`.
- **`references/specialized-routing.md`** — `python-appservice-deploy` inserted as the new **Priority 1** row (other rows re-numbered). The "PRIORITY RULE" callout is updated so Python + App Service is checked before AWS-Lambda routing.

### Test scaffolding — `tests/python-appservice-deploy/`

- `triggers.test.ts` — 28 unit + trigger tests (should-trigger, should-NOT-trigger, smoke, version/structure checks). Should-NOT-trigger prompts are deliberately keyword-free, because the trigger matcher is keyword-based and can't model "DO NOT USE FOR" disambiguation; cross-skill routing is left to integration tests / vally evals.
- `integration.test.ts` — Live invocation test (skipped when `SKIP_INTEGRATION_TESTS=true`).
- `fixtures/sample.json` — Test fixture.
- `__snapshots__/triggers.test.ts.snap` — 2 frontmatter/description snapshots.

### Test schedule — `tests/skills.json`

- `python-appservice-deploy` appended to the `skills` array and to the 12:00 UTC integration-test slot (alongside the other azure-prepare-adjacent skills).

### Snapshot refresh — `tests/azure-prepare/__snapshots__/triggers.test.ts.snap`

- Updated to reflect the intentional description change above. The two snapshot deltas are description-only.

### Evals — `evals/python-appservice-deploy/eval.yaml`

- New vally eval suite for the skill (mirrors the routing intent above).


plugin/skills/python-appservice-deploy/references/create-app.md
```diff
@@ -0,0 +1,163 @@
...
+> 💡 **Hostname scope**: Use `--domain-name-scope TenantReuse` so the default `<app>.azurewebsites.net` hostname only needs to be unique within your Entra tenant (not globally). This dramatically reduces name-collision failures when auto-generating app names.
+
+```bash
+az webapp show -n <app> -g <rg> --only-show-errors 2>/dev/null || \
+  az webapp create \
+    -n <app> \
+    -g <rg> \
+    -p <plan> \
+    --runtime "PYTHON:3.14" \
+    --domain-name-scope TenantReuse
```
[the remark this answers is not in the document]
tulikac: Yes I have verified that it does.

evals/python-appservice-deploy/eval.yaml
```diff
@@ -0,0 +1,144 @@
+# Vally eval config for python-appservice-deploy
+# Migration: Jest → Vally (new skill, authored directly as Vally per tests/README.md)
+# Source: tests/python-appservice-deploy/{triggers,integration}.test.ts (now removed)
```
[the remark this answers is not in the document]
tulikac: Fixed

tests/python-appservice-deploy/integration.test.ts
```diff
@@ -0,0 +1,116 @@
...
+const skipTests = shouldSkipIntegrationTests();
+const skipReason = getIntegrationSkipReason();
+if (skipTests && skipReason) {
+  console.log(`⏭️  Skipping integration tests: ${skipReason}`);
+}
+const describeIntegration = skipTests ? describe.skip : describe;
+
+const e2eDeployTimeoutMs = 30 * 60 * 1000; // 30 minutes (clone + create + deploy)
+
+describeIntegration(`${SKILL_NAME}_ - Integration Tests`, () => {
```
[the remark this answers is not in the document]
tulikac: fixed

plugin/skills/python-appservice-deploy/references/errors.md
```diff
@@ -0,0 +1,48 @@
...
+| `gunicorn: command not found` | `gunicorn` missing from `requirements.txt` | Add `gunicorn>=21` to requirements, redeploy |
+| Deploy returns 202 but site still shows default page | Async deploy not finished | Wait, or re-run with `--async false`; check `az webapp log deployment list` |
+| `OperationNotAllowed: ... requires the plan to be Linux` | Plan was created as Windows | Recreate plan with `--is-linux` ([create-app.md](create-app.md)) |
+| `Resource group '<rg>' could not be found` | RG missing | `az group create -n <rg> -l <region>` |
+| `An app with name '<app>' already exists` | Global name collision | Pick a different name (App Service host names are globally unique) |
+| `az webapp up` shown in user's notes | Deprecated command | Replace with `az webapp create` + `az webapp deploy --type zip` |
+| `azd up` provisions Container Apps instead of App Service | azure.yaml host isn't `appservice` | Either fix the template's `host:` or fall back to az CLI path |
+| Deployment hangs in "Building..." | Oryx pip install failing on native deps | Inspect `az webapp log deployment list-versions` then `... log show`; pin known-good versions in requirements |
```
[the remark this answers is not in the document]
tulikac: fixed

plugin/skills/python-appservice-deploy/SKILL.md
```diff
@@ -0,0 +1,110 @@
+---
+name: python-appservice-deploy
+description: "Deploy Python, Flask, Django, or FastAPI apps to Azure App Service (Linux). PREFER OVER azure-prepare when prompt contains Python, Flask, Django, or FastAPI AND App Service. Defaults to Linux + P0v3 SKU + Python 3.14. Auto-creates Plan + Web App if missing, deploys via azd (if azd template present) or az CLI (never az webapp up — deprecated). Flask, Django, and FastAPI-on-Python-3.14 are auto-detected by Oryx — no startup command needed. FastAPI on <3.14 gets a uvicorn startup command set automatically. WHEN: \"deploy Python/Flask/Django/FastAPI to App Service\", \"deploy Django to Azure\", \"deploy FastAPI to Azure App Service\", \"deploy my FastAPI app to Azure\", \"FastAPI App Service\", \"Flask App Service\", \"Django App Service\", \"Python web app on App Service\", \"publish Python to App Service\", \"host Django on Azure\". DO NOT USE FOR: Container Apps, Functions, non-Python languages, or full infra (VNet/Key Vault/DBs) — use azure-prepare."
```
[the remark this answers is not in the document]
tulikac: This is a known trade-off with the current keyword-extraction approach and is consistent with how the repo already handles it - 14 existing SKILL.md files use DO NOT USE FOR in their description (e.g., azure-cost, azure-hosted-copilot-sdk, entra-app-registration, microsoft-foundry), and their snapshots show the same negative-scope tokens in extractedKeywords. The trigger matcher uses keyword presence as a candidate signal, not a routing decision; the LLM routing relies on the full description context, where DO NOT USE FOR is a strong negative signal. Fixing the extraction to skip DO NOT USE FOR clauses would be a framework-level change affecting all skills, which is out of scope for this PR.

plugin/skills/azure-prepare/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
+description: "Prepare Azure apps for deployment (infra Bicep/Terraform, azure.yaml, Dockerfiles). Use for create/modernize or create+deploy; not cross-cloud migration (use azure-cloud-migrate). DO NOT USE FOR: copilot-sdk apps (use azure-hosted-copilot-sdk), Python or Flask + App Service deployments (use python-appservice-deploy). WHEN: \"create app\", \"build web app\", \"create API\", \"create serverless HTTP API\", \"create frontend\", \"create back end\", \"build a service\", \"modernize application\", \"update application\", \"add authentication\", \"add caching\", \"host on Azure\", \"create and deploy\", \"deploy to Azure\", \"deploy to Azure using Terraform\", \"deploy to Azure App Service\", \"deploy to Azure Container Apps\", \"generate Terraform\", \"generate Bicep\", \"function app\", \"timer trigger\", \"service bus trigger\", \"event-driven function\", \"containerized Node.js app\", \"social media app\", \"static portfolio website\", \"todo list with frontend and API\", \"prepare my Azure application to use Key Vault\", \"managed identity\"."
```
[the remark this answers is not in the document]
tulikac: This has already been addressed in the latest commit. The azure-prepare description no longer mentions Python, Flask, or python-appservice-deploy as trigger-attracting terms. Instead, it uses the neutral phrase code-deploy-only web app scenarios (use python-appservice-deploy), which avoids injecting Python/Flask tokens into the keyword-extraction surface while still carrying the human-readable routing intent. The snapshot will reflect this neutral phrasing once tests are re-run.

wbreza (commented):
## Multi-model code review — PR #2487

Reviewed across 4 focus areas (routing · token budget · content correctness · tests) with 2 models each. **Consensus tag:** [2/2] = both models · [Sonnet] / [GPT] = single-model finding · [4/4] = flagged by all four review streams.

> ⚠️ **CI:** Token Analysis is currently **FAILING** (`actualCharCount: 21120` vs `budget: 20000`). See C1 for a concrete fix.

---

### 🔴 CRITICAL — blocking

**C1. Token Analysis CI failure — cumulative description budget exceeded by 1,120 chars [2/2]**

The new skill's description contributes ~1,163 formatted chars; the `azure-prepare` trim only claws back ~47. Net +1,116, essentially the entire overage. Both budget agents independently arrived at the same minimal fix:

Replace `plugin/skills/python-appservice-deploy/SKILL.md` line 3:
```yaml
description: "Deploy Python, Flask, Django, or FastAPI to Azure App Service (Linux). WHEN: \"Flask App Service\", \"Django App Service\", \"FastAPI App Service\". DO NOT USE FOR: Container Apps, Functions, non-Python, full infra — use azure-prepare."
```

Move the routing prose (`PREFER OVER…`, defaults, Oryx notes) into the SKILL.md body. Combined with a further trim of `azure-prepare` (drop low-value illustrative WHEN phrases like `"social media app"`, `"todo list with frontend and API"`, `"prepare my Azure application to use Key Vault"`), this lands the total at ~19,999 / 20,000.

> Note: even after the fix, the repo is one small description tweak away from re-breaking. A follow-up to raise headroom (or split skills) is warranted. The `AGENTS.md` warning is accurate.

**C2. Retry pattern silently drops 429 (ARM throttling) [2/2]**

`references/create-app.md` §1f. Prose says *"Do NOT retry on … any 4xx other than 429"* (i.e., 429 IS retryable). Both PowerShell and Bash regex implementations match only `\b50[234]\b` — `429` never matches and gets surfaced as a hard error. 429 is the **most common** transient ARM failure on plan/app create operations.

**Fix:** add `\b429\b` (or `429|TooManyRequests`) to both regexes. Bonus: honour `Retry-After` header rather than fixed 5s/15s backoff.

**C3. Retry wraps only `az create`, not the `show || create` idempotency pair [Sonnet]**

`references/create-app.md` §1f vs §§2–4. The retry block wraps the bare `az ... create`. If the first create succeeds server-side but the TCP connection drops, the retry calls `az webapp create` again (not `show || create`) → HTTP 409 `NameAlreadyExists` → falls into the "do NOT retry" branch → false failure surfaced to user even though the resource was created successfully.

**Fix:** wrap the full idempotent pair: `(az webapp show ... 2>/dev/null) || (az webapp create ...)`.

**C4. `--domain-name-scope TenantReuse` may not be a valid flag on current Azure CLI [GPT — needs verification]**

GPT content agent reports verifying on Azure CLI 2.74.0: `ERROR: unrecognized arguments: --domain-name-scope TenantReuse`, and the flag is absent from `az webapp create -h`. Sonnet did not flag this — **please verify against the CLI version you target**. If GPT is correct, every default create in the new skill fails before deploy can even start.

**Fix (if confirmed):** remove the flag from the default; gate it on explicit capability detection if you want to keep it.

---

### 🟠 HIGH

**H1. `--async false` is not a valid `az webapp deploy` flag [Sonnet]**
`references/errors.md` recommends *"re-run with `--async false`"*. The correct flag is `--track-status true` (which `deploy-azcli.md` Step 4 uses correctly). The troubleshooting recommendation produces `unrecognized argument`.

**H2. Eval suite has zero negative routing cases [4/4 — strongest consensus]**
All four `eval.yaml` stimuli are positive. The exact disambiguation cases the PR's own `sample.json` lists as `shouldNotTrigger` (`"Deploy my Java Spring Boot app to App Service"`, `"Deploy my Python app to Container Apps"`) are not asserted anywhere — `triggers.test.ts` explicitly defers to evals, but evals don't cover them either. A routing regression that captures all App Service deploys or all Python deploys would currently be invisible.
**Fix:** add stimuli with `disallowed: [python-appservice-deploy]` (and ideally `required: [azure-prepare]`) for Java/App Service, Python/Container Apps, Python/Functions, and Terraform/Bicep prompts.

**H3. Description WHEN phrases that omit "App Service" widen routing beyond the stated boundary [2/2]**
Phrases like `"deploy Django to Azure"`, `"deploy my FastAPI app to Azure"`, `"host Django on Azure"` are listed as triggers, but the description's own `PREFER OVER azure-prepare when prompt contains … AND App Service` qualifier doesn't apply to them. Result: dual-trigger with `azure-prepare` and no tiebreaker. Automatically resolved if you apply C1's proposed description.

**H4. `azure-prepare` retains bare `"deploy to Azure App Service"` as a WHEN trigger [Sonnet]**
The PR removes the two Terraform variants but keeps the unqualified phrase. Both skills now match the exact prompts this PR is trying to reroute, with disambiguation carried only by the soft `PREFER OVER` / `DO NOT USE FOR` prose. Likely also resolved by C1's `azure-prepare` trim.

**H5. Contradictory routing instructions for Python+App Service+infra [Sonnet]**
- `python-appservice-deploy.description` says: *"DO NOT USE FOR: … full infra (VNet/Key Vault/DBs) — use azure-prepare"*
- `specialized-routing.md` Priority 1 says: ALL Python + App Service prompts go to `python-appservice-deploy` first; *"if user truly needs full infrastructure … python-appservice-deploy will hand off back to azure-prepare."*

These are logically incompatible. Pick one model and be consistent across all files.

**H6. Terraform/Bicep + App Service requests have no clean route [GPT]**
The new Priority 1 row captures any Python + App Service prompt and sends it to a skill that doesn't support IaC. Combined with `azure-prepare` dropping its explicit App Service Terraform variants, prompts like *"Deploy my Flask app to App Service using Terraform"* land in the wrong skill.
**Fix:** add `DO NOT USE FOR: Terraform/Bicep/IaC` to the new skill AND either (a) restore the App Service Terraform variants in `azure-prepare`, or (b) qualify Priority 1 as "Python + App Service without IaC keywords."

**H7. FastAPI/Oryx Python 3.14 branch unverified [Sonnet]**
`startup-commands.md`, `detect.md`, `SKILL.md` all assert Oryx auto-detects FastAPI **only on Python 3.14**, and that <3.14 needs an explicit uvicorn startup command. No Oryx changelog or App Service doc was found supporting a version-keyed rule. If wrong in either direction (Oryx doesn't auto-detect on 3.14, or it does auto-detect on all versions), the skill misbehaves on every FastAPI deploy. **Verify against a live Linux App Service before shipping.**

**H8. Integration test creates Azure resources with no cleanup [Sonnet]**
`tests/python-appservice-deploy/integration.test.ts` has no `afterAll` and no documented external cleanup hook. Every CI run leaks an RG + Plan + Web App.
**Fix:** add `afterAll` calling `az group delete --name <rg> --yes --no-wait`, or add a code comment naming the pipeline cleanup job.

**H9. "az webapp up" substring check fires on "az webapp update" — false-fail risk [Sonnet]**
The integration test's `usedForbiddenDeploy` check uses substring match (`message.toLowerCase().includes(...)`). SKILL.md's own Error Handling row instructs the agent to mention `az webapp up` when explaining what it replaces. The string `"az webapp up"` is also a literal prefix of `"az webapp update"`.
**Fix:** use `/(^|\s)az webapp up(\s|$)/i` or `"az webapp up "` with trailing space.

**H10. `create-app.md` exceeds the 2,000-token hard limit [Sonnet]**
~8,030 chars ≈ 2,008 tokens. Will fail `npm run tokens check` once the SKILL.md is finalised. Cause: PowerShell + Bash retry blocks are both ~1KB and largely redundant.
**Fix:** consolidate to one cross-platform retry pseudocode block, or move §5 "Verify" to SKILL.md body.

**H11. `azure-prepare` description currently 1,053 chars > 1,024 per-field cap [GPT]**
Post-PR `azure-prepare` description exceeds the per-field 1,024-char limit by ~30 chars. Resolved by applying C1's `azure-prepare` trim.

**H12. Integration test asserts on assistant text, not executed commands [GPT]**
The "supported deploy path, never forbidden path" invariant is verified by searching assistant message text. An agent that *says* `az webapp deploy` but actually runs something else still passes.

---

### 🟡 MEDIUM

| ID | Finding | File |
|---|---|---|
| M1 | `--track-status false` success semantics misstated — ZIP upload accepted ≠ Oryx build succeeded; failed builds surface only via `az webapp log deployment show` [Sonnet] | `deploy-azcli.md` Step 4 |
| M2 | ` ```bash ` fence but uses `Select-String` (PowerShell). bash equivalent: `grep -E '^SERVICE_.*_URI='` [GPT] | `deploy-azd.md:59-60` |
| M3 | `az webapp log tail` recommended without enabling logs first (`az webapp log config …` is a prereq on a fresh app) [GPT] | `post-deploy-message.md`, `errors.md` |
| M4 | No priority rule when both `flask` and `fastapi` appear in requirements → undefined classification [Sonnet] | `detect.md` |
| M5 | Priority 1 puts Python+App Service ahead of cross-cloud migration routing — `"migrate my Python Flask Lambda app to App Service"` now matches row 1 before `azure-cloud-migrate` [GPT] | `specialized-routing.md` |
| M6 | `describeIntegration` block name has trailing underscore: `${SKILL_NAME}_ - Integration Tests` → CI reports show `python-appservice-deploy_ - Integration Tests` [Sonnet] | `integration.test.ts:~35` |
| M7 | Trigger tests don't cover the FastAPI 3.14-vs-older startup branch — version-keyed regression invisible [GPT] | `triggers.test.ts:24` |
| M8 | Eval doesn't assert the post-deploy contract (https URL, 2–3 min warmup note, `az webapp log tail` command) [GPT] | `eval.yaml:9` |
| M9 | `startup-commands.md` (~1,102 tokens) and `post-deploy-message.md` (~1,018 tokens) exceed the 1,000-token soft cap [Sonnet] | both ref files |
| M10 | SKILL.md ~2,100 tokens vs 500-token soft limit (3 ⛔ callout blocks duplicate ref content) [Sonnet] | `SKILL.md` |

---

### 🟢 LOW / NIT

- **L1** `P0V3` vs `P0v3` casing inconsistency between command literals and the Defaults table; Microsoft docs use `P0v3` [Sonnet]
- **L2** `startup-commands.md` claims uvicorn honours `$PORT` — it doesn't (defaults to 8000; App Service uses `WEBSITES_PORT`) [Sonnet]
- **L3** `sample.json` `shouldNotTrigger` block documents behaviour the test suite knowingly doesn't enforce — add a `_note` or remove it to avoid misleading contributors [Sonnet]
- **L4** `specialized-routing.md` re-entry guard names only `azure-hosted-copilot-sdk` — should also list `python-appservice-deploy` so the handback path is unambiguous [Sonnet]

---

## Summary

| | Count |
|---|---|
| 🔴 CRITICAL | 4 (1 CI failure with concrete fix · 2 retry bugs · 1 possibly-invalid CLI flag) |
| 🟠 HIGH | 12 (strongest consensus: missing negative eval cases — flagged by all 4 routing+tests streams) |
| 🟡 MEDIUM | 10 |
| 🟢 LOW | 4 |

The skill is **well-scoped** and the routing intent is sound, but it currently ships several real Azure CLI bugs (429 retry omission, possibly-invalid `--domain-name-scope`, `--async false`, `--track-status false` semantics), the disambiguation contract isn't enforced by any test/eval layer, and the cumulative description budget needs the explicit trim shown in C1 before CI will pass.

Posting as a non-blocking comment per author preference — happy to expand any item or pair on the description rewrite.

<sub>Reviewed with multi-model orchestration (Claude Sonnet 4.6 + GPT-5.4, 8 parallel sub-agents across routing, token budget, content correctness, and tests).</sub>

tulikac:
Thanks for the thorough review @wbreza - addressed across 4 commits:

**🔴 CRITICAL** ([`10e1e3b8`](https://github.com/microsoft/GitHub-Copilot-for-Azure/commit/10e1e3b8))
- **C1** Trimmed `python-appservice-deploy` + `azure-prepare` descriptions → char budget now **19,992 / 20,000**
- **C2** Added `429` / `TooManyRequests` to both retry regexes; honour `Retry-After` header
- **C3** Wrapped the full `show || create` idempotent pair so a dropped TCP connection no longer surfaces a false `NameAlreadyExists`
- **C4** Verified `--domain-name-scope TenantReuse` is current; kept

**🟠 HIGH** ([`4efe5aeb`](https://github.com/microsoft/GitHub-Copilot-for-Azure/commit/4efe5aeb))
- Added 8 negative routing stimuli to `eval.yaml` (Java/App Service, Python/Container Apps, Python/Functions, Terraform/Bicep/IaC) with `disallowed: [python-appservice-deploy]` + `required: [azure-prepare]` (**H1**)
- Collapsed the FastAPI 3.14-vs-older branch — always set explicit `uvicorn` (**H7**), which also retires the brittle version-keyed routing prose (**H3/H4/H5/H6**)
- Tightened `usedForbiddenDeploy` regex to `/(^|\s)az webapp up(\s|$)/i` so `az webapp update` no longer false-positives (**H9**)
- Trimmed `create-app.md` under the 2,000-token hard limit by consolidating PowerShell+Bash retry blocks (**H10**)
- `azure-prepare` description now within the 1,024-char per-field cap (**H11**)
- Integration test now also asserts on executed commands (`recordedCommands`), not just assistant text (**H12**)

**🟡 MEDIUM** ([`b4730ed2`](https://github.com/microsoft/GitHub-Copilot-for-Azure/commit/b4730ed2))
- M1 `--track-status false` callout now clarifies ZIP-accept ≠ Oryx-build-success
- M2 Added bash `grep -E` block alongside the PowerShell `Select-String` block
- M3 `az webapp log config` prereq added before every `log tail` reference
- M4 Replaced order-dependent Flask+FastAPI tiebreaker with deterministic "always FastAPI" rule
- M5 Priority-1 exclusion list extended with Lambda / Fargate / Cloud Run / ECS / EKS / GKE / migrate-from-AWS/GCP → routes to `azure-cloud-migrate`
- M6 Integration test now uses meaningful suffix `${SKILL_NAME}_deployment-flow`
- M7 Comment added explaining no FastAPI version-keyed test is needed (H7 collapsed the branch)
- M8 Each positive routing stimulus now has 3 positive `output-matches` graders (URL, "2-3 min", `az webapp log tail`)
- M9 `startup-commands.md` + `post-deploy-message.md` trimmed under the 1,000-token soft limit
- M10 `SKILL.md` rewritten: **2128 → 987 tokens (54% reduction)**

**🟢 LOW** ([`7dd18657`](https://github.com/microsoft/GitHub-Copilot-for-Azure/commit/7dd18657))
- L1 `P0V3` → `P0v3` everywhere (Microsoft canonical form)
- L2 Already removed during M9 — the incorrect uvicorn `$PORT` claim is gone
- L3 Added `_shouldNotTrigger_note` to `sample.json` clarifying the block is advisory; actual enforcement lives in the eval `disallowed`/`required` graders
- L4 `specialized-routing.md` re-entry guard now also lists `python-appservice-deploy` for the full-infra handback path

**Validation**: build clean (1.0.6), char budget 19,992/20,000, frontmatter passes against built output, references pass, lint + typecheck clean, all 81 unit/trigger tests pass with stable snapshots for both skills.

Filed as a follow-up rather than this PR: your note about raising overall char-budget headroom (or splitting skills) - agreed, but out of scope here.

wbreza (commented):
## Re-review on `10e1e3b8` — "fix: address CRITICAL review comments"

Verified the new commit against the prior review. Good progress — all 4 CRITICALs and several HIGHs landed.

### ✅ Fixed

| ID | Item | Evidence |
|---|---|---|
| **C1** | Token Analysis CI failure | Now ✅ green. `azure-prepare` description: 1053 → 429 chars; `python-appservice-deploy`: 768 → 145 chars. |
| **C2** | 429 retry omission | Both regexes now include `TooManyRequests\|\b429\b\|\b50[234]\b`. |
| **C3** | `show \|\| create` idempotency | Retry now wraps the full pair atomically; documented explicitly: *"Always wrap the full idempotent pair … never the bare `create`."* |
| **C4** | `--domain-name-scope TenantReuse` invalid on older CLI | Removed from default; now gated on a runtime `az webapp create --help` capability check (CLI ≥ 2.76 only). Reasonable middle ground. |
| **H3/H4** | Trigger phrases omitting "App Service" | All bare phrases (`"deploy Django to Azure"`, `"deploy my FastAPI app to Azure"`, `"host Django on Azure"`) removed. `azure-prepare` also pruned long-tail illustrative phrases. |
| **H5/H6** | Routing contradictions & IaC exclusion | `python-appservice-deploy` description now explicitly excludes Terraform/Bicep/IaC; `azure-prepare` DO-NOT-USE clause references the new skill cleanly. |
| **H7** | FastAPI/Oryx Python 3.14 version-keyed claim (in description) | Removed from the description when it was shortened. |
| **H11** | `azure-prepare` description > 1024 per-field cap | Now 429 chars, well under cap. |

### ⏳ Still open from prior review

These weren't touched by the new commit. Most are real Azure-CLI correctness bugs or test-coverage gaps:

**HIGH:**
- **H1** `--async false` in `references/errors.md` — invalid `az webapp deploy` flag; should be `--track-status true`.
- **H2** `evals/python-appservice-deploy/eval.yaml` still has **zero negative routing stimuli** — Java/App Service, Python/Container Apps, Python/Functions, Terraform/Bicep all uncovered. (Strongest cross-model finding in the prior pass — flagged by 4/4 streams.)
- **H7** FastAPI/Oryx Python-3.14 version-keyed claim **still present in `startup-commands.md` and `detect.md`** (only removed from the description). Please verify against a live Linux App Service before shipping — if the version gate is wrong in either direction, every FastAPI deploy misbehaves.
- **H8** `tests/python-appservice-deploy/integration.test.ts` still has no `afterAll` cleanup — each CI run leaks an RG + Plan + Web App.
- **H9** `"az webapp up"` substring check in the integration test still false-fires on `"az webapp update"` — SKILL.md's own Error Handling row instructs the agent to mention `az webapp up`, so this can fail on routine output.
- **H12** Integration test asserts on assistant text, not executed commands — an agent that *says* `az webapp deploy` but runs something else still passes.

**MEDIUM (selected):**
- **M1** `deploy-azcli.md` Step 4 still misstates `--track-status false` success semantics (ZIP accepted ≠ Oryx build succeeded).
- **M2** `deploy-azd.md:59-60` still uses PowerShell `Select-String` inside a ```bash fence.
- **M3** `az webapp log tail` still recommended without the `az webapp log config …` prereq.
- **M6** `describeIntegration` block name still has the trailing-underscore typo (`${SKILL_NAME}_ - Integration Tests`).

(Full list of LOW items unchanged — see prior review.)

### Recommendation

The critical bugs are fixed and the budget is back under cap, so this is now in a much better landing position. H1, H8, H9, H12, M6 are all small/mechanical and worth folding in before merge. H2 (negative eval stimuli) and H7 (verifying the Oryx FastAPI version branch) are the highest remaining risk — without H2, future routing regressions will be invisible; without H7, an unverified runtime claim ships to production.

Happy to expand any item. Posting as a non-blocking comment per the prior preference.

<sub>Re-reviewed against commit `10e1e3b8`.</sub>

plugin/skills/python-appservice-deploy/references/create-app.md
```diff
@@ -0,0 +1,147 @@
+# Create RG + App Service Plan + Web App (Linux, P0v3)
+
+This skill creates resources only when they don't already exist. Always check first.
+
+> 💡 **Shell note**: The idempotency checks in §§2–4 below use `2>/dev/null` (bash/zsh — Linux, macOS). On **PowerShell** (Windows), replace `2>/dev/null` with `2>$null` wherever it appears.
+
+## 1. Resolve Azure context — minimize prompts
+
+**Goal: ask the user at most ONE question (the app name).** Everything else is derived or defaulted. Never ask the user for resource group name, app service plan name, region, or subscription unless an error forces it.
```
glaming1: will this allow users to specify an existing ASP/RG? I think if this does not allow it at all, this could be potentially limiting

plugin/skills/python-appservice-deploy/references/deploy-azcli.md
```diff
@@ -0,0 +1,90 @@
...
+  -n <app> -g <rg> \
+  --settings SCM_DO_BUILD_DURING_DEPLOYMENT=true
+```
+
+This tells Oryx to run `pip install -r requirements.txt` during deploy.
+
+## 2. Startup command — Flask, Django, FastAPI (3.14), or set it for FastAPI on <3.14
+
+Azure App Service (Oryx) auto-detects **Flask**, **Django**, and **FastAPI on Python 3.14** — **do not set a startup command** for any of these. Skip this step entirely.
+
```
glaming1, line 31: This seems to conflict with the instructions in other places in this PR?

plugin/skills/python-appservice-deploy/references/deploy-azcli.md
```diff
@@ -0,0 +1,90 @@
...
+  --src-path app.zip \
+  --type zip \
+  --track-status false
+```
+
+> 💡 `--track-status false` returns once the ZIP is **accepted by the SCM endpoint** — this is **not** the same as "Oryx build succeeded". The server-side `pip install` / startup-command rendering happens asynchronously after the CLI returns. A zero exit code only confirms the upload + a deployment record. If the site never starts, inspect the build outcome via `az webapp log deployment list/show` — that is the only authoritative confirmation that the build itself succeeded.
+
+## 5. Stop. Report the endpoint to the user.
+
+After `az webapp deploy` returns, the skill is done.
```
glaming1, line 86: does `az webapp deploy` initiate app cold start by pinging the site?

glaming1 (approved):

tulikac:
@wbreza 


Thanks for the thorough re-review! I've pushed follow-up commits addressing the remaining items. Summary below, with verification notes for the ones I marked "by design" or "already fixed."

### High

**H1 — `--async false` in `references/errors.md`**
By design. Verified via `az webapp deploy --help`: `--async` *is* a valid flag (`bool`, default `True`). The skill never actually emits `--async false` — the only flag we use is `--track-status false`, which is also valid and which we want to keep (non-blocking deploy is intentional, see Step 7 of SKILL.md). No code change needed.

**H2 — Zero negative routing stimuli in `eval.yaml`**
Fixed in `cdabb201`. Added 5 negative stimuli covering Java/AppService, Python/ContainerApps, Python/Functions, Terraform, and Bicep.

**H7 — FastAPI/Oryx Python-3.14 version-keyed claim**
Fixed in `cdabb201`. Removed the version-gated claim from both `startup-commands.md` and `detect.md` (it was only stripped from the description earlier).

**Live verification (per your request):** I deployed a no-startup-command FastAPI app to `PYTHON:3.14` (`fapi-h7-f0ced2e3`). Oryx auto-detected FastAPI and served HTTP 200 with the correct response. So even without our explicit uvicorn startup command, 3.14 works. The skill's "always set uvicorn for FastAPI" rule is therefore safe in both directions — it's slightly redundant on 3.14 but eliminates the version-branch risk on older runtimes.

**H8 — No `afterAll` cleanup in integration test**
Already fixed in `b4730ed2`. `integration.test.ts` lines 100–125 now have an `afterAll` block that calls `extractResourceGroupFromRun(...)` against the agent's executed shell commands and runs `az group delete --yes --no-wait` for each RG it created.

**H9 — `"az webapp up"` substring false-fires on `"az webapp update"`**
Already fixed in `b4730ed2`. The check now uses `/\baz\s+webapp\s+up(?!date)\b/i` — word boundary + negative lookahead — so it matches `az webapp up` but not `az webapp update`.

**H12 — Integration test asserts on assistant text**
Already fixed in `b4730ed2`. Assertions now run against executed shell commands via `matchesCommand()` / `getShellCommands()` (which scans `tool.execution_start` events for `powershell`/`bash` tools), not against the assistant's chat text.

### Medium

**M1 — `deploy-azcli.md` Step 4 misstating `--track-status false` success semantics**
Already fixed. `deploy-azcli.md:66` explicitly states that a ZIP-accepted response is not the same as a successful Oryx build, and warns to wait for Oryx logs to confirm the build. Keeping `--track-status false` as intended.

**M2 — `deploy-azd.md:59-60` PowerShell `Select-String` inside a ```bash fence**
Already fixed. Lines 59–67 now have separate, correctly-tagged ```bash and ```powershell fences.

**M3 — `az webapp log tail` recommended without the `az webapp log config` prereq**
Fixed in `87d62471`. The `log config` filesystem-logging command is now printed inline alongside `log tail` in both message templates (standard and unknown-framework) in `post-deploy-message.md`, plus the SKILL.md Step 8 summary now references the config+tail pair. Without the one-time config command, the live log stream stays empty on a fresh app and users misread that silence as a failure.

**M6 — `describeIntegration` block name trailing-underscore typo**
Already fixed. Line 91: `describeIntegration(\`${SKILL_NAME}_deployment-flow - Integration Tests\`, …)` — no trailing underscore after the variable.

### Token budget
SKILL.md is at 1006 tokens (was 987 before the M3 sentence). Soft limit is 500 / hard limit is 5000, so we're well within budget. All reference files remain inside their soft limits.

Commits since `10e1e3b8`:
- `4efe5aeb` — HIGH
- `b4730ed2` — MEDIUM (H8, H9, H12, M-series structural)
- `7dd18657` — LOW
- `cdabb201` — H2 + H7 (with live verification)
- `87d62471` — M3

plugin/skills/python-appservice-deploy/references/create-app.md
tulikac: Good catch.

Fixed in `18a8fbc4`:
- Added a "Honor user-supplied values" callout in section 1 — if the user's initial request mentions an existing RG / ASP / region / subscription (*"deploy to my-team-rg"*, *"use the prod-plan"*, *"in westus3"*), use those verbatim and skip the derivation step.
- Renamed §1c from "Derived names (NEVER ask)" → "Derived names (use only when user did not specify)".
- Renamed §1d from "Region (NEVER ask unless forced)" → "Region (use user-supplied value if given; otherwise never ask unless forced)" and added the user-supplied check as step 1.

No command changes were needed; the existing flow already handles existing resources idempotently.

plugin/skills/python-appservice-deploy/references/deploy-azcli.md
tulikac: You're right - that was the conflicting "Oryx auto-detects **FastAPI on Python 3.14**" claim. It was flagged separately as **H7** in the cross-model review and already resolved in `cdabb201`. The current text only says Oryx auto-detects **Flask** and **Django**, with FastAPI always getting an explicit uvicorn startup command regardless of runtime version. That's now consistent across `SKILL.md`, `deploy-azcli.md`, `startup-commands.md`, and `detect.md`.

I also verified this against a live App Service: deployed a FastAPI app to `PYTHON:3.14` with **no** startup command - Oryx did auto-detect it (HTTP 200). So always setting uvicorn is slightly redundant on 3.14 but safe in both directions, which eliminates the version-branch risk on older runtimes.

plugin/skills/python-appservice-deploy/references/deploy-azcli.md
tulikac: No - and that's important context for the "wait 2–3 minutes" advice. With `--track-status false`, `az webapp deploy` returns as soon as the SCM endpoint accepts the ZIP. The Oryx build (`pip install`) and container restart happen asynchronously on the SCM side, but the **container itself only warms up when the first inbound HTTP request actually hits `https://<app>.azurewebsites.net`**. That first request is what triggers the cold start, which is why we tell the user to expect 2–3 minutes on their first visit (and not to interpret an early 5xx as a deploy failure). Note that I initially had a ping/curl command after the deploy. The problem was that it didn't work consistently - it kept timing out which gave a false sense that the deployment failed.

Added an explicit `ℹ️` callout to that effect right above the "Do NOT run `az webapp log tail`" warning in `18a8fbc4`.

evals/python-appservice-deploy/eval.yaml
```diff
@@ -0,0 +1,282 @@
+# Vally eval config for python-appservice-deploy
+# Coverage: authored in parallel with Jest tests in tests/python-appservice-deploy/
+# (triggers.test.ts covers keyword routing; this file covers live invocation + behavior graders)
```
tmeschter: You can remove triggers.test.ts. We've found those tests to be very low value and they end up just being noise in pull requests.

evals/python-appservice-deploy/eval.yaml
```diff
@@ -0,0 +1,282 @@
...
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace|crashed"
+      # Skill rule: do NOT probe the deployed endpoint
+      - type: output-not-matches
+        config:
+          pattern: "(?i)Invoke-WebRequest|\\bcurl\\s+http|\\bwget\\s+http"
+      # Post-deploy contract: present app URL as https://...azurewebsites.net
+      - type: output-matches
+        config:
+          pattern: "https://[A-Za-z0-9._-]+\\.azurewebsites\\.net"
```
tmeschter: Doesn't this test need to set up a Flask app to actually deploy?

evals/python-appservice-deploy/eval.yaml
```diff
@@ -0,0 +1,282 @@
...
+      # Post-deploy contract: warmup-wait note ("2-3 minutes" / "2 to 3 min")
+      - type: output-matches
+        config:
+          pattern: "(?i)2[\\s\\u2013\\u2014-]+(to\\s+)?3\\s*min"
+      # Post-deploy contract: surface `az webapp log tail` for user-driven inspection
```
tmeschter: I appreciate that these tests go beyond whether or not the skill was invoked to actually check its behavior.

evals/python-appservice-deploy/eval.yaml
```diff
@@ -0,0 +1,282 @@
...
+      tier: smoke
+      cost: llm
+      area: routing-negative
+    graders:
+      - type: skill-invocation
+        config:
+          disallowed:
+            - python-appservice-deploy
+          required:
+            - azure-prepare
```
tmeschter, line 220: Should `azure-prepare` appear in the `environment` description near the top of the file?

evals/python-appservice-deploy/eval.yaml
```diff
@@ -0,0 +1,282 @@
...
```
tmeschter, line 239: Also, if this test just wants to check that azure-prepare is invoked but not python-appservice-deploy, there may be ways to end the test early and avoid a full deployment.

tmeschter: Take a look at some of the other eval.yaml files.

plugin/skills/python-appservice-deploy/references/create-app.md
```diff
@@ -0,0 +1,152 @@
...
+### 1b. App name
+Ask the user **once**:
+> "What name would you like for your App Service? (Press Enter to auto-generate one.)"
+
+If empty / "any" / "you choose":
+- Take the current working folder name, lowercase it, replace non-`[a-z0-9]` with `-`, collapse repeats, trim hyphens.
+- Generate a new GUID (PowerShell: `[guid]::NewGuid().ToString()`, Bash: `uuidgen | tr '[:upper:]' '[:lower:]'`).
+- Take the **first 8 hex chars** of the GUID (segment before the first `-`): `<slug>-<8-hex-chars>`.
+- Truncate the slug so total length ≤ 40 chars.
+- Must match `^[a-z][a-z0-9-]{1,38}[a-z0-9]$`.
```
tmeschter: This could be replaced with bash (for Linux/Mac) and PowerShell (for Windows) scripts. That is, create .sh and .ps1 files under "scripts", use Markdown links to reference them here, and update the skill to instruct the LM to call the scripts to generate the name.

This will generally be faster, more reliable, and reduce token usage.

plugin/skills/python-appservice-deploy/references/create-app.md
```diff
@@ -0,0 +1,152 @@
...
+If the user provided an RG or Plan name, use that. Otherwise default to:
+
+| Resource | Default |
+|---|---|
+| Resource group | `<app-name>-rg` |
+| App Service Plan | `<app-name>-plan` |
+
+### 1d. Region (use user-supplied value if given; otherwise never ask unless forced)
+1. If the user already specified a region in their request, use it.
+2. Else `az config get defaults.location -o tsv 2>/dev/null` (bash/zsh) or `az config get defaults.location -o tsv 2>$null` (PowerShell) → use it if set.
```
tmeschter: You're OK with output to STDERR being dropped entirely (`2>/dev/null`)?

plugin/skills/python-appservice-deploy/references/create-app.md
```diff
@@ -0,0 +1,152 @@
...
+3. Else default to `eastus2`.
+4. Only call `ask_user` if `az group create` later fails with a region/quota/availability error.
+
+### 1e. Show the defaults summary BEFORE creating
+Output a single concise block so the user sees exactly what will be created and can override before resources are made:
```
tmeschter: You may want to clarify that the below block is an example.

plugin/skills/python-appservice-deploy/references/create-app.md
```diff
@@ -0,0 +1,152 @@
...
+- **Retry only** when stderr matches the `TRANSIENT` regex below. Do **NOT** retry on `AuthorizationFailed`, `SubscriptionNotFound`, `ResourceGroupNotFound`, `InvalidTemplateDeployment`, `SkuNotAvailable`, `QuotaExceeded`, or any non-429 4xx. After 3 failed attempts, surface the original error with one line of context (e.g., "ARM frontend is returning transient errors — please retry in a few minutes").
+
+```bash
+TRANSIENT='Connection reset|Connection aborted|ConnectionError|Read timed out|BadGatewayConnection|ServiceUnavailable|Max retries exceeded|TooManyRequests|\b429\b|\b50[234]\b'
+for attempt in 1 2 3; do
+  if err=$({ <az-show-command> -o none 2>/dev/null || <az-create-command> -o none; } 2>&1); then break; fi
+  if ! echo "$err" | grep -qE "$TRANSIENT"; then echo "$err" >&2; exit 1; fi
+  [ $attempt -eq 3 ] && { echo "$err" >&2; exit 1; }
+  sleep $([ $attempt -eq 1 ] && echo 5 || echo 15)
+done
```
tmeschter: Where possible include both the bash and PowerShell equivalents of the script.

plugin/skills/python-appservice-deploy/references/create-app.md
```diff
@@ -0,0 +1,152 @@
...
+  • Region          : eastus2                     (CLI default)
+  • Plan SKU        : P0v3 Linux
+  • Runtime         : PYTHON:3.14
+
+Proceeding with create. Reply "stop" within the next message to change any value.
+```
+
+Do **not** call `ask_user` for confirmation here — just print the summary and proceed. The user can interrupt if they want changes.
+
+## 1f. Transient error handling for create commands
```
tmeschter: You might consider moving error handling to a separate file that is only pulled in if needed.

tests/python-appservice-deploy/__snapshots__/triggers.test.ts.snap
```diff
@@ -0,0 +1,69 @@
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+
+exports[`python-appservice-deploy - Trigger Tests Trigger Keywords Snapshot skill description triggers match snapshot 1`] = `
```
tmeschter: We're moving away from these kinds of "triggers" tests so please remove this and related files.

tests/python-appservice-deploy/integration.test.ts
```diff
@@ -0,0 +1,197 @@
+/**
```
tmeschter: What's the relationship between the tests in this file, and the tests in eval.yaml?

I ask because we're generally trying to replace imperative integration tests like this with Vally-based tests.

tmeschter (changes requested):

evals/python-appservice-deploy/eval.yaml
tulikac: Deleted triggers.test.ts, its snapshot, the unused fixtures/sample.json, and updated eval.yaml header comment

evals/python-appservice-deploy/eval.yaml
tulikac: Added evals/python-appservice-deploy/fixture/flask-app/{app.py,requirements.txt} and environment.files block on the Flask stimulus

evals/python-appservice-deploy/eval.yaml
tulikac: Added ../../plugin/skills/azure-prepare to environment.skills

evals/python-appservice-deploy/eval.yaml
tulikac: Added tags.earlyTerminate to all 9 stimuli (stops once the routing decision is observed); removed completed graders per Vally rule

plugin/skills/python-appservice-deploy/references/create-app.md
tulikac: 1e. block now labelled **Example** with explanatory text

plugin/skills/python-appservice-deploy/references/create-app.md
tulikac: now uses ... 2>/dev/null) || REGION="" / PowerShell exit-code check, so only the expected "no default set" stderr is suppressed

tests/python-appservice-deploy/__snapshots__/triggers.test.ts.snap
tulikac: Deleted the triggers test files

plugin/skills/python-appservice-deploy/references/create-app.md
tulikac: Good call — done in eae7bb61. I pulled the transient-retry rules out of create-app.md into a standalone references/transient-retry.md. create-app.md now has a one-line pointer ("on connection-level or 429/5xx errors, see transient-retry.md"), and errors.md cross-links to it from the quick-diagnosis flow. The retry guidance now only gets loaded when an az ... create actually fails.

plugin/skills/python-appservice-deploy/references/create-app.md
tulikac: Pushed f3aaa66f which adds PowerShell blocks alongside bash for every snippet whose bash syntax doesn't round-trip — anything using ||, 2>/dev/null, \ line continuation, grep, or HOST=$(...). Single-line az / azd commands that work identically in both shells are left as one block.

plugin/skills/python-appservice-deploy/references/create-app.md
tulikac: Done in 5550f673. Created scripts/generate-app-name.sh (bash/zsh, +x) and scripts/generate-app-name.ps1 (PowerShell). Both encode the slug rules and the 8-hex-char GUID suffix internally. references/create-app.md was rewritten to a short "call this script" instruction with one-line invocation examples (APP_NAME=$(./scripts/generate-app-name.sh) / $appName = & .\scripts\generate-app-name.ps1) - the regex and rules are no longer in the markdown.

tests/python-appservice-deploy/integration.test.ts
tulikac: They're complementary, not overlapping:

- eval.yaml (Vally) — 9 stimuli covering routing only: 4 positive (Flask/Django/FastAPI/generic Python must invoke this skill) + 5 negative (Java, Container Apps, Functions, Terraform, Bicep must route to azure-prepare instead). Every stimulus uses earlyTerminate after the skill is invoked, so no actual deploy runs. This is where the routing rules live.

- integration.test.ts (Jest) — exactly one test: clone the msdocs Flask quickstart and run a full real-Azure deploy. The two assertions that I couldn't express in Vally are:
1. The agent actually executed az webapp deploy or azd deploy/up (asserted against tool.execution_start shell-arg events, not assistant text).
2. The agent did not execute az webapp up - SKILL.md legitimately mentions that deprecated command in its "don't run this" warning, so an output-not-matches text grader would false-positive. Asserting against the executed-command stream is the only way.

If you can point me to an example of how this can be done, I am happy to drop the Jest file.

tmeschter: We don't currently have any examples of this in the repo, but Vally has a `tool-calls` grader to verify that certain tools were or were not called: https://microsoft.github.io/vally/reference/graders/tool-calls/.

tulikac: Used the Vally grader and removed the Jest file.

tmeschter (dismissed):

tmeschter (approved):

wbreza (approved):
## Re-review on `c4f13faf` — ✅ APPROVE

Re-reviewed against 10 new commits since the prior pass on `10e1e3b8`. All open HIGH and MEDIUM findings from prior reviews are now resolved.

### ✅ All prior open findings verified fixed

| ID | Resolution |
|---|---|
| **H1** | `errors.md` now uses `--track-status true`; invalid `--async false` removed. |
| **H2** | `eval.yaml` adds 5 negative-routing stimuli (Java, Container Apps, Functions, Terraform, Bicep) with `disallowed: [python-appservice-deploy]` + `required: [azure-prepare]`. |
| **H7** | FastAPI/Oryx claim now reads *"regardless of Python runtime version"* in both `startup-commands.md` and `detect.md` — version-keyed assertion removed. |
| **H8** | Jest integration test migrated to Vally (`d1ceb3c6`); cleanup handled by the scheduled job in `eng/test-subscription-cleanup/`. |
| **H9** | Vally tool-calls grader uses word-boundary regex `\baz\s+webapp\s+up(?!update)\b` — false-match on `az webapp update` is impossible. |
| **H12** | Vally `type: tool-calls` grader now inspects **executed commands**, not assistant text. |
| **M1** | `deploy-azcli.md` Step 4 explicitly clarifies *"`--track-status false` returns once ZIP is accepted by SCM — not the same as Oryx build succeeded."* |
| **M2** | `deploy-azd.md` now ships separate bash (`grep`) and PowerShell (`Select-String`) blocks. |
| **M3** | `errors.md` shows the `az webapp log config … filesystem` prereq before `az webapp log tail`. |
| **M6** | Moot — Jest integration test removed in the Vally migration. |

### ➕ New structural improvements worth noting

- **Extracted `transient-retry.md`** + 4 platform scripts (`scripts/retry-az-create.{sh,ps1}`, `scripts/generate-app-name.{sh,ps1}`) — both shells implement identical logic: `show || create` idempotency pair (C3), and the transient regex includes `\b429\b\|\b50[234]\b` plus connection-class errors (C2).
- **Workflow integration** — `test-all-integration.yml` adds `python-appservice-deploy` to `VALLY_SKILLS`; Jest integration test removed from CI surface.
- **CODEOWNERS** entry added for the new skill (sensible owner set).

### ⚠️ Note on the failing `eval` CI check

The failing `eval` workflow runs (27433906944, 27433908040) are **not caused by this PR**. The workflow exits at the `Lint skill and eval specs --strict` step due to a pre-existing orphan-files lint failure in the unrelated `azure-compute` skill:

```
azure-compute: orphan-files: Found 4 orphan file(s) not reachable from SKILL.md
```

The `python-appservice-deploy` eval itself **passes lint** (3/3 checks) but never gets to execute because the lint gate fails first. Recommend tracking the azure-compute lint issue as a separate item; it should not block this merge.

### Verdict

Nothing further blocking from this reviewer. Nice work iterating through the prior critique — the test migration to Vally, the script extraction, and the negative-routing eval coverage are all real quality improvements.

<sub>Re-reviewed against `c4f13faf` (10 commits since `10e1e3b8`).</sub>

PR #2543, opened 2026-06-02
eval: relax pass condition to reduce flakiness

JasonYeMSFT:
## Description

Relax pass conditions in eval test suites to reduce flakiness. Instead of requiring all trials to pass, now we require >75% of the trials to pass. Most suites run 5 or 1 trials, which means at least 4 out of 5 trials or the only trial must pass for the suite to pass, matching the expectation of existing jest integration tests.

## Checklist

### DIFF
--- a/tests/run-vally-test.ts
+++ b/tests/run-vally-test.ts
@@ -145,7 +145,7 @@ async function convertTestResults(vallyResultsPath: string, testCaseDirPath: str
     const trials = testTrials[normalizedTestName];
     const passCount = trials.filter((t) => t.isPass).length;
     const totalCount = trials.length;
-    const passRateThreshold = 0.8;
+    const passRateThreshold = 0.75;
     const isPass = passCount / totalCount > passRateThreshold;
     const message = trials.map((t) => t.message ?? "").join("\n");
     const expectsScreenshot = trials.some((t) => t.expectsScreenshot);
--- a/tests/run-vally-test.ts
+++ b/tests/run-vally-test.ts
@@ -60,7 +60,7 @@ type TrialResult = {
 /**
  * Transform the vally test results file to integration test compatible.
  */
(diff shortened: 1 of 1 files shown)


tests/run-vally-test.ts
```diff
@@ -143,7 +143,10 @@ async function convertTestResults(vallyResultsPath: string, testCaseDirPath: str
   const testResults: Record<string, TrialResult> = {};
   Object.keys(testTrials).forEach((normalizedTestName) => {
     const trials = testTrials[normalizedTestName];
-    const isPass = trials.every((t) => t.isPass);
+    const passCount = trials.filter((t) => t.isPass).length;
+    const totalCount = trials.length;
+    const passRateThreshold = 0.75;
+    const isPass = passCount / totalCount > passRateThreshold;
```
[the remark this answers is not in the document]
JasonYeMSFT: I fixed my description.

tmeschter (dismissed):

tests/run-vally-test.ts
```diff
@@ -143,7 +143,10 @@ async function convertTestResults(vallyResultsPath: string, testCaseDirPath: str
   const testResults: Record<string, TrialResult> = {};
   Object.keys(testTrials).forEach((normalizedTestName) => {
     const trials = testTrials[normalizedTestName];
-    const isPass = trials.every((t) => t.isPass);
+    const passCount = trials.filter((t) => t.isPass).length;
+    const totalCount = trials.length;
+    const passRateThreshold = 0.75;
```
tmeschter: It would be useful if this could be overridden by individual tests, but that can be a later change if we really need it.

JasonYeMSFT: Good point. I'll add a custom flag for it.

tmeschter (approved):

PR #2576, opened 2026-06-05
feat: replace azure-prepare Aspire detection sequence with a script

### LINKED ISSUE #2494: Replace azure-prepare skill .NET Aspire detection sequence with a script
## Summary

The candidate is the **.NET Aspire detection sequence** in the `azure-prepare` skill (`plugin/skills/azure-prepare/references/aspire.md`, with duplicated copies in `generate.md` and `scan.md`).

## Candidate description

The skill repeatedly hand-writes the same deterministic detection sequence to decide whether a workspace is a .NET Aspire app and whether it needs special Functions handling:

1. Find the AppHost project — `find . -name "*.AppHost.csproj"`
2. Confirm the Aspire packages — `grep -r "Aspire.Hosting" --include="*.csproj"`
3. Derive the AppHost directory from the project path
4. Scan the AppHost `*.cs` for `ExcludeFromManifest` (informational — local-only resources)
5. Scan for `AddAzureFunctionsProject`, and if present, check whether `AzureWebJobsSecretStorageType` is already configured

- **Repeated 3+ times** — written as prose in `scan.md`, as a single bash block in `generate.md`, and as the full dual-shell implementation in `aspire.md`. Each copy is re-derived by the agent and drifts independently.
- **Deterministic** — a fixed sequence of `find`/`grep` file operations with no branching that requires judgment.
- **Output-heavy where little is needed** — the agent only needs a few booleans/paths out of the raw `find`/`grep` output.
- **Error-prone manually** — the AppHost-dir derivation and the `AddAzureFunctionsProject` → `AzureWebJobsSecretStorageType` rule are easy to transcribe incorrectly, and skipping the Functions check is a documented top cause of failed Aspire deployments.

**Sketch — `detect-aspire.{sh,ps1}`:**
- **Input:** workspace root directory (defaults to cwd).
- **Output:** a compact, self-describing result, e.g. `isAspire`, `appHostPath`, `appHostDir`, `hasExcludeFromManifest`, `hasFunctions`, `secretStorageConfigured`, so the agent can branch on the result without parsing raw command output. The remediation decision (whether/how to add `.WithEnvironment("AzureWebJobsSecretStorageType", "Files")`) stays with the agent.

## Affected file and lines

- [`aspire.md` — Step 1 detect (L47–L50)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-prepare/references/aspire.md#L47-L50)
- [`aspire.md` — Step 1a ExcludeFromManifest scan, bash + PowerShell (L57–L70)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-prepare/references/aspire.md#L57-L70)
- [`aspire.md` — Step 4b Functions / AzureWebJobsSecretStorageType scan, bash + PowerShell (L187–L211)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-prepare/references/aspire.md#L187-L211)
- [`generate.md` — "Check for .NET Aspire Projects FIRST" detection block (L9–L15)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-prepare/references/generate.md#L9-L15)
- [`scan.md` — ".NET Aspire Detection" (L57–L69)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-prepare/references/scan.md#L57-L69)

## Next steps

tmeschter:
Replaces the duplicated .NET Aspire detection sequence in the `azure-prepare` skill with cross-platform `detect-aspire.sh`/`detect-aspire.ps1` scripts.

The scripts live at `plugin/skills/azure-prepare/references/scripts/` and are invoked from the reference docs via doc-relative links.

- Scripts run the full detection in one pass and emit `key=value` fields (`isAspire`, `appHostPath`, `appHostDir`, `hasExcludeFromManifest`, `hasFunctions`, `secretStorageConfigured`) plus an informational summary.
- `aspire.md`, `generate.md`, and `scan.md` now invoke the script instead of re-deriving find/grep commands; all decision-making stays in the reference files.

### Review-feedback fixes
- Normalize `appHostPath`/`appHostDir` to workspace-relative, forward-slash paths in both shells (consistent cross-shell contract).
- Replace GNU-only `grep --include` with portable `find -exec grep` (macOS/BSD-safe).
- Use `-LiteralPath` in PowerShell `Get-ChildItem` to avoid wildcard interpretation of `[`, `]`, `*`, `?`.
- Prune `bin/`/`obj/` from `*.cs` scans.
- Sort AppHost candidates (case-insensitive) for deterministic, cross-shell-consistent selection.
- Treat `Aspire.AppHost.Sdk` as an Aspire indicator (scripts + `scan.md`/`aspire.md`).

Fixes #2494


wbreza (commented):
## Code Review — PR #2576

Automated multi-model review (Standard, 2 models: `claude-opus-4.7` + `gpt-5.3-codex`). Posted as **COMMENT** at author's discretion; findings include 2 High-priority items the author may want to address before merge.

### Summary

| Priority | Count |
|----------|-------|
| Critical | 0 |
| High     | 2 |
| Medium   | 3 |
| Low      | 1 |
| **Total**| **6** |

**Overall Assessment:** Comment — does not block, but 2 High findings (cross-shell `appHostPath` contract divergence; macOS-incompatible `grep --include`) are worth resolving in this PR or a quick follow-up.

### ✅ What Looks Good
- Single-pass detection with a flat `key=value` contract plus a separate `Summary:` block is a clean, LLM-friendly shape; field names are namespaced and extensible.
- Both scripts validate the workspace input, exit non-zero with a clear stderr message; bash uses `set -euo pipefail`; PS sets `$ErrorActionPreference = 'Stop'`.
- Doc updates correctly collapse three duplicated detection blocks into references to the script; within-skill markdown links stay inside the skill directory (no link-validator regression).
- Script invocation pattern (`./scripts/<name>.{sh,ps1}` shown in dual bash + PowerShell fences) matches the established convention used by `azure-quotas/check-quota.{sh,ps1}`.

### Findings Index
Inline comments are attached to specific lines for each finding. Where multiple files/lines are involved, one inline comment is posted per location.

1. **[High]** `appHostPath` / `appHostDir` format diverges between bash (relative) and PowerShell (absolute, backslashes). Source: consensus (both models). See inline on `detect-aspire.sh` and `detect-aspire.ps1`.
2. **[High]** Bash script uses GNU-only `grep --include`, which is unsupported by default BSD grep on macOS. Source: single-source (`gpt-5.3-codex`). See inline on `detect-aspire.sh`.
3. **[Medium]** PowerShell uses wildcard-interpreting `-Path` instead of `-LiteralPath` (workspaces with `[`, `]`, `*`, `?` silently return zero files). Source: single-source (`claude-opus-4.7`). See inline on `detect-aspire.ps1`.
4. **[Medium]** New scripts recurse through `bin/`/`obj/` — behavioral change vs. prior non-recursive inline commands; widens false-positive surface and increases scan cost. Source: single-source (`claude-opus-4.7`). See inline on both scripts.
5. **[Medium]** Multiple AppHost projects handled nondeterministically (first match wins, varies by traversal order). Source: single-source (`gpt-5.3-codex`). See inline on both scripts.
6. **[Low — Description Alignment]** PR description and file-tree summary state the scripts land at `plugin/skills/azure-prepare/scripts/`, but they actually land in `plugin/skills/azure-prepare/references/scripts/`. The doc-relative markdown links inside the skill are correct — only the PR body is wrong. Suggested fix: update the PR description.

---
*Reviewed with 2 models in parallel. Tags: `source: consensus` = flagged by both; `single-source: <model>` = flagged by one. Priority for finding #1 was `claude-opus-4.7=High, gpt-5.3-codex=Medium` → promoted to High based on impact (directly contradicts the PR's stated determinism goal).*

plugin/skills/azure-prepare/references/scripts/detect-aspire.sh
```diff
@@ -0,0 +1,110 @@
...
+# Defaults (emitted when the workspace is not an Aspire app)
+IS_ASPIRE="false"
+APPHOST_PATH=""
+APPHOST_DIR=""
+HAS_EXCLUDE_FROM_MANIFEST="false"
+HAS_FUNCTIONS="false"
+SECRET_STORAGE_CONFIGURED="false"
+
+# Step 1: Find the AppHost project
+APPHOST_PATH=$(find "$WORKSPACE_ROOT" -name "*.AppHost.csproj" 2>/dev/null | head -1 || true)
```
wbreza: **[High] `appHostPath`/`appHostDir` format diverges from PowerShell** *(consensus: opus + codex)*

Bash emits a relative POSIX path here (`./OrleansVoting.AppHost/...csproj`) because `find "$WORKSPACE_ROOT"` preserves the input prefix. The PowerShell script (lines ~67, ~73) emits an absolute Windows path with backslashes because it uses `$appHostProject.FullName`/`.DirectoryName`.

Same default invocation, materially different output → any agent consuming `appHostPath`/`appHostDir` produces platform-dependent results, contradicting the PR's determinism goal. The `aspire.md` example only shows the relative form.

**Suggested fix:** normalize both scripts to the same shape (absolute OR workspace-relative) AND normalize separator to `/`. Bash example:
```suggestion
APPHOST_PATH=$(cd "$(dirname "$APPHOST_PATH")" && pwd)/$(basename "$APPHOST_PATH")
```
In PS, after locating the project: `Resolve-Path -Relative` (with `Push-Location $WorkspaceRoot`) for the relative form. Update `aspire.md`'s Example output to match and document the contract explicitly.

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
```diff
@@ -0,0 +1,130 @@
...
+    if ($csprojFiles | Select-String -Pattern "Aspire.Hosting" -SimpleMatch -List -ErrorAction SilentlyContinue) {
+        $hasAspirePackage = $true
+    }
+}
+
+if ($appHostProject -or $hasAspirePackage) {
+    $isAspire = $true
+}
+
+if ($appHostProject) {
```
wbreza, line 60: **[High] Mirror of `appHostPath`/`appHostDir` divergence** *(consensus: opus + codex)*

`$appHostProject.FullName` / `.DirectoryName` always returns absolute Windows paths with backslashes. The bash counterpart emits relative POSIX paths. See companion comment on `detect-aspire.sh`. Pick one canonical shape and apply it here too.

plugin/skills/azure-prepare/references/scripts/detect-aspire.sh
```diff
@@ -0,0 +1,110 @@
...
+
+# Step 2: Confirm Aspire.Hosting package references anywhere in the workspace
+HAS_ASPIRE_PACKAGE="false"
+if grep -rql "Aspire.Hosting" "$WORKSPACE_ROOT" --include="*.csproj" 2>/dev/null; then
```
wbreza: **[High] `grep --include` is a GNU extension; breaks on macOS default tooling** *(single-source: `gpt-5.3-codex`)*

`grep --include` is not supported by BSD grep, which ships as the default on macOS. The flag is treated as a search pattern, causing the script to silently return wrong results or error on macOS users.

Also occurs at lines ~60, ~65, ~67. Recommend replacing all four with a portable `find ... -name "*.cs" -print0 | xargs -0 grep -lq ...` pattern (or `find ... -name ... -exec grep -lq ... {} +`).

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
```diff
@@ -0,0 +1,130 @@
...
+
+# Defaults (emitted when the workspace is not an Aspire app)
+$isAspire = $false
+$appHostPath = ""
+$appHostDir = ""
+$hasExcludeFromManifest = $false
+$hasFunctions = $false
+$secretStorageConfigured = $false
+
+# Step 1: Find the AppHost project
```
wbreza: **[Medium] Use `-LiteralPath` instead of `-Path` to avoid wildcard interpretation** *(single-source: `claude-opus-4.7`)*

`Test-Path` correctly uses `-LiteralPath` (line ~37), but `Get-ChildItem` here (and at ~line 55, ~line 70) uses `-Path`, which interprets `[`, `]`, `*`, `?` as wildcards. A workspace path like `C:\repos\my[client]\app` would silently return zero files, causing `isAspire=false` on a valid Aspire repo.

```suggestion
$appHostCandidates = @(Get-ChildItem -LiteralPath $WorkspaceRoot -Recurse -Filter "*.AppHost.csproj" -File -ErrorAction SilentlyContinue)
```
Apply the same change to the other two `Get-ChildItem` calls.

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
```diff
@@ -0,0 +1,130 @@
...
+}
+
+if ($appHostProject -or $hasAspirePackage) {
+    $isAspire = $true
+}
+
+if ($appHostProject) {
+    $appHostPath = $appHostProject.FullName
+
+    # Step 3: Derive the AppHost source directory
```
wbreza: **[Medium] Recursion change scans `bin/`/`obj/` — behavioral regression vs. prior inline commands** *(single-source: `claude-opus-4.7`)*

The previous in-doc PowerShell used `Get-ChildItem -Filter "*.cs"` (non-recursive). This now recurses, which scans `bin/`, `obj/`, and SDK-generated `*.g.cs` artifacts — widening the false-positive surface (e.g., generated DI registrations could contain `AddAzureFunctionsProject` strings as SDK evolves) and increasing scan cost on built solutions. Not noted in the PR description.

Either restrict to top-level `*.cs` in `$appHostDir`, or keep recursion but exclude `bin`/`obj`:
```suggestion
Get-ChildItem -LiteralPath $appHostDir -Recurse -Filter "*.cs" -File |
    Where-Object { $_.FullName -notmatch '[\\/](bin|obj)[\\/]' }
```

plugin/skills/azure-prepare/references/scripts/detect-aspire.sh
```diff
@@ -0,0 +1,110 @@
...
+    HAS_ASPIRE_PACKAGE="true"
+fi
+
+if [ -n "$APPHOST_PATH" ] || [ "$HAS_ASPIRE_PACKAGE" = "true" ]; then
+    IS_ASPIRE="true"
+fi
+
+if [ -n "$APPHOST_PATH" ]; then
+    # Step 3: Derive the AppHost source directory
+    APPHOST_DIR=$(dirname "$APPHOST_PATH")
```
wbreza: **[Medium] Bash counterpart to the `bin`/`obj` recursion concern** *(single-source: `claude-opus-4.7`)*

`grep -r ... "$APPHOST_DIR" --include="*.cs"` scans recursively, including `bin`/`obj/` and source-generator output. Prior inline command used `"$APPHOST_DIR"/*.cs` (top-level only). Either match prior semantics or prune `bin`/`obj`:
```suggestion
find "$APPHOST_DIR" -type d \( -name bin -o -name obj \) -prune -o -type f -name "*.cs" -print | xargs -r grep -lq "AddAzureFunctionsProject"
```
Same applies to the `AzureWebJobsSecretStorageType` / `ExcludeFromManifest` greps.

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
```diff
@@ -0,0 +1,130 @@
...
+# Defaults (emitted when the workspace is not an Aspire app)
+$isAspire = $false
+$appHostPath = ""
+$appHostDir = ""
+$hasExcludeFromManifest = $false
+$hasFunctions = $false
+$secretStorageConfigured = $false
+
+# Step 1: Find the AppHost project
+$appHostProject = Get-ChildItem -Path $WorkspaceRoot -Recurse -Filter "*.AppHost.csproj" -File -ErrorAction SilentlyContinue |
```
wbreza: **[Medium] Multiple AppHost projects: first-match wins is nondeterministic** *(single-source: `gpt-5.3-codex`)*

`Select-Object -First 1` silently picks one AppHost in repos with multiple candidates, and the choice can vary by filesystem ordering. Downstream `hasFunctions` / `secretStorageConfigured` then reflect only that one project.

Suggest collecting all candidates, sorting deterministically (e.g., by `FullName`), and either failing with an explicit ambiguity message or emitting `appHostCount`/`appHostCandidates` so callers can disambiguate.

plugin/skills/azure-prepare/references/scripts/detect-aspire.sh
wbreza: **[Medium] Bash counterpart to nondeterministic AppHost selection** *(single-source: `gpt-5.3-codex`)*

`find ... | head -1` mirrors the PowerShell first-match-wins behavior and inherits the same nondeterminism. Resolve in lock-step with the PS-side fix; sort and either fail-ambiguous or expose all candidates.

tmeschter: Fixed in 399550b8: both scripts now emit a workspace-relative, forward-slash `appHostPath`/`appHostDir`. Bash strips the workspace-root prefix; PowerShell uses `Resolve-Path -Relative` with `\`->`/` normalization. Verified identical output across bash and PowerShell against the Aspire sample apps (e.g. `./OrleansVoting.AppHost/OrleansVoting.AppHost.csproj`).

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
tmeschter: Fixed in 399550b8: both scripts now emit a workspace-relative, forward-slash `appHostPath`/`appHostDir`. Bash strips the workspace-root prefix; PowerShell uses `Resolve-Path -Relative` with `\`->`/` normalization. Verified identical output across bash and PowerShell against the Aspire sample apps (e.g. `./OrleansVoting.AppHost/OrleansVoting.AppHost.csproj`).

plugin/skills/azure-prepare/references/scripts/detect-aspire.sh
tmeschter: Fixed in 399550b8: replaced all GNU-only `grep --include` usages with a portable `find ... -exec grep -lE {} +` helper (`csproj_match`/`cs_match`), which works on BSD grep (the macOS default). Verified on the Aspire samples.

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
tmeschter: Fixed in 399550b8: all three `Get-ChildItem` calls now use `-LiteralPath` instead of `-Path`, so workspace paths containing `[`, `]`, `*`, or `?` are no longer treated as wildcards. Verified detection still returns `isAspire=true` for a sample copied into a `my[client]` path.

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
tmeschter: Fixed in 399550b8: the `*.cs` scans now prune `bin/` and `obj/` (bash via `find ... -prune`; PowerShell via a `-notmatch '[\\/](bin|obj)[\\/]'` filter), restoring the prior non-build-output scope.

plugin/skills/azure-prepare/references/scripts/detect-aspire.sh
tmeschter: Fixed in 399550b8: the `*.cs` scans now prune `bin/` and `obj/` (bash via `find ... -prune`; PowerShell via a `-notmatch '[\\/](bin|obj)[\\/]'` filter), restoring the prior non-build-output scope.

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
tmeschter: Fixed in 399550b8: AppHost candidates are sorted before the first match is selected (bash `sort -f | head -1`; PowerShell `Sort-Object FullName`). The sort is case-insensitive in both shells, so multi-AppHost repos resolve to the same project across bash and PowerShell — verified against the `samples` root (both select `./aspire-shop/AspireShop.AppHost/...`).

plugin/skills/azure-prepare/references/scripts/detect-aspire.sh
tmeschter: Fixed in 399550b8: AppHost candidates are sorted before the first match is selected (bash `sort -f | head -1`; PowerShell `Sort-Object FullName`). The sort is case-insensitive in both shells, so multi-AppHost repos resolve to the same project across bash and PowerShell — verified against the `samples` root (both select `./aspire-shop/AspireShop.AppHost/...`).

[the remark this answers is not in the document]
tmeschter: Fixed in 399550b8: AppHost candidates are sorted before the first match is selected (bash `sort -f | head -1`; PowerShell `Sort-Object FullName`). The sort is case-insensitive in both shells, so multi-AppHost repos resolve to the same project across bash and PowerShell — verified against the `samples` root (both select `./aspire-shop/AspireShop.AppHost/...`).

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
```diff
@@ -0,0 +1,130 @@
...
+    Select-Object -First 1
+
```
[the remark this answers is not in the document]
tmeschter: Fixed in 399550b8: AppHost candidates are sorted before the first match is selected (bash `sort -f | head -1`; PowerShell `Sort-Object FullName`). The sort is case-insensitive in both shells, so multi-AppHost repos resolve to the same project across bash and PowerShell — verified against the `samples` root (both select `./aspire-shop/AspireShop.AppHost/...`).

plugin/skills/azure-prepare/references/scripts/detect-aspire.sh
```diff
@@ -0,0 +1,110 @@
...
+SECRET_STORAGE_CONFIGURED="false"
+
+# Step 1: Find the AppHost project
+APPHOST_PATH=$(find "$WORKSPACE_ROOT" -name "*.AppHost.csproj" 2>/dev/null | head -1 || true)
+
+# Step 2: Confirm Aspire.Hosting package references anywhere in the workspace
+HAS_ASPIRE_PACKAGE="false"
+if grep -rql "Aspire.Hosting" "$WORKSPACE_ROOT" --include="*.csproj" 2>/dev/null; then
+    HAS_ASPIRE_PACKAGE="true"
+fi
```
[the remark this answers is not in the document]
tmeschter: Fixed in 399550b8: detection now treats `Aspire.AppHost.Sdk` as an Aspire indicator alongside `Aspire.Hosting` in both scripts; the 'no Aspire detected' summaries and the `isAspire` descriptions in `scan.md`/`aspire.md` were updated to match. Verified a csproj with only `Aspire.AppHost.Sdk` yields `isAspire=true` in both shells.

plugin/skills/azure-prepare/references/scripts/detect-aspire.sh
```diff
@@ -0,0 +1,110 @@
...
+echo "hasExcludeFromManifest=$HAS_EXCLUDE_FROM_MANIFEST"
+echo "hasFunctions=$HAS_FUNCTIONS"
+echo "secretStorageConfigured=$SECRET_STORAGE_CONFIGURED"
+
+# Human-readable summary
+echo ""
+echo "Summary:"
+if [ "$IS_ASPIRE" != "true" ]; then
+    echo "- No .NET Aspire app detected in '$WORKSPACE_ROOT' (no *.AppHost.csproj or Aspire.Hosting package reference)."
+    exit 0
```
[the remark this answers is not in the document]
tmeschter: Fixed in 399550b8: detection now treats `Aspire.AppHost.Sdk` as an Aspire indicator alongside `Aspire.Hosting` in both scripts; the 'no Aspire detected' summaries and the `isAspire` descriptions in `scan.md`/`aspire.md` were updated to match. Verified a csproj with only `Aspire.AppHost.Sdk` yields `isAspire=true` in both shells.

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
```diff
@@ -0,0 +1,130 @@
...
+# Step 2: Confirm Aspire.Hosting package references anywhere in the workspace
+$hasAspirePackage = $false
+$csprojFiles = Get-ChildItem -Path $WorkspaceRoot -Recurse -Filter "*.csproj" -File -ErrorAction SilentlyContinue
+if ($csprojFiles) {
+    if ($csprojFiles | Select-String -Pattern "Aspire.Hosting" -SimpleMatch -List -ErrorAction SilentlyContinue) {
+        $hasAspirePackage = $true
+    }
```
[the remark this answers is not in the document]
tmeschter: Fixed in 399550b8: detection now treats `Aspire.AppHost.Sdk` as an Aspire indicator alongside `Aspire.Hosting` in both scripts; the 'no Aspire detected' summaries and the `isAspire` descriptions in `scan.md`/`aspire.md` were updated to match. Verified a csproj with only `Aspire.AppHost.Sdk` yields `isAspire=true` in both shells.

plugin/skills/azure-prepare/references/scripts/detect-aspire.ps1
```diff
@@ -0,0 +1,130 @@
...
+Write-Output "hasExcludeFromManifest=$(ConvertTo-Lower $hasExcludeFromManifest)"
+Write-Output "hasFunctions=$(ConvertTo-Lower $hasFunctions)"
+Write-Output "secretStorageConfigured=$(ConvertTo-Lower $secretStorageConfigured)"
+
+# Human-readable summary
+Write-Output ""
+Write-Output "Summary:"
+if (-not $isAspire) {
+    Write-Output "- No .NET Aspire app detected in '$WorkspaceRoot' (no *.AppHost.csproj or Aspire.Hosting package reference)."
+    exit 0
```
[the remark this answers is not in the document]
tmeschter: Fixed in 399550b8: detection now treats `Aspire.AppHost.Sdk` as an Aspire indicator alongside `Aspire.Hosting` in both scripts; the 'no Aspire detected' summaries and the `isAspire` descriptions in `scan.md`/`aspire.md` were updated to match. Verified a csproj with only `Aspire.AppHost.Sdk` yields `isAspire=true` in both shells.

plugin/skills/azure-prepare/references/scan.md
```diff
@@ -56,16 +56,25 @@ Before classifying components, grep dependency files for SDKs that require a spe
...
+```bash
+./scripts/detect-aspire.sh [workspace-root]
+```
+
+**PowerShell:**
+```powershell
+./scripts/detect-aspire.ps1 -WorkspaceRoot <workspace-root>
+```
+
+A project is Aspire when the script returns `isAspire=true` (a `*.AppHost.csproj` or an `Aspire.Hosting` / `Aspire.Hosting.AppHost` package reference was found). See [aspire.md](aspire.md) Step 1 for the full field list.
```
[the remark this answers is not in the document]
tmeschter: Fixed in 399550b8: detection now treats `Aspire.AppHost.Sdk` as an Aspire indicator alongside `Aspire.Hosting` in both scripts; the 'no Aspire detected' summaries and the `isAspire` descriptions in `scan.md`/`aspire.md` were updated to match. Verified a csproj with only `Aspire.AppHost.Sdk` yields `isAspire=true` in both shells.

plugin/skills/azure-prepare/references/aspire.md
```diff
@@ -40,36 +40,49 @@ orleans-voting/
...
+./scripts/detect-aspire.ps1 -WorkspaceRoot <workspace-root>
 ```
 
-### ⛔ Step 1a: Pre-Check for Custom/Non-Deployable Resources (MANDATORY)
+`workspace-root` defaults to the current directory. The script reports these fields:
 
-**Before running `azd init --from-code`, scan the AppHost source code to understand whether the app may contain local-only custom resources.**
+| Field | Meaning |
+|-------|---------|
+| `isAspire` | `true` if a `*.AppHost.csproj` or `Aspire.Hosting` package reference was found |
```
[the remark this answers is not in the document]
tmeschter: Fixed in 399550b8: detection now treats `Aspire.AppHost.Sdk` as an Aspire indicator alongside `Aspire.Hosting` in both scripts; the 'no Aspire detected' summaries and the `isAspire` descriptions in `scan.md`/`aspire.md` were updated to match. Verified a csproj with only `Aspire.AppHost.Sdk` yields `isAspire=true` in both shells.

plugin/skills/azure-prepare/references/scan.md
```diff
@@ -56,16 +56,25 @@ Before classifying components, grep dependency files for SDKs that require a spe
 
 ### .NET Aspire Detection
 
-**.NET Aspire projects** are identified by:
-- A project ending with `.AppHost.csproj` (e.g., `OrleansVoting.AppHost.csproj`)
-- Reference to `Aspire.Hosting` or `Aspire.Hosting.AppHost` package in .csproj files
-- Multiple .NET projects in a solution, typically including an AppHost orchestrator
+Detect **.NET Aspire projects** by running the detection script ([detect-aspire.sh](scripts/detect-aspire.sh) / [detect-aspire.ps1](scripts/detect-aspire.ps1)), which reports `isAspire` along with AppHost path, `ExcludeFromManifest`, and Azure Functions signals:
```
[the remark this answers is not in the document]
tmeschter: Not changing this: per the established `azure-prepare`/`azure-quotas` convention, bundled `./scripts/*` are invoked directly from the skill's reference scripts via doc-relative links rather than copied into the user's workspace. Adding copy-into-workspace + `chmod +x` instructions here would diverge from that convention, so we're intentionally keeping the existing invocation pattern. Resolving as won't-fix.

plugin/skills/azure-prepare/references/generate.md
```diff
@@ -4,16 +4,20 @@ Generate infrastructure and configuration files based on selected recipe.
 
 ## ⛔ CRITICAL: Check for .NET Aspire Projects FIRST
 
-**MANDATORY: Before generating any files, detect .NET Aspire projects:**
+**MANDATORY: Before generating any files, detect .NET Aspire projects** using the detection script ([detect-aspire.sh](scripts/detect-aspire.sh) / [detect-aspire.ps1](scripts/detect-aspire.ps1)). It runs the full detection sequence and prints `key=value` fields plus a summary:
```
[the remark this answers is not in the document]
tmeschter: Not changing this: per the established `azure-prepare`/`azure-quotas` convention, bundled `./scripts/*` are invoked directly from the skill's reference scripts via doc-relative links rather than copied into the user's workspace. Adding copy-into-workspace + `chmod +x` instructions here would diverge from that convention, so we're intentionally keeping the existing invocation pattern. Resolving as won't-fix.

plugin/skills/azure-prepare/references/aspire.md
```diff
@@ -40,36 +40,49 @@ orleans-voting/
 
 ### Step 1: Detection
 
-When scanning the codebase (per [scan.md](scan.md)), detect Aspire by:
+When scanning the codebase (per [scan.md](scan.md)), run the detection script ([detect-aspire.sh](scripts/detect-aspire.sh) / [detect-aspire.ps1](scripts/detect-aspire.ps1)). It performs the full deterministic detection sequence in one pass and prints `key=value` lines plus a human-readable summary, so you can branch on the result instead of parsing raw `find`/`grep` output.
```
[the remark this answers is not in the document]
tmeschter: Not changing this: per the established `azure-prepare`/`azure-quotas` convention, bundled `./scripts/*` are invoked directly from the skill's reference scripts via doc-relative links rather than copied into the user's workspace. Adding copy-into-workspace + `chmod +x` instructions here would diverge from that convention, so we're intentionally keeping the existing invocation pattern. Resolving as won't-fix.

JasonYeMSFT (dismissed):

tmeschter:
@wbreza This is ready for another look.

tmeschter:
I'm converting this back to a draft. While the limited data from my integration test runs gives me reason to believe this reduces token counts, the script is not being invoked consistently.

JasonYeMSFT (approved):

plugin/skills/azure-prepare/references/scripts/gather-aspire-info.sh
```diff
@@ -0,0 +1,141 @@
...
+IS_ASPIRE="false"
+APPHOST_PATH=""
+APPHOST_DIR=""
+HAS_EXCLUDE_FROM_MANIFEST="false"
+HAS_FUNCTIONS="false"
+SECRET_STORAGE_CONFIGURED="false"
+
+# Step 1: Find the AppHost project (case-insensitive sort for deterministic,
+# cross-shell-consistent selection)
+APPHOST_RAW=$(find "$WORKSPACE_ROOT" -type f -name "*.AppHost.csproj" 2>/dev/null | sort -f | head -1 || true)
```
jongio, line 71: Steps 1-2 here (and the `to_relative`/`csproj_match` helpers) are identical to detect-aspire.sh. Same pattern in the PS1 pair. That's ~30 lines duplicated across 4 files, so detection criteria changes need 4-file lockstep updates.

One option: have `gather-aspire-info` call `detect-aspire` first (or source it), then layer on the deeper scans (Steps 3-5). That keeps the quick-check script self-contained while the full-facts script avoids the duplication.

Not blocking since self-contained scripts are easier for agents to invoke, but worth considering for maintainability.

plugin/skills/azure-prepare/references/scripts/gather-aspire-info.sh
```diff
@@ -0,0 +1,141 @@
...
+
+if [ -n "$APPHOST_RAW" ] || [ "$HAS_ASPIRE_PACKAGE" = "true" ]; then
+    IS_ASPIRE="true"
+fi
+
+if [ -n "$APPHOST_RAW" ]; then
+    APPHOST_PATH=$(to_relative "$APPHOST_RAW")
+
+    # Step 3: Derive the AppHost source directory (scan the real path on disk)
+    APPHOST_DIR=$(dirname "$APPHOST_PATH")
```
jongio, line 87: If a `*.AppHost.csproj` sits directly in `WORKSPACE_ROOT`, `dirname` gives `.`, so `APPHOST_DIR_RAW` becomes the workspace root itself. The subsequent `cs_match` calls then scan ALL `.cs` files in the entire workspace. Unlikely for real Aspire apps (AppHost is almost always in a subdirectory), but could cause noticeable slowness on large repos.

jongio (commented):
Two items on top of @wbreza's review:

1. Detection logic duplication across the four scripts (see inline on gather-aspire-info.sh)
2. Edge case when AppHost sits at workspace root (see inline on gather-aspire-info.sh)

The PR also has merge conflicts with main that need resolving.

PR #2602, opened 2026-06-08
misc: Remove a workflow input

tmeschter:
Remove the `publish-reports` input from test-azure-deploy.yml. This would be set to true when test-azure-deploy.yml was kicked off by a _scheduled_ run of test-all-integration.yml, but would be false otherwise. This meant that manually started runs against feature/dev branches wouldn't end up recording their token usage or saving test reports to Azure Storage blobs.

## Checklist

- [X] Title has one of the prefixes: `fix:`, `feat:`, `feature:`, `chore:`, `misc:`, `test:`, `eval:`

### DIFF
--- a/.github/workflows/test-azure-deploy.yml
+++ b/.github/workflows/test-azure-deploy.yml
@@ -45,11 +45,6 @@ on:
         required: false
         type: boolean
         default: false
-      publish-reports:
-        description: 'Whether to publish reports to Azure Storage'
-        required: false
-        type: boolean
-        default: false
     secrets:
       COPILOT_CLI_TOKEN:
         required: true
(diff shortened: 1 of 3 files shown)


.github/workflows/test-azure-deploy.yml
```diff
@@ -223,7 +218,7 @@ jobs:
 
       # Note: The managed identity must have write permission to the target storage account
       - name: Publish report to Azure Storage
```
JasonYeMSFT, line 220: Could you follow the example in test-all-integration to add another step that publish the same artifacts except to a different storage account when the workflow is triggered manually?

https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/c3c7d981ec45c71d6931fc2f1eb440953ac18aa7/.github/workflows/test-all-integration.yml#L368

tmeschter: @JasonYeMSFT Done.

.github/workflows/test-azure-deploy.yml
```diff
@@ -223,7 +218,7 @@ jobs:
...
-        if: always() && inputs.publish-reports && vars.REPORT_STORAGE_ACCOUNT != ''
+        if: always() && vars.REPORT_STORAGE_ACCOUNT != ''
```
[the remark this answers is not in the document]
tmeschter: Done.

JasonYeMSFT (approved):

PR #2608, opened 2026-06-09
azure-upgrade, replace personal repo in integration test with repos under well-known orgs

XiaofeiCao:
Replaces the personal `weidongxu-microsoft/java-update-examples` repo in `tests/azure-upgrade/integration.test.ts` with the three well-known repos. Each entry is pinned to the latest commit on its default branch for reproducibility.

## Changes

| Scenario | Old (folder in java-update-examples) | New upstream (pinned) |
|---|---|---|
| File-based Azure auth -> `DefaultAzureCredential` | `azure-legacy-sdk-update-azure-client-initialization` | [`Azure-Samples/aad-java-manage-service-principals`](https://github.com/Azure-Samples/aad-java-manage-service-principals/blob/master/src/main/java/com/microsoft/azure/management/graphrbac/samples/ManageServicePrincipal.java) @ `83b4460` |
| EventProcessorHost `InMemory*` -> `BlobCheckpointStore` | `azure-legacy-sdk-update-eventhubs-v3` | [`logstash-plugins/logstash-input-azure_event_hubs`](https://github.com/logstash-plugins/logstash-input-azure_event_hubs/blob/main/.ci/integration/event_hub_consumer/src/main/java/com/microsoft/azure/eventprocessorhost/Consumer.java) @ `4a05eea` (sparse: `.ci/integration/event_hub_consumer`) |
| Batch `defineNewApplicationPackage` -> `applicationPackages().define()` | `azure-legacy-sdk-update-batch-java-manage-batch-accounts` | [`Azure-Samples/batch-java-manage-batch-accounts`](https://github.com/Azure-Samples/batch-java-manage-batch-accounts/blob/master/src/main/java/com/microsoft/azure/management/batch/samples/ManageBatchAccount.java) @ `105a076` |

`runJavaMigration` now takes a `{ repoUrl, commitId, sparseCheckoutPath? }` config and only adds the *""project can be found under ...""* hint when a sparse subpath is used.

## Test results

All three `java-sdk-migration-e2e` tests pass against the new baselines:

- `migrates legacy file-based Azure auth (client init) to DefaultAzureCredential` -> pass
- `migrates EventProcessorHost InMemory managers to BlobCheckpointStore` -> pass
- `migrates Batch defineNewApplicationPackage chain to applicationPackages().define()` -> pass

image

Migration assertions and skill-invocation tests are unchanged.


tests/azure-upgrade/integration.test.ts
```diff
@@ -50,8 +50,25 @@ const SKILL_NAME = "azure-upgrade";
...
+  repoUrl: "https://github.com/Azure-Samples/aad-java-manage-service-principals.git",
+  commitId: "83b446085a7ef35cf26b6d36255aa833aeb287bb",
+} as const;
+
+const EVENT_PROCESSOR_HOST_REPO = {
+  repoUrl: "https://github.com/logstash-plugins/logstash-input-azure_event_hubs.git",
+  commitId: "4a05eea79342f1d451162d1a3e26270f6a484b4f",
+  // The Java sample lives in a subdirectory of this primarily-Ruby repo.
+  sparseCheckoutPath: ".ci/integration/event_hub_consumer",
+} as const;
```
weidongxu-microsoft, line 66: nit, this is not our org?

weidongxu-microsoft (approved):

JasonYeMSFT:
FYI, we are migrating testing from old integration tests to vally eval suites. Here is a PR converting azure-upgrade test cases. Could you please help review and merge the migration PR and then apply these changes to the vally eval suites?
https://github.com/microsoft/GitHub-Copilot-for-Azure/pull/2648

tests/azure-upgrade/integration.test.ts
XiaofeiCao: No. I assume logstash-plugins be well-known org, and will not be removed any time sooner.

tests/azure-upgrade/integration.test.ts
```diff
@@ -0,0 +1,404 @@
+/**
```
JasonYeMSFT, line 1: Could you delete this file? We no longer run js tests for azure-upgrade in our scheduled tests.

evals/azure-upgrade/eval.yaml
```diff
@@ -250,9 +250,9 @@ stimuli:
         - "Continue with recommended options until complete."
     environment:
       commands:
-        - git clone --depth 1 --sparse https://github.com/weidongxu-microsoft/java-update-examples.git
-        - cd java-update-examples && git sparse-checkout set azure-legacy-sdk-update-azure-client-initialization
-        - cd java-update-examples && git fetch --depth 1 origin 6d071296df8929482b0903241a23713a0bb952a4 && git checkout 6d071296df8929482b0903241a23713a0bb952a4
+        - git clone --depth 1 --sparse https://github.com/Azure-Samples/aad-java-manage-service-principals.git
```
jongio, line 253: `--sparse` enables sparse-checkout mode but no `git sparse-checkout set` follows. The default cone pattern only materializes root-level files. I checked the repo tree at this commit and Java sources live under `src/`, so they won't be checked out.

Drop `--sparse` since this repo doesn't need sparse checkout:
```suggestion
        - git clone --depth 1 https://github.com/Azure-Samples/aad-java-manage-service-principals.git
```

Same fix needed on line 364 for `batch-java-manage-batch-accounts`.

evals/azure-upgrade/eval.yaml
```diff
@@ -250,9 +250,9 @@ stimuli:
         - "Continue with recommended options until complete."
     environment:
       commands:
-        - git clone --depth 1 --sparse https://github.com/weidongxu-microsoft/java-update-examples.git
-        - cd java-update-examples && git sparse-checkout set azure-legacy-sdk-update-azure-client-initialization
-        - cd java-update-examples && git fetch --depth 1 origin 6d071296df8929482b0903241a23713a0bb952a4 && git checkout 6d071296df8929482b0903241a23713a0bb952a4
+        - git clone --depth 1 --sparse https://github.com/Azure-Samples/aad-java-manage-service-principals.git
+        - cd aad-java-manage-service-principals
```
jongio, line 254: This bare cd is a no-op. The next command already starts with cd aad-java-manage-service-principals && ..., so this line does nothing. It was meaningful before when git sparse-checkout set was chained here.

Remove it (same for line 365):
```suggestion
        - cd aad-java-manage-service-principals && git fetch --depth 1 origin 83b446085a7ef35cf26b6d36255aa833aeb287bb && git checkout 83b446085a7ef35cf26b6d36255aa833aeb287bb
```

jongio (dismissed):
Two issues in eval.yaml need fixing:

1. `git clone --depth 1 --sparse` is used for `aad-java-manage-service-principals` and `batch-java-manage-batch-accounts`, but no `git sparse-checkout set` follows. The default sparse-checkout cone only materializes root-level files. Both repos keep their Java sources under `src/`, so those files won't be checked out. This likely explains the failing `eval` CI check. Drop `--sparse` from these two clone commands (keep it for the logstash repo that genuinely uses sparse checkout).

2. The bare `cd repo-name` lines (254, 365) are no-ops left over from when `git sparse-checkout set` was chained there. Remove them or fold the fetch into that line.

Separately, @JasonYeMSFT flagged that integration tests are migrating to eval suites and asked to delete `tests/azure-upgrade/integration.test.ts`. That should be resolved before merging.

jongio (commented):
The `eval` CI check is failing. The likely root cause: `--sparse` is passed to `git clone` for `aad-java-manage-service-principals` (line 252) and `batch-java-manage-batch-accounts` (line 363), but no `git sparse-checkout set` follows either clone. The default sparse-checkout cone only materializes root-level files, so `src/` (where the Java sources live) won't be checked out.

Fix: drop `--sparse` from those two clones. The logstash repo correctly uses sparse checkout because it chains `git sparse-checkout set .ci/integration/event_hub_consumer` on the next line.

Lines 253 and 364 (`cd <repo-name>`) are also no-ops. The subsequent command already starts with `cd <repo-name> && ...`, making the standalone `cd` redundant. Remove both lines.

Separately: @JasonYeMSFT asked to delete `tests/azure-upgrade/integration.test.ts` since integration tests are migrating to eval suites (see #2648). This PR adds a 404-line version of that file. Please resolve that conflict before merging.

evals/azure-upgrade/eval.yaml
```diff
@@ -250,9 +250,9 @@ stimuli:
         - "Continue with recommended options until complete."
     environment:
       commands:
```
jongio, line 252: `--sparse` enables sparse-checkout mode but no `git sparse-checkout set` follows. The default cone pattern only materializes root-level files. Java sources live under `src/`, so they won't be checked out. This is likely why the `eval` CI check is failing.

Drop `--sparse` since this repo doesn't need sparse checkout:

```suggestion
        - git clone --depth 1 https://github.com/Azure-Samples/aad-java-manage-service-principals.git
```

evals/azure-upgrade/eval.yaml
```diff
@@ -361,9 +361,9 @@ stimuli:
...
```
jongio, line 363: Same issue here: `--sparse` without a subsequent `git sparse-checkout set` means `src/` won't be checked out.

```suggestion
        - git clone --depth 1 https://github.com/Azure-Samples/batch-java-manage-batch-accounts.git
```

tests/azure-upgrade/integration.test.ts
```diff
@@ -0,0 +1,404 @@
...
+        shouldEarlyTerminate: (m) => {
+          if (!isSkillInvoked(m, SKILL_NAME)) {
+            return getToolCalls(m).length > 30;
+          }
+          const navigated = /redis-to-amr\.md/.test(getAllToolText(m));
+          const surfaced = doesAssistantOrToolsIncludeKeyword(m, "amr-migration-skill");
+          return (navigated && surfaced) || getToolCalls(m).length > 30;
+        }
+      });
+
```
jongio, line 160: Trailing underscore in the describe block name: the suite reports as `azure-upgrade_ - Integration Tests`.

```suggestion
describeIntegration(`${SKILL_NAME} - Integration Tests`, () => {
```

jongio (commented):
One new nit since my last review (trailing underscore in the describe block name). My prior eval.yaml inline comments about `--sparse` without `sparse-checkout set` and no-op `cd` lines remain unresolved and are the likely cause of the failing eval CI check.

PR #2661, opened 2026-06-17
eval: migrate foundry agent lifecycle integration tests to vally

anchenyi:
## Description

migrate foundry agent lifecycle integration tests to vally

## Checklist


evals/microsoft-foundry/eval.yaml
```diff
@@ -265,6 +265,124 @@ stimuli:
...
+  # ═══════════════════════════════════════════════════════════════════════════
+
+  - name: "Create Python hosted Foundry agent"
+    prompt: "Create a new hosted agent for Foundry using Python."
+    tags:
+      type: integration
+      tier: smoke
+      cost: llm
+      area: routing
+      earlyTerminate: '[{"type":"tool-call-match","toolPattern":"^create$","argsPattern":".*"}]'
```
JasonYeMSFT: The grader only checks for skill invocation, if you don't check the content written by the create tool call, terminating on skill invocation will terminate the test earlier. Otherwise, add a grader that checks if the content has expected patterns. You can update this in this PR or do it in your next iteration on foundry eval suites.

Although I haven't seen this in a while, we have seen agents occasionally decide to write to files via terminal commands like `echo "the file content" > file.txt` . When it happens, the early terminate condition won't work as expected. This is not something we can address right now but just fyi.

JasonYeMSFT (dismissed):

evals/microsoft-foundry/eval.yaml
anchenyi: Fixed. Thanks.

JasonYeMSFT (approved):

PR #2663, opened 2026-06-18
feat: add azure-advisor skill

sharadhisrikanth28:
## Summary
Adds the `azure-advisor` product-area router skill, ported and adapted to repo conventions.

### What's included
- **Skill**: `plugin/skills/azure-advisor/` with a read-only **review** capability that uses whichever `advisor_*` MCP tools are available (routes by capability, not hard-coded tool names).
- **Shared references**: capability-routing and subscription-discovery.
- **Tests**: trigger tests under `tests/azure-advisor/` (16 passing).
- **Registration**: added to `tests/skills.json` (skills array + `0 12 * * 2-6` integration batch).
- **Versioning**: `version.json` added; frontmatter uses `0.0.0-placeholder` (NBGV stamps at build).

### Validation
- References validator passes
- Build / version stamping (stamped 1.1.71)
- Trigger tests 16/16
- Typecheck + lint clean


jongio (commented):
Testing standards gap per `tests/AGENTS.md`:

1. The "Should NOT Trigger" section has only 4 negative prompts (minimum is 5).
2. None of the negative prompts test adjacent Azure services. The AGENTS.md guide explicitly requires prompts about "different Azure services not covered by this skill." Since the description's DO NOT USE FOR section mentions `azure-cost` and `azure-diagnostics`, boundary prompts like "analyze my Azure costs" or "troubleshoot my Azure subscription" are needed to verify the skill doesn't false-positive on adjacent domains. This matters because keyword extraction pulls "cost" and "diagnostics" into the trigger set from the DO NOT USE FOR text.
3. No `unit.test.ts` file is present. The testing checklist requires unit tests that verify the skill's content contains expected sections.

The `eval` CI check is also failing. Worth confirming whether it's related to this PR.

tests/azure-advisor/triggers.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        const result = triggerMatcher.shouldTrigger(prompt);
+        expect(result.triggered).toBe(false);
+      }
+    );
+  });
+
+  describe("Trigger Keywords Snapshot", () => {
+    test("skill keywords match snapshot", () => {
+      expect(triggerMatcher.getKeywords()).toMatchSnapshot();
+    });
```
jongio: The repo testing guide (`tests/AGENTS.md`) requires at least 5 negative prompts including prompts about "different Azure services not covered by this skill." This array has 4 prompts, all completely unrelated topics.

Consider adding boundary cases like:
```suggestion
    const shouldNotTriggerPrompts: string[] = [
      "What is the weather today?",
      "Help me write a poem",
      "Explain quantum computing",
      "How do I write a Python web scraper?",
      "Analyze my Azure costs and spending trends",
      "Troubleshoot my Azure App Service deployment failure",
      "Help me set up Azure Key Vault for secrets management",
    ];
```
The last three test that prompts targeting `azure-cost`, `azure-diagnostics`, and `azure-keyvault` don't accidentally activate this skill. The extracted keywords include "cost" and "diagnostics" (pulled from the DO NOT USE FOR text), so these boundaries are important to validate.

tests/azure-advisor/triggers.test.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Tests that verify the skill triggers on appropriate prompts
+ * and does NOT trigger on unrelated prompts.
+ *
+ */
+
+import { TriggerMatcher } from "../utils/trigger-matcher";
+import { loadSkill, LoadedSkill } from "../utils/skill-loader";
+
+const SKILL_NAME = "azure-advisor";
+
```
jongio: This triggers file is present but `unit.test.ts` is missing. Every other skill has one (check `tests/_template/unit.test.ts`). Per the testing checklist, it needs to verify skill metadata and content sections.

jongio (dismissed):
Two things need fixing before this can merge:

1. No `unit.test.ts`. Every other skill in the repo has one (the template includes it). `tests/AGENTS.md` Step 5 and the checklist require unit tests that verify skill metadata and content sections.

2. The `shouldNotTriggerPrompts` array has 4 entries; minimum is 5. None test adjacent Azure services. Since the skill's `DO NOT USE FOR` calls out `azure-cost` and `azure-diagnostics`, add boundary prompts like "analyze my Azure subscription costs" or "troubleshoot my Azure VM" to verify the skill doesn't false-positive on neighboring domains.

The Copilot review also flagged "IaaC" typos (should be "IaC") and `.env*` scanning in `subscription-discovery.md`. Worth fixing in the same pass.

tests/azure-advisor/triggers.test.ts
sharadhisrikanth28: `shouldNotTriggerPrompts` now has 9 entries, including the adjacent-Azure-service boundaries that were flagged.

Re: the failing eval check — that's unrelated to this PR. It fails on azure-compute, a pre-existing issue in another skill.

tests/azure-advisor/triggers.test.ts
sharadhisrikanth28: Re: `unit.test.ts` — the `tests/_template` that reference points to, no longer exists, and most current skills ship only triggers.test.ts + Vally evals per `AGENTS.md`.
Left it out to match the current pattern, but happy to add one to this skill if you'd like.

plugin/skills/azure-advisor/README.md
```diff
@@ -0,0 +1,54 @@
...
+| [references/capability-routing.md](references/capability-routing.md) | How to resolve an `advisor_*` MCP tool by capability (catalog, recommendations, summary, IaC fix). |
+| [references/subscription-discovery.md](references/subscription-discovery.md) | How to resolve the target subscription from repo config / env without hardcoding. |
+| [review/review.md](review/review.md) | **Capability:** holistic read-only Advisor sweep. |
+| [review/review.md](review/review.md) | The review workflow (the only file the agent reads for this capability). |
+
+## Conventions (match the rest of the repo)
+
+- **One folder per capability**, named after the capability (e.g. `review/`,
+  `summarize/`). This mirrors `microsoft-foundry` (`quota/quota.md`, `rbac/rbac.md`).
+- **Capability file is named after its folder**: `review/review.md`,
```
jongio: Duplicate table entries: `references/capability-routing.md` appears on two rows with different purpose text, and `review/review.md` also appears twice. Looks like a copy-paste artifact. Consider merging each pair into one row, e.g.:

```suggestion
| [references/capability-routing.md](references/capability-routing.md) | **Shared.** Capability-agnostic routing docs: resolve which `advisor_*` MCP tool to call for each capability (catalog, recommendations, summary, IaC fix). |
| [references/subscription-discovery.md](references/subscription-discovery.md) | How to resolve the target subscription from repo config / env without hardcoding. |
| [review/review.md](review/review.md) | **Capability:** holistic read-only Advisor sweep (the only file the agent reads for this capability). |
```

jongio (commented):
Correction on my earlier unit.test.ts feedback: no other skill in this repo actually has one (checked azure-compute, azure-deploy, azure-validate, azure-prepare, microsoft-foundry). The AGENTS.md documents it but nobody follows it. Withdrawing that as a blocker.

One remaining nit: README.md folder map table has duplicate rows for `references/capability-routing.md` and `review/review.md`. Each path appears twice with different purpose text. Consider collapsing each pair into a single row.

plugin/skills/azure-advisor/README.md
```diff
@@ -0,0 +1,54 @@
...
+user intent. Each capability is a self-contained folder; shared concerns live in the
+[references](references/capability-routing.md) folder.
+
+## Folder map
+
+| Path | Purpose |
+|------|---------|
+| [SKILL.md](SKILL.md) | **Router.** Frontmatter (makes the skill discoverable) + a capability table that routes intent to a capability file. Keep it thin. |
+| [references/capability-routing.md](references/capability-routing.md) | **Shared, capability-agnostic docs** reused by every capability. Don't duplicate these inside a capability. |
+| [references/capability-routing.md](references/capability-routing.md) | How to resolve an `advisor_*` MCP tool by capability (catalog, recommendations, summary, IaC fix). |
```
sachinsoni29: Duplicate rows, keep single row for capability and review

plugin/skills/azure-advisor/SKILL.md
```diff
@@ -0,0 +1,64 @@
+---
+name: azure-advisor
+description: "Azure Advisor reviews and recommendations using available advisor_* MCP tools. WHEN: \"run an advisor review\", \"check my Azure advisor recommendations\", \"summarize advisor findings\", \"what does Advisor say about my subscription\", \"give me an advisor health check\", \"audit my Azure resources with Advisor\". USE FOR: read-only Advisor sweeps with catalog, recommendations, and IaC fixes. DO NOT USE FOR: changing resources, billing analysis (use azure-cost), or non-Advisor troubleshooting (use azure-diagnostics)."
```
sachinsoni29: Update start to -> "Azure Advisor reviews resources and provides recommendations...."

plugin/skills/azure-advisor/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+
+Planned capabilities will each be added as a sibling folder under `azure-advisor/`,
+reusing the shared references above. Add a row to the table above when one ships:
+
+- **summarize** — focused recommendation summarization / aggregation views
+- **resource-analysis** — per-resource Advisor analysis and drill-down
+- **greenfield** — Advisor-informed guidance for new/empty subscriptions
+- **cost** — Advisor cost-category optimization (coordinates with `azure-cost`)
+- **reliability** — Advisor reliability-category reviews
+- **governance** — Advisor operational-excellence / governance reviews
```
sachinsoni29: We can also add security and performance reviews to the roadmap here. But consider it as optional

plugin/skills/azure-advisor/README.md
sharadhisrikanth28: Collapsed the duplicate rows so each path appears once with merged purpose text. Thanks for withdrawing the `unit.test.ts` blocker.

tests/azure-advisor/triggers.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    });
+
+    test("is case insensitive", () => {
+      const result1 = triggerMatcher.shouldTrigger(
+        "RUN AN AZURE ADVISOR REVIEW"
+      );
+      const result2 = triggerMatcher.shouldTrigger(
+        "run an azure advisor review"
+      );
+      expect(result1.triggered).toBe(result2.triggered);
```
sachinsoni29: Check one of the results to be true as well here.

plugin/skills/azure-advisor/review/review.md
```diff
@@ -0,0 +1,118 @@
...
+## Constraints
+
+- ✅ **Always** discover the subscription from repo files / env first; ask only as last resort.
+- ✅ **Always** include every advisor_* tool name actually invoked in the "Tools used" line so the user can see what the skill chose.
+- ✅ **Always** mention which steps were skipped because no matching capability was found.
+- ❌ **Never** hardcode a subscription id, tenant id, or resource group.
+- ❌ **Never** modify Azure state — this sub-skill is read + suggest only.
+- ❌ **Never** call a tool whose name does not contain `advisor_` as a substitute; if no capability matches, report it and skip. (Match on the substring `advisor_` — clients prepend a server-name prefix like `azure-mcp-`.)
+- ❌ **Never** write the summary to a file unless the user explicitly asks for one in their prompt.
+- ❌ **Never** write helper scripts, scratch files, or parsing utilities to disk (no `.tmp/*.js`, no `.tmp/*.json`, no temp files of any kind). Reason over MCP tool responses directly in-context. Prefer aggregating via the `advisor_recommendation_summary` tool over computing counts yourself.
```
jongio: Both this constraint and the one below reference `advisor_recommendation_summary` by exact name. The capability-routing reference doc says "match tools by capability description, never by hard-coded name," and the rest of the skill follows that principle.

Consider replacing both with references to the Aggregation/Summary capability:
- "Prefer aggregating via the Aggregation/Summary capability (see [Capability Routing](../references/capability-routing.md)) over computing counts yourself."
- "If you need server-side aggregation, use the Aggregation/Summary capability from [Capability Routing](../references/capability-routing.md)."

jongio (commented):
My prior changes-request items are all resolved: negative prompts now cover adjacent-service boundaries (9 total), IaaC typos are fixed, and the README folder-map duplicates are collapsed.

One consistency issue remains: review.md references `advisor_recommendation_summary` by exact name in its constraints, but the rest of the skill routes by capability, not by hard-coded tool name. See inline comment.

The eval CI failure is pre-existing (azure-compute) and isn't related to this PR.

plugin/skills/azure-advisor/references/subscription-discovery.md
```diff
@@ -0,0 +1,18 @@
+# Shared Reference — Subscription Discovery
+
+> **Shared across all `azure-advisor` capabilities.** Any capability that operates on a
+> subscription should link here instead of re-defining discovery logic.
+
+Subscription **must never be hardcoded**. Resolve in this order and stop at the first hit:
```
sachinsoni29: Better to not stop at the first sub, rather scan all subscriptions and then run the skill for all Subscriptions and provide results accordingly, as this could lead to us missing the important Production subscriptions for a user.

plugin/skills/azure-advisor/review/review.md
```diff
@@ -0,0 +1,118 @@
...
+  `advisor_*` MCP tool by capability (catalog, recommendations, summary, IaC fix).
+- **[Subscription Discovery](../references/subscription-discovery.md)** — how to resolve
+  the target subscription from repo config / env without hardcoding.
+
+## Workflow
+
+### Step 1 — Resolve subscription
+
+Run [Subscription Discovery](../references/subscription-discovery.md). Save the resolved
+value; mention its *source* in the final chat summary ("Subscription pulled from
```
sachinsoni29: values (in case of multiple subscriptions)

plugin/skills/azure-advisor/review/review.md
```diff
@@ -0,0 +1,118 @@
...
+`infra/main.parameters.json`").
+
+### Step 2 — Probe the catalog (metadata)
+
+Invoke the **metadata / catalog** capability (see
+[Capability Routing](../references/capability-routing.md)) with no filter to learn what
```
sachinsoni29, line 63: Instead of no filter, We should have a step to discover the relevant resource Types part of Infra Files within the repo and add those as filters to one by one invoke metadata tool for each resource type

OR 

Get the results without any filters but only show results belonging to specific resource types part of repo files.

plugin/skills/azure-advisor/review/review.md
sharadhisrikanth28: Fixed — both constraints now reference the Aggregation / summary capability (via Capability Routing) instead of the hard-coded `advisor_recommendation_summary` name, so the whole skill routes by capability consistently.

plugin/skills/azure-advisor/review/review.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Step 6 — Compose chat summary
+
+Reply in chat with this structure (no files written):
+
+```
+## Azure Advisor Review
+
+**Subscription:** <id> (from <source>)
+**Tenant catalog:** N categories, M impact levels, K resource types
```
sachinsoni29: Displaying the whole catalog of recommendations does not make sense. Only specific to customer resource types is better.

Also, remove Tenant Keyword from output.

sachinsoni29 (changes requested):

plugin/skills/azure-advisor/review/review.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Error Handling
+
+| Symptom | Probable cause | Action |
+|---|---|---|
+| No tool name contains `advisor_` | MCP server not configured / not running, **or** a strict starts-with match rejected prefixed names | Substring-match on `advisor_` (names look like `azure-mcp-advisor_*`). If still none, tell user to check `.vscode/mcp.json`, that `azmcp.exe` is reachable, and that tools are exposed individually (`--mode all`). |
+| Tenant-wide enumeration requested but no subscription-list tool | Azure MCP subscription tool not exposed | Fall back to repo-discovered subscriptions; if none, ask the user. |
+| Catalog call returns empty | Tenant has no Advisor coverage yet | Stop after Step 2; report empty tenant. || Recommendation list 401/403 | Auth not scoped to subscription | Tell user to run `az login` and verify subscription access. |
```
jongio: These two table rows got concatenated onto one line. The `Catalog call returns empty` row and the `Recommendation list 401/403` row are separated by `||` instead of a line break. Markdown won't render the second row as a table entry. Split them onto separate lines.

jongio (commented):
Prior changes-request items are all resolved: capability routing now references capability-routing.md instead of hard-coded tool names.

One formatting defect in the new commit: the Error Handling table in review.md has two rows merged onto a single line (the "Catalog call returns empty" and "Recommendation list 401/403" rows are separated by `||` instead of a newline). This breaks table rendering.

plugin/skills/azure-advisor/review/review.md
sharadhisrikanth28: Fixed - I've split them onto separate lines so both render as proper table rows.

plugin/skills/azure-advisor/references/resource-discovery.md
```diff
@@ -0,0 +1,61 @@
...
+1. **Resource group(s)** — the strongest scope signal:
+   - `azure.yaml` → `resourceGroup:` / `resourceGroupName:` key
+   - `.azure/*/config.json` → `resourceGroup` field
+   - `*.bicepparam` / `*.parameters.json` → a `resourceGroup` / `resourceGroupName` parameter
+   - `.env*` files → read **only** the `AZURE_RESOURCE_GROUP` / `AZURE_RESOURCEGROUP` line; never load, echo, or summarize other `.env*` contents (they routinely hold secrets)
+   - `main.tf` / `*.tf` → `azurerm_resource_group` `name =` value (or a `resource_group_name` local/variable default)
+2. **Resource type(s)** — derive from the IaC resource declarations:
+   - Bicep `resource x 'Microsoft.<Provider>/<type>@...'` → `Microsoft.<Provider>/<type>`
+   - ARM `"type": "Microsoft.<Provider>/<type>"`
+   - Terraform `resource "azurerm_<kind>"` → map to its `Microsoft.*` provider type
```
jongio, line 30: Nit: Bicep and ARM patterns above are literal string extractions (the provider type is right there in the source), but the Terraform mapping requires the agent to *derive* the ARM type from `azurerm_<kind>`. For common resources this is fine, but for uncommon ones the mapping can be non-obvious (e.g., `azurerm_linux_function_app` -> `Microsoft.Web/sites`). Consider adding a fallback note: if the ARM type mapping is uncertain, prefer the resource group filter over risking a wrong type filter.

jongio (commented):
Incremental review of 106786a (resource-scoping commit). The table formatting defect from my last review is fixed, and the new resource-discovery.md is well-structured with good security awareness around .env file handling.

One minor nit on the new file (non-blocking).

plugin/skills/azure-advisor/references/resource-discovery.md
sharadhisrikanth28: Good catch — added a fallback note.

jongio (commented):
Incremental review of 59e0776 (Terraform type-mapping fallback note).

The distinction between literal extraction (Bicep/ARM) and derived mapping (Terraform) is the right framing. The `azurerm_linux_function_app` example illustrates why guessing is dangerous, and the fallback-to-resource-group-filter guidance prevents silent data loss from wrong type filters.

No new issues. My prior changes-requested items (negative prompts, IaC typos, README duplicates, hard-coded tool names, table formatting) are all resolved across the earlier commits.

plugin/skills/azure-advisor/README.md
```diff
@@ -0,0 +1,53 @@
+# Azure Advisor Skill — Contributor Guide
```
JasonYeMSFT: All the files under the skill directory will be shipped as a part of the skill. As of today, we don't have a pattern for keeping per-skill contributor only documentations. For now, I suggest removing this file and only keep files that are intended for LLM consumption.

plugin/skills/azure-advisor/SKILL.md
```diff
@@ -0,0 +1,67 @@
+---
+name: azure-advisor
+description: "Azure Advisor reviews resources and provides recommendations using advisor_* MCP tools. WHEN: \"run an advisor review\", \"check my Azure advisor recommendations\", \"summarize advisor findings\", \"what does Advisor say about my subscription\", \"give me an advisor health check\", \"audit my Azure resources with Advisor\". USE FOR: read-only sweeps with catalog, recommendations, and IaC fixes. DO NOT USE FOR: changing resources, billing analysis (use azure-cost), or non-Advisor troubleshooting (use azure-diagnostics)."
```
JasonYeMSFT: You should tell the model which exact tool to use. If it's Azure MCP, mention that the tool is from Azure MCP.

plugin/skills/azure-advisor/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+## Capabilities
+
+Route the user's request to the matching capability. **Use these instead of the main
+skill when they match the task:**
+
+| Capability | When to Use | Reference |
+|-----------|-------------|-----------|
+| **review** | Run a holistic, read-only Advisor sweep across one subscription — or **all** subscriptions classified by environment (dev/staging/prod) — probing the catalog, pulling active recommendations, aggregating by category/impact, spotlighting high-impact items, and proposing IaC fix snippets. | [review](review/review.md) |
+
+### Roadmap (not yet implemented)
```
JasonYeMSFT: What does the "(not yet implemented)" mean?

JasonYeMSFT:
We require all skills to have integration tests implemented as vally eval suites to express the scenarios they cover. Please read the tests/Readme.md file and follow the documentation + existing vally suite examples to implemented integration tests for your skill. Make sure to run your integration tests locally and refine the skill until they pass with meaningful results.

tests/azure-advisor/__snapshots__/triggers.test.ts.snap
```diff
@@ -0,0 +1,87 @@
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
```
JasonYeMSFT: We no longer require JavaScript based tests. See tests/Readme.md for the details.

sharadhisrikanth28: Migrated to a Vally eval suite (eval.yaml) covering the review/recommendation/audit scenarios, and removed the legacy tests/azure-advisor/ Jest tests.

plugin/skills/azure-advisor/SKILL.md
sharadhisrikanth28: "(not yet implemented)" flags planned capabilities that aren't built yet — today only review ships. The list is a roadmap so future capabilities reuse the same shared references and get added as sibling folders. Happy to drop the roadmap section entirely if you'd prefer the skill only document what's live.

plugin/skills/azure-advisor/SKILL.md
sharadhisrikanth28: Done — description now specifies the advisor_* tools come from the Azure MCP server.

plugin/skills/azure-advisor/README.md
sharadhisrikanth28: Done — deleted `README.md` so the skill ships only LLM-facing files. Contributor docs live outside the skill directory now.

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,154 @@
...
+      cost: llm
+      area: response-quality
+    graders:
+      - type: output-matches
+        config:
+          pattern: "(?i)recommendation|advisor"
+      - type: completed
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
```
jongio, line 209: The old `triggers.test.ts` had 9 `shouldNotTriggerPrompts` covering adjacent-service boundaries (cost analysis, diagnostics, Key Vault, App Service, RBAC). This eval has zero negative stimuli.

The `skill-invocation` grader supports a `disallowed` config (the azure-compute eval uses it for boundary assertions). Consider adding 2-3 boundary stimuli, for example:

```yaml
- name: "Cost query does not route to advisor"
  prompt: "Analyze my Azure subscription costs and spending trends"
  tags:
    type: integration
    tier: smoke
    cost: llm
    area: routing
  graders:
    - type: skill-invocation
      config:
        disallowed:
          - azure-advisor
```

Without negative stimuli, there's no automated coverage to catch false-positive routing on the adjacent services called out in the skill's DO NOT USE FOR section (`azure-cost`, `azure-diagnostics`).

jongio (commented):
Incremental review of 3c79fb8 and f8ee26d (Vally eval migration + description trim).

Prior items are addressed: README.md removed per JasonYeMSFT's request, description now specifies Azure MCP tool sourcing, code fence language tags added to review.md, and the contributing callout removed since its target was deleted.

One gap in the new eval coverage: the old `triggers.test.ts` had 9 negative prompts for adjacent-service boundary testing. The new eval.yaml has zero negative stimuli. See inline comment.

evals/azure-advisor/eval.yaml
sharadhisrikanth28: Added 3 boundary stimuli — restores the false-positive routing coverage from the old `shouldNotTriggerPrompts`

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,211 @@
...
+        config:
+          required:
+            - azure-advisor
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  # ═══════════════════════════════════════════
+  # Boundary (negative) routing — adjacent services from
+  # the skill's DO NOT USE FOR section must NOT route to azure-advisor.
```
jongio: The old `triggers.test.ts` covered 5 adjacent-service categories with 9 negative prompts. This adds 3 covering cost, diagnostics, and RBAC. The explicit DO NOT USE FOR services are covered. Key Vault and App Service boundaries from the old set aren't ported; consider adding them if false-positive routing on those surfaces is a concern.

jongio (commented):
Incremental review of 6fadac7 (boundary stimuli).

evals/azure-advisor/eval.yaml
jongio: The old `triggers.test.ts` covered 5 adjacent-service categories with 9 negative prompts. This adds 3 covering cost, diagnostics, and RBAC. The explicit DO NOT USE FOR services are covered. Key Vault and App Service boundaries from the old set aren't ported; consider adding them if false-positive routing on those surfaces is a concern.

jongio (commented):
Incremental review of 6fadac7 (boundary stimuli).

evals/azure-advisor/eval.yaml
sharadhisrikanth28: Ported the remaining two — added Key Vault and App Service boundary stimuli. All 5 adjacent-service categories from the old `triggers.test.ts` are now covered.

jongio (commented):
Previous comments addressed. The two new boundary stimuli (Key Vault, App Service) bring the negative-prompt count to 5, satisfying the AGENTS.md minimum. Structure matches the existing boundary tests. Nothing new to flag.

CI: the eval-job failure is in azure-compute lint, not this PR.

jongio (commented):
Previous comments addressed. The two new boundary stimuli (Key Vault, App Service) bring the negative-prompt count to 5, satisfying the AGENTS.md minimum. Structure matches the existing boundary tests. Nothing new to flag.

CI: the eval-job failure is in azure-compute lint, not this PR.

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,245 @@
...
+  Integration evaluation for the azure-advisor skill. Verifies routing for Advisor
+  review, recommendation-check, health-check, and audit prompts via skill-invocation
+  rate (5 runs, 80% threshold), plus response-quality checks for Advisor recommendation
+  output.
+
+tags:
+  type: integration
+  skill: azure-advisor
+
+environment:
```
JasonYeMSFT: You should remove this environment. The custom executor loads all skills by default and uses the skill tag to tell which skill it is testing. I don't remember if vally has fixed the bug where the top level tag is not propagated to the executor. If not, you may need to add a skill tag in each stimuli.

plugin/skills/azure-advisor/SKILL.md
JasonYeMSFT: You should remove this section and only keep the content that are ready.

plugin/skills/azure-advisor/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+- **summarize** — focused recommendation summarization / aggregation views
+- **resource-analysis** — per-resource Advisor analysis and drill-down
+- **greenfield** — Advisor-informed guidance for new/empty subscriptions
+- **cost** — Advisor cost-category optimization (coordinates with `azure-cost`)
+- **reliability** — Advisor reliability-category reviews
+- **governance** — Advisor operational-excellence / governance reviews
+- **security** — Advisor security-category reviews
+- **performance** — Advisor performance-category reviews
+
+## Constraints
```
JasonYeMSFT: These constraints are repeated in the references.

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,245 @@
...
+      cost: llm
+      area: response-quality
+    graders:
+      - type: output-matches
+        config:
+          pattern: "(?i)recommendation|advisor"
+      - type: completed
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
```
JasonYeMSFT, line 209: Could you add at least two tests that validates the behavior up to when the agent calls the advisor tool?

I have some pending change that would be needed for vally to grade it. https://github.com/microsoft/GitHub-Copilot-for-Azure/pull/2726

JasonYeMSFT: By the way, if the user let the agent work for these scenarios until completion, how much time and tokens would it roughly take?

evals/azure-advisor/eval.yaml
sharadhisrikanth28: Removed environment (executor loads all skills and uses the skill tag) and added a skill tag per stimulus for the propagation bug.

plugin/skills/azure-advisor/SKILL.md
sharadhisrikanth28: Fixed

evals/azure-advisor/eval.yaml
sharadhisrikanth28: Added 2 tests that stop right when the agent calls an advisor_ tool.
On cost: the review is scoped down, it narrows to the resources the repo actually defines (falling back to the full subscription only if the repo defines none), and Advisor returns pre-computed recommendations, not raw resource data. So, each run works over a bounded list. The routing/behavior tests early-terminate at the skill/tool call (~a minute, a few thousand tokens)

jongio (approved):
Incremental review of 515a56f and b512a45 (2 commits since my last review).

Changes address JasonYeMSFT's feedback:

- Removed dead environment block (executor loads skills by tag)
- Added per-stimulus skill: azure-advisor tags (matches azure-cost convention)
- Added 2 tool-call trajectory tests validating the agent invokes an advisor_ tool
- Removed duplicated Constraints section from SKILL.md (already expressed in description and capability-routing reference)

Trajectory tests use earlyTerminate with tool-call-match to stop as soon as the agent calls an advisor tool, and a count:8 safety cap. Structure is clean, graders are correct (skill-invocation + tool-calls + output-not-matches). No new issues.

jongio (dismissed):
Incremental review of 515a56f..b512a45 (2 commits since my last review).

Changes look good:

- Dead nvironment block removed, per-stimulus skill: azure-advisor tags added (matches azure-compute convention).
- Tool-call trajectory tests validate the agent actually calls an dvisor_ MCP tool, not just that the skill is invoked. The arlyTerminate config with 	ool-call-count: 8 is a sensible cap.
- Duplicated Constraints section removed from SKILL.md. Those rules already live in 
eview/review.md with more detail.

All prior feedback addressed. Eval coverage now spans routing (positive + negative boundary), behavior (tool-call trajectory), and response quality. No new issues.

jongio (commented):
Incremental review of 515a56f..b512a45 (2 commits since my last review).

Changes look good:

- Dead `environment` block removed, per-stimulus `skill: azure-advisor` tags added (matches azure-compute convention).
- Tool-call trajectory tests validate the agent actually calls an `advisor_` MCP tool, not just that the skill is invoked. The `earlyTerminate` config with `tool-call-count: 8` is a sensible cap.
- Duplicated Constraints section removed from SKILL.md. Those rules already live in `review/review.md` with more detail.

All prior feedback addressed. Eval coverage now spans routing (positive + negative boundary), behavior (tool-call trajectory), and response quality. No new issues.

(Reposted to fix formatting from previous comment where backticks were corrupted.)

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,305 @@
...
+        config:
+          disallowed:
+            - azure-advisor
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  # ── boundary-diagnostics ──
+  - name: "Diagnostics query does not route to advisor"
+    prompt: "My App Service keeps returning 500 errors, help me troubleshoot it"
```
JasonYeMSFT: I would recommend removing this test. It assumes a non-existent app service instance. The agent is smart and often it attempts to get information of the resource before doing anything. The test cannot proceed because the test resource doesn't exist.

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,305 @@
...
+        config:
+          disallowed:
+            - azure-advisor
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  # ── boundary-rbac ──
+  - name: "RBAC query does not route to advisor"
+    prompt: "Grant my team Reader access to the production resource group"
```
JasonYeMSFT: I recommend removing all the negative tests.

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,305 @@
...
+  # ── check-recommendations-calls-advisor-tool ──
+  - name: "Recommendation check reaches an advisor tool call"
+    prompt: "Check my Azure Advisor recommendations"
+    tags:
+      type: integration
+      skill: azure-advisor
+      tier: full
+      cost: llm
+      area: behavior
+      earlyTerminate: '[{"type":"tool-call-match","toolPattern":"advisor_","argsPattern":".*"},{"type":"tool-call-count","count":8}]'
```
JasonYeMSFT: You need to use the tool-call-result early terminate condition. The tool-calls grader checks for tool results. too-call-match condition terminates run when the tool is called meaning they don't get their results so the vally grader will always fail.

JasonYeMSFT:
Please add a new line in https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/main/.github/CODEOWNERS to list owners of your skill. As we discussed, you should add Sachin Soni, Ravi Gupta, Koushik B, yourself and Rick Winter (our project lead) as the skill owners. Our project lead is added as the owner for all skills as a final backup.

JasonYeMSFT:
You also need to add your skill's name to this list so the CI test runner run your eval suites. https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/f88faee73883147ff80067d9a880149633edc4a3/.github/workflows/test-all-integration.yml#L181

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,209 @@
...
+      - type: skill-invocation
+        config:
+          required:
+            - azure-advisor
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  # ═══════════════════════════════════════════
+  # Tool-call trajectory — validate behavior up to the point the agent
```
jongio, line 135: The 5 boundary stimuli that validated the skill does not activate for adjacent services (cost, diagnostics, RBAC, Key Vault, App Service) are removed here. I confirmed these met the AGENTS.md minimum on 6/30.

The new trajectory tests validate a different axis (advisor_ tool gets called once routing succeeds). They don't cover the negative case (skill stays out of unrelated queries). azure-compute's eval uses `disallowed` graders for exactly this purpose.

Consider keeping at least 2-3 boundary stimuli (e.g., cost + diagnostics + Key Vault) alongside the trajectory tests. They're cheap (early-terminate on first tool call) and prevent routing drift as new skills get added.

jongio (commented):
Incremental review of 4 new commits since my 6/30 review.

The tool-call-result early-terminate condition in tag-helpers.ts is solid: correct mapping of execution_start to execution_complete via toolCallId, and the eval trajectory tests exercise it well.

One concern: the boundary/negative routing stimuli I confirmed on 6/30 (cost, diagnostics, RBAC, Key Vault, App Service) are all removed. The eval now only validates positive routing and tool-call behavior. Other established evals (e.g., azure-compute) include disallowed stimuli to prevent routing drift. See inline comment.

jongio (commented):
Incremental review of the 5 commits since my last pass (96efad2...2ee8b49).

**Merge conflict**: This PR is in a conflicting state. `tests/vally/tag-helpers.ts` on main already has the `tool-call-result` type definition (with an optional `argsPattern` field). This PR re-adds it with a different shape (no `argsPattern`, different JSDoc). Needs a rebase to reconcile.

**CODEOWNERS**: JasonYeMSFT's comment asked you to add "yourself" to the CODEOWNERS entry. The current line lists `@sachinsoni29 @ravigupta99 @koushik137 @RickWinter` but not `@sharadhisrikanth28`. Was that intentional, or should you be listed too?

**Boundary test removal**: The previous iteration had 5 negative routing stimuli (cost, diagnostics, rbac, keyvault, appservice) that I validated. They're now replaced with tool-call trajectory tests. The trajectory tests verify the happy path (an `advisor_*` tool IS called), but nothing currently verifies the skill stays quiet for adjacent services. Is the plan to rely solely on the routing stimuli's `skill-invocation required` checks for correctness, or will negative tests return in a follow-up?

**What looks good in these commits**:
- Per-stimulus `skill: azure-advisor` tags are the right pattern for CI filtering.
- CI registration in `VALLY_SKILLS` and `tests/skills.json` both done.
- SKILL.md cleanup (roadmap removal, constraint dedup) makes the shipped surface cleaner.
- The `tool-call-result` handler implementation correctly maps `execution_start` events by `toolCallId` before checking `execution_complete`, so result-based graders see the payload.

tests/vally/tag-helpers.ts
```diff
@@ -39,6 +39,15 @@ export type EarlyTerminateCondition = {
...
+  /**
+   * A regex pattern matching the tool name of a *completed* tool execution.
+   * Unlike `tool-call-match` (which fires when the tool is invoked, before its
+   * result is available), this fires only after the matching tool has returned
+   * a result, so result-based graders (e.g. `tool-calls`) still see the result.
+   */
+  toolPattern: string;
 };
 
 export type TakeScreenshotCondition = {
```
jongio, line 60: This type variant already exists on main (added in a prior merge) with a slightly different shape that includes `argsPattern?: string`. The conflict needs resolving during rebase. The main-branch version also has a briefer JSDoc. Easiest path: drop this addition entirely and just add the runtime handler (lines 124-144), since the type already covers your use case.

.github/CODEOWNERS
```diff
@@ -11,6 +11,7 @@
 /plugin/skills/ @tmeschter @RickWinter
 /plugin/skills/airunway-aks-setup/ @tmeschter @RickWinter
 /plugin/skills/appinsights-instrumentation/ @JasonYeMSFT @RickWinter
+/plugin/skills/azure-advisor/ @sachinsoni29 @ravigupta99 @koushik137 @RickWinter
```
jongio, line 14: Per JasonYeMSFT's comment on this PR: "you should add Sachin Soni, Ravi Gupta, Koushik B, yourself and Rick Winter." Your handle (`@sharadhisrikanth28`) isn't in this list. Add it if that was the intent.

evals/azure-advisor/eval.yaml
jongio, line 135: The tool-call trajectory tests are a solid addition for verifying end-to-end behavior. One question: the `tool-calls` grader uses `name: "azure-advisor"` as the required match. Since actual tool names look like `azure-mcp-advisor_recommendation_list`, does this grader do substring/regex matching on the tool name? If it expects an exact match, this won't fire. Worth confirming the grader's matching semantics.

jongio (commented):
Incremental review of the 5 commits since my last pass (96efad2 to 2ee8b49):

**CODEOWNERS + CI registration**: Both address @JasonYeMSFT's requests. The CODEOWNERS entry is alphabetically correct. The VALLY_SKILLS array addition looks right.

**eval.yaml restructuring**: Replacing negative routing stimuli with tool-call trajectory tests is a reasonable shift. The trigger tests in `tests/azure-advisor/` already cover negative routing; the vally eval is better spent validating behavioral correctness (does the skill actually call `advisor_*` tools). The `earlyTerminate` with `tool-call-result` ensures the grader sees the tool result before termination fires. The `tool-call-count` fallback of 8 is a sensible safety net.

**tool-call-result early terminate (tag-helpers.ts)**: The implementation maps `execution_complete` events back to tool names by scanning `execution_start` events for matching `toolCallId`. This is correct given that `execution_complete` only carries the call ID, not the tool name. The Map rebuild on each check is fine for bounded test event streams.

**SKILL.md cleanup**: The removed constraints (never hardcode, read-only only, never call non-advisor tools) are all covered by the description frontmatter and the Capability Routing / Subscription Discovery / Resource Scope Discovery references. The roadmap section was speculative and doesn't belong in a shipped skill definition.

No issues to flag.

evals/azure-advisor/eval.yaml
sharadhisrikanth28: These were removed at JasonYeMSFT's request, he noted we don't require negative routing stimuli for plugin skills, and that routing for adjacent services should live in those skills' own tests.

tests/vally/tag-helpers.ts
sharadhisrikanth28: Fixed. Dropped my duplicate type and kept main's version.

.github/CODEOWNERS
sharadhisrikanth28: Intentional — I'm an intern wrapping up in a couple days, so I left myself off. Sachin, Ravi, Koushik, and Rick will own it going forward, which is why they're the listed owners.

evals/azure-advisor/eval.yaml
sharadhisrikanth28: The grader matches the tool name as a pattern (substring), not exact, so `azure-advisor` matches fine. In the mode CI uses, the tool is actually just named `azure-advisor` and the `recommendation_list` part is an argument, so it fires.

.github/CODEOWNERS
jongio, line 14: My earlier comment noted JasonYeMSFT's request to add "yourself" (the PR author) to this CODEOWNERS line. The current list has `@sachinsoni29 @ravigupta99 @koushik137 @RickWinter` but not `@sharadhisrikanth28`. Was that intentional, or should your handle be added here too?

tests/skills.json
```diff
@@ -31,6 +32,6 @@
     "integrationTestSchedule": {
         "0 5 * * 2-6": "microsoft-foundry",
         "0 8 * * 2-6": "azure-deploy",
-        "0 12 * * 2-6": "airunway-aks-setup,appinsights-instrumentation,azure-ai,azure-aigateway,azure-cloud-migrate,azure-compliance,azure-compute,azure-cost,azure-diagnostics,azure-enterprise-infra-planner,azure-kubernetes,azure-kusto,azure-messaging,azure-prepare,azure-quotas,azure-rbac,azure-resource-lookup,azure-resource-visualizer,azure-storage,azure-upgrade,azure-validate,entra-agent-id,entra-app-registration,azure-reliability,python-appservice-deploy"
+        "0 12 * * 2-6": "airunway-aks-setup,appinsights-instrumentation,azure-advisor,azure-ai,azure-aigateway,azure-cloud-migrate,azure-compliance,azure-compute,azure-cost,azure-diagnostics,azure-enterprise-infra-planner,azure-kubernetes,azure-kusto,azure-messaging,azure-prepare,azure-quotas,azure-rbac,azure-resource-lookup,azure-resource-visualizer,azure-storage,azure-upgrade,azure-validate,entra-agent-id,entra-app-registration,azure-reliability,python-appservice-deploy"
     }
 }
```
jongio, line 37: Missing trailing newline. Most editors and `git diff` prefer files to end with `\n`.

jongio (commented):
Incremental review against ba2ab239 (8 new commits since my last pass on 96efad29).

Most of my earlier feedback has been addressed: CODEOWNERS populated, skill registered in VALLY_SKILLS and the integration batch, tool-call trajectory tests added with the correct namespace pattern, and the duplicated Constraints section removed from SKILL.md.

The removal of negative routing stimuli is fine. Checking other Vally eval suites (e.g. azure-cost, azure-diagnostics), none include negative-routing stimuli either. The `skill-invocation` grader on positive prompts is the established pattern here.

One remaining nit: `tests/skills.json` is missing a trailing newline (the diff shows `\ No newline at end of file`). Not blocking, but worth fixing to keep git diffs clean.

Skill structure is solid: capability-routing by description rather than hard-coded tool names, auto-discovery of subscriptions and resource scope from IaC, read-only constraints, and .env secret awareness.

.github/CODEOWNERS
JasonYeMSFT: This is fine.

tests/skills.json
JasonYeMSFT: The missing trailing new line is fine, especially since the existing content doesn't have it.

JasonYeMSFT (dismissed):

jongio (commented):
One non-blocking item still outstanding: `tests/skills.json` lacks a trailing newline (`\ No newline at end of file` in the diff).

tmeschter (commented):
Are we confident this is making a difference (e.g. leading to the "advisor" MCP tools being called appropriately)?

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,209 @@
...
+  Integration evaluation for the azure-advisor skill. Verifies routing for Advisor
+  review, recommendation-check, health-check, and audit prompts via skill-invocation
+  rate (5 runs, 80% threshold), plus response-quality checks for Advisor recommendation
+  output.
+
+tags:
+  type: integration
+  skill: azure-advisor
+
+config:
```
jongio: Every other eval file in this repo (26 of them) uses `defaults:` as the top-level configuration key, not `config:`. The per-stimulus override key is `config:`, but the file-level one is `defaults:`.

```suggestion
defaults:
```

Without this fix the eval runner may not pick up the intended run count, timeout, executor, or model for stimuli that don't override them individually.

jongio (commented):
Incremental review (new commits since my 6/30 review). The latest round addresses prior feedback well: CODEOWNERS added, CI registration done, trajectory tests added, negative routing tests removed per maintainer guidance, duplicated constraints dropped. One remaining config issue below.

evals/azure-advisor/eval.yaml
JasonYeMSFT: `config` is deprecated and replaced by `defaults`. Please update it and we'll review again.

sharadhisrikanth28: Updated.

sharadhisrikanth28:
> Are we confident this is making a difference (e.g. leading to the "advisor" MCP tools being called appropriately)?

@tmeschter 
Yes. I've locally tested the skill across a range of prompts (Advisor reviews, recommendation checks, health checks, audits) and it reliably routes to the skill and calls the right advisor MCP tools.

evals/azure-advisor/eval.yaml
```diff
@@ -0,0 +1,209 @@
...
+      - type: skill-invocation
+        config:
+          required:
+            - azure-advisor
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  # ═══════════════════════════════════════════
+  # Tool-call trajectory — validate behavior up to the point the agent
```
jongio, line 135: The 5 boundary/negative routing stimuli (cost, diagnostics, RBAC, Key Vault, App Service) are gone, replaced by 2 tool-call trajectory tests. The trajectory tests are more valuable since they validate the agent actually calls an `advisor_*` tool, not just that the skill was invoked.

That said, negative routing coverage ("cost prompts don't land here") is a different concern from positive trajectory ("advisor prompts do call the right tool"). Only 2 other evals in this repo use `disallowed`, so it's a rare pattern, and I suspect these were removed because they were flaky in CI.

Worth confirming: is negative routing coverage intentionally deferred, or considered unnecessary given the skill description's `DO NOT USE FOR` boundaries?

jongio (commented):
Incremental review of 7 commits since my last review (96efad29).

Changes since then:
- `defaults:` replaces the deprecated `config:` key in eval.yaml
- `skill: azure-advisor` tag added to every stimulus (needed for CI routing)
- Roadmap section and duplicated constraints removed from SKILL.md (they live in review.md)
- 5 negative routing tests replaced with 2 tool-call trajectory tests
- CODEOWNERS, CI workflow, and skills.json registration added per @JasonYeMSFT's feedback

CI is all green. The structural changes look correct.

One observation inline on the negative-to-trajectory test swap.

tmeschter:
> > Are we confident this is making a difference (e.g. leading to the "advisor" MCP tools being called appropriately)?
> 
> @tmeschter Yes. I've locally tested the skill across a range of prompts (Advisor reviews, recommendation checks, health checks, audits) and it reliably routes to the skill and calls the right advisor MCP tools.

@sharadhisrikanth28 Are some of those prompts the basis for the evaluations in this change?

Also note that you have a couple of conflicts to resolve in the PR.

tmeschter:
@sharadhisrikanth28 My concern with this skill is that it seems to be primarily about helping the LM understand how to call the Azure Advisor MCP tools--information that should generally be covered by the tools' descriptions. 

Do you have test prompts where the LM failed to call the tools when the skill was _not_ present, and reliably called the skill when it was?

PR #1827, opened 2026-04-12
Add: New deploy-to-aks skill

gambtho:
## Description

Adds a new `deploy-to-aks` skill that guides developers through deploying applications to **existing** Azure Kubernetes Service (AKS) clusters. This is complementary to the existing `azure-kubernetes` skill — while `azure-kubernetes` covers cluster provisioning and Day-0 configuration, `deploy-to-aks` focuses on the application deployment workflow: framework detection, containerization, manifest generation, safeguards validation, and deployment.

### Scope: Quick Deploy

This skill is scoped to the **quick deploy** path — deploying to an existing AKS cluster with minimal questions. It detects the application framework, generates production-ready artifacts (Dockerfile, K8s manifests), validates against AKS Deployment Safeguards (DS001-DS013), and deploys with confirmation gates.

### What's included

**Core skill files:**
- `SKILL.md` — Main skill file with prerequisites, 5-step workflow, and references
- `phases/quick-deploy.md` — Detailed quick deploy workflow (detection → generation → safeguards → deploy → verify)
- `references/` — Detection tables, safeguards checklist, workload identity patterns, rollback guidance

**Knowledge packs** (9 frameworks):
- `spring-boot`, `express`, `nextjs`, `fastapi`, `django`, `nestjs`, `aspnet-core`, `go`, `flask`
- Each provides framework-specific Dockerfile patterns, health endpoints, database profiles, writable paths, and common AKS issues

**Templates** (20 files):
- Dockerfiles (Node, Python, Java, Go, .NET, Rust + matching `.dockerignore` per language)
- Kubernetes manifests (deployment, service, ingress, gateway, HPA, PDB, etc.)
- GitHub Actions CI/CD workflow
- Mermaid diagram templates

**Tests:**
- Trigger tests with 8 positive and 7 negative prompts
- Unit tests for metadata, description, and frontmatter validation
- Integration test scaffold
- Registered in `tests/skills.json`

### Token budget compliance

All files have been trimmed to fit within the repository's token limits:
- SKILL.md: 949 tokens (under 5000 hard limit)
- All reference files under 2000 token hard limit
- Knowledge packs trimmed to under 2000 token fallback limit

## Checklist

- [x] Tests pass locally (`cd tests && npm test`)
- [x] **If modifying skill descriptions:** verified routing correctness with integration tests (`npm run test:skills:integration -- <skill>`)
[...]

PR #2696, opened 2026-06-24
Add app-deploy workflow to the azure-kubernetes skill

gambtho:
## Description

Folds the `deploy-to-aks` capability from #1827 into the existing
`azure-kubernetes` skill as a nested sub-skill, `azure-kubernetes-app-deploy`,
directly addressing the review request to make deploy "a workflow within
azure-kubernetes" rather than a separate top-level skill.

Structure mirrors the existing `azure-kubernetes-automatic-readiness/` sub-skill:
no separate `version.json` (parent `pathFilters` covers the subtree), no new
`tests/skills.json` entry, and tests added as Vally stimuli in the parent's
`evals/azure-kubernetes/eval.yaml`.

The `app-deploy` name makes the cluster-vs-app boundary explicit:
- `azure-kubernetes` — provision/configure the cluster
- `azure-kubernetes-automatic-readiness` — assess migration to AKS Automatic
- `azure-kubernetes-app-deploy` — deploy an app to an existing cluster

## Review comments from #1827

- Carried-over content already includes the post-review fixes made on the
  original branch (configmap `data:`, deployment replicas/HPA note, dotnet/java
  Dockerfile headers, workload-identity imports and "Microsoft Entra ID"
  rename).
- Additionally corrected the go Dockerfile header comments (CMD -> ENTRYPOINT).
- Obsolete by the restructure: all Bicep comments (templates removed) and all
  `tests/deploy-to-aks/*` Jest comments (repo migrated to Vally). The
  `.gitignore` / `settings.local.json` hygiene items are resolved by omission.

## Testing

Six Vally stimuli (routing, routing boundary, output, workspace-fixture) added
to `evals/azure-kubernetes/eval.yaml`, grading on the `azure-kubernetes` skill
invocation. `checkCopilotCliCharBudget` passes (19315/20000; the nested
sub-skill description is not counted against the top-level budget).

Supersedes #1827.


plugins/azure-skills/skills/azure-kubernetes/azure-kubernetes-app-deploy/knowledge-packs/frameworks/django.md
```diff
@@ -0,0 +1,239 @@
+# Django Knowledge Pack
```
jongio, line 1: Broad comment for whole PR.  I don't think we should be adding all of these verbose md files to the azure-skills repo.  Are these also published in docs that we can reference instead?

jongio (commented):
The CI eval failure is pre-existing (`azure-compute` orphan files, not from this PR). The `azure-kubernetes` skill and eval both pass.

Two concerns beyond my earlier comment about content volume:

**Safeguards duplication**: `references/safeguards.md` (DS001-DS013) overlaps with the constraint spec already bundled in the sibling `azure-kubernetes-automatic-readiness` sub-skill. Consider referencing the existing constraint spec rather than maintaining a second copy of safeguard documentation within the same parent skill tree.

**Knowledge pack maintenance**: 9 framework-specific guides totaling ~2,060 lines (django, express, flask, fastapi, spring-boot, nextjs, nestjs, aspnet-core, go). Each contains Dockerfile patterns, health endpoint setup, database config, security context, and resource sizing. This framework knowledge changes with each major version. What's the plan for keeping these current? If they drift, the skill will generate outdated Dockerfiles or incorrect health endpoint configs.

The three Copilot review bot comments on this PR (frontmatter format, configmap `data: {}`, deploy.yml placeholder validation) appear to be incorrect. The description is an inline string already, `data: {}` is a valid empty map (not null), and `<image>` is explicitly excluded from the placeholder grep.

plugins/azure-skills/skills/azure-kubernetes/azure-kubernetes-app-deploy/knowledge-packs/frameworks/django.md
gambtho: that would be really convenient, but not that I'm aware of

gambtho:
> **Knowledge pack maintenance**: 9 framework-specific guides totaling ~2,060 lines (django, express, flask, fastapi, spring-boot, nextjs, nestjs, aspnet-core, go). Each contains Dockerfile patterns, health endpoint setup, database config, security context, and resource sizing. This framework knowledge changes with each major version. What's the plan for keeping these current? If they drift, the skill will generate outdated Dockerfiles or incorrect health endpoint configs.

Fair point -- let me see if i can genericize them a bit more to still be valuable without requiring constant updates

jongio (dismissed):
Formalizing the outstanding concern as a blocking request since you acknowledged it needs work but haven't pushed changes yet.

The framework knowledge packs (9 files, ~2,060 lines) contain version-pinned Dockerfile patterns, specific health endpoint implementations, database connection snippets, and security context recommendations. This content has a short shelf life. When Django 6, Spring Boot 4, or Next.js 16 ship, these files become actively misleading rather than helpful.

You mentioned you'd genericize them. Until that's done, this is blocking because deploying with stale framework advice (wrong base image tags, deprecated health check patterns, outdated security contexts) is worse than having no framework-specific advice at all.

Suggested direction: keep the detection logic and K8s manifest templates (those are framework-agnostic and stable), but reduce framework packs to a brief pointer table (framework, default port, health path, signal files) rather than full Dockerfile recipes and config examples. The Dockerfile templates in 	emplates/dockerfiles/ already cover the build patterns generically.

plugin/skills/azure-kubernetes/azure-kubernetes-app-deploy/phases/quick-deploy.md
```diff
@@ -0,0 +1,230 @@
...
+az aks get-credentials -g <resource_group> -n <aks_cluster_name> --overwrite-existing
+```
+
+### Verify Gateway API CRDs (only if Istio Gateway API detected)
+
+```bash
+kubectl get crd gateways.gateway.networking.k8s.io httproutes.gateway.networking.k8s.io 2>/dev/null
+```
+
+If missing: `kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.0.0/standard-install.yaml`
```
jongio: This pins Gateway API CRDs to v1.0.0, which shipped October 2023. The current stable channel is well past that.

Since this skill generates instructions at agent runtime (not static docs), the agent can look up the latest tag. Consider replacing this with a note like:

```
If missing, install the latest stable Gateway API CRDs:
kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/latest/download/standard-install.yaml
```

Or add a comment noting the version should be bumped when the skill is next touched.

jongio (commented):
Two items from the prior review cycle remain unresolved:

1. **Knowledge pack genericization still pending.** The 9 framework packs (~2,060 lines) pin specific base image versions (node:22, python:3.12, openjdk:21) and contain full Dockerfile recipes that will drift with major framework releases. @gambtho acknowledged this needs work in the conversation thread but no changes landed. Suggestion: add a `Last validated: YYYY-MM` line to each pack header so future maintainers know when to audit, and consider reducing the Dockerfile sections to a brief "key decisions" table (base image choice rationale, signal handling approach, writable paths) rather than full copy-paste recipes. The actual Dockerfile templates in `templates/dockerfiles/` already serve as the canonical artifact.

2. **Gateway API CRD version pinned to v1.0.0** (see inline comment on quick-deploy.md).

plugins/azure-skills/skills/azure-kubernetes/azure-kubernetes-app-deploy/references/base-images.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Per-language images
+
+| Language | Default source (template today) | Tag policy | Preferred Microsoft / Azure Linux image |
+|----------|---------------------------------|-----------|------------------------------------------|
+| .NET | `mcr.microsoft.com/dotnet/{sdk,aspnet}` | floating major (e.g. `9.0`) | already Microsoft ✓ |
+| Java | `eclipse-temurin:<maj>-{jdk,jre}-alpine` | floating major LTS | `mcr.microsoft.com/openjdk/jdk:<maj>-azurelinux` (also `-distroless`) |
+| Python | `python:<maj.min>-slim` | floating major.minor | `mcr.microsoft.com/azurelinux/base/python:<maj.min>` |
+| Node | `node:<maj>-alpine` | floating major LTS | `mcr.microsoft.com/azurelinux/base/nodejs:<maj>` |
+| Go | build `golang:<ver>-alpine`, runtime `gcr.io/distroless/static-debian12` | floating major.minor | `mcr.microsoft.com/oss/go/microsoft/golang:<ver>-azurelinux3.0` |
```
jongio, line 30: The build stages across all templates now use `<LATEST_STABLE_*>` placeholders, but Go and Rust runtime stages still reference `gcr.io/distroless/static-debian12` and `gcr.io/distroless/cc-debian12` literally. That's probably intentional since distroless images are identified by base OS variant rather than language version, and Google maintains these for the full Debian lifecycle. Worth adding a one-line note here (e.g., "Runtime images use the Debian-suffixed distroless name directly; they don't need a placeholder because they float their patch level within the Debian release") so someone doesn't try to "fix" this asymmetry later.

jongio (commented):
One non-blocking nit on the new base-images doc (inline below). The placeholder pattern and framework-pack reduction both look correct.

gambtho:
> One non-blocking nit on the new base-images doc (inline below). The placeholder pattern and framework-pack reduction both look correct.

Thoughts on this approach in general- should be less maintenance, but it's not as concise as what you suggested?    I think there is some value in being a bit more descriptive to help the llm make good choices, but the size is a tradeoff.

Happy to slim it down to what you suggested if you feel strongly @jongio

jongio (commented):
Re the verbosity question: the descriptive approach works for LLM consumption. Explicit rationale (like the distroless note) prevents the model from generating wrong `fixes` where an implicit convention would be ambiguous. The extra lines are justified when the reasoning isn't obvious from structure alone. Don't slim it down further.

gambtho:
@chandraneel  @jongio ok to approve this?

Tatsinnit (approved):
☕️👍 for the eval specs @JasonYeMSFT for eyes. Thank you!

chandraneel (approved):

PR #2705, opened 2026-06-26
feat: improve foundry skill in more real user scenarios.

anchenyi:
## Description

1. Add az/azd installation and login check
2. If users have not specified a Foundry project, ask them to choose one instead of silently selecting an existing one.
3. Make the toolbox boundary more explicit, and always require users to create the toolbox and write back the endpoints/names.
4. Generate an AGENTS.md-like file during scaffolding to ensure the Foundry skill is loaded for all relevant topics.

## Checklist


jongio (commented):
The az CLI addition and explicit project-selection gates are good improvements for user experience. The toolbox boundary clarification prevents silent provisioning surprises.

Two concerns with the auth-checking logic in both verify-environment scripts:

1. The regex patterns matching `error` and `failed` as bare alternatives are overly broad. If azd ever outputs something like "error handler ready" or "failed to find cached token, using fresh one" in a non-failure context, the script would incorrectly flag the user as not logged in.

2. In the PowerShell script, `Test-AzdAuthLoggedIn` returns `False` when output matches neither the positive nor the negative regex. Any unexpected azd output format (new version, locale differences) silently reports "not logged in," which blocks the workflow. Consider keeping the exit-code check as a fallback when text matching is inconclusive.

CI note: the `eval` check is failing.

plugin/skills/microsoft-foundry/foundry-agent/create/scripts/verify-environment.sh
```diff
@@ -47,10 +62,45 @@ for ext in azure.ai.agents azure.ai.projects; do
 done
 
 # 3. Auth status
-if azd auth login --check-status >/dev/null 2>&1; then
+AZD_AUTH_OUTPUT="$(azd auth login --check-status 2>&1 || true)"
+if printf '%s' "$AZD_AUTH_OUTPUT" | grep -Eiq '(not[[:space:]]+logged[[:space:]]+in|not[[:space:]]+authenticated|no[[:space:]]+account|login[[:space:]]+required|please[[:space:]]+run.*azd[[:space:]]+auth[[:space:]]+login|run.*azd[[:space:]]+auth[[:space:]]+login|expired|failed|error)'; then
+  note_action "Not logged in to azd. Ask the user to run 'azd auth login' (it opens a browser; never run it for them)."
+elif printf '%s' "$AZD_AUTH_OUTPUT" | grep -Eiq '(logged[[:space:]]+in|authenticated|already[[:space:]]+logged[[:space:]]+in)'; then
   note_ok "Logged in to azd."
```
jongio, line 69: The regex alternative `|failed|error` at the end of this pattern is very broad. A version of azd that prints something like `error_handler: none` or `failed_attempts: 0` in its status output would match and incorrectly trigger a not-logged-in action.

Consider narrowing to specific phrases (`authentication failed`, `token expired`) or falling back to the exit code when the text doesn't match a known positive pattern:

```suggestion
azdAuthExitCode=$?
if printf '%s' "$AZD_AUTH_OUTPUT" | grep -Eiq '(not[[:space:]]+logged[[:space:]]+in|not[[:space:]]+authenticated|login[[:space:]]+required|expired)'; then
  note_action "Not logged in to azd. Ask the user to run 'azd auth login' (it opens a browser; never run it for them)."
elif printf '%s' "$AZD_AUTH_OUTPUT" | grep -Eiq '(logged[[:space:]]+in|authenticated)'; then
  note_ok "Logged in to azd."
else
  # Unrecognized output -- fall back to exit code
  if [ "$azdAuthExitCode" -eq 0 ]; then
    note_ok "Logged in to azd."
  else
    note_action "Unable to verify azd auth status. Ask the user to run 'azd auth login' and re-run this script."
  fi
fi
```

plugin/skills/microsoft-foundry/foundry-agent/create/scripts/verify-environment.ps1
```diff
@@ -32,38 +32,101 @@ function Get-AzdJson {
...
+    if ($raw -match "(?i)(logged\s+in|authenticated|already\s+logged\s+in)") {
+        return $true
+    }
+
+    return $false
+}
+
+# 1. Required CLIs
+# Check PATH first, then probe common install locations (winget, MSI, chocolatey)
+$azdCommand = Add-CommandFallbackPath "azd" @(
```
jongio, line 82: Same concern here: matching `error` and `failed` as bare alternatives catches too many false positives. Additionally, the final `return $false` means any unrecognized output format silently blocks the user. Consider preserving the exit-code fallback:

```suggestion
function Test-AzdAuthLoggedIn {
    $raw = ""
    try {
        $null = & azd auth login --check-status 2>&1 | Tee-Object -Variable raw
    } catch {
        $raw = $_ | Out-String
    }
    $rawText = ($raw -join "`n")

    if ($rawText -match '(?i)(not\s+logged\s+in|not\s+authenticated|login\s+required|expired)') {
        return $false
    }
    if ($rawText -match '(?i)(logged\s+in|authenticated)') {
        return $true
    }
    # Unrecognized output -- fall back to exit code
    return ($LASTEXITCODE -eq 0)
}
```

plugin/skills/microsoft-foundry/foundry-agent/create/references/use-toolbox-in-hosted-agent.md
```diff
@@ -2,6 +2,8 @@
 
 Hosted agents access Foundry-managed tools through a **Toolbox MCP endpoint**. Unlike prompt agents that wire tools directly, hosted agents connect to a single MCP-compatible endpoint that exposes all configured tools. The platform handles credential injection, token refresh, and policy enforcement.
 
+> 🚦 **Creating the toolbox/connection:** write the **full agent-side code** that consumes the toolbox (not just an env var). The **only** thing you leave out is the toolbox endpoint/connection **value**: put a placeholder there, and have the user create the toolbox/connection in [Foundry Toolkit (VS Code)](https://code.visualstudio.com/docs/intelligentapps/tool-catalog) or the [Foundry Portal](https://ai.azure.com/) and write the real value back. Create the toolbox/connection yourself only when the user explicitly asks you to (or supplies the real values).
```
jongio: This exact callout is duplicated verbatim in 5 files (`tools.md`, `tool-work-iq.md`, `foundry-tool-catalog.md`, `create-hosted.md`, and this file). If the wording needs updating later, all 5 must be edited in sync. Consider extracting to a single shared reference file and linking to it, or defining it once in `create-hosted.md` Step 6 and referencing that anchor from the others.

plugin/skills/microsoft-foundry/foundry-agent/create/scripts/verify-environment.sh
anchenyi: Fixed. Thanks!

plugin/skills/microsoft-foundry/foundry-agent/create/scripts/verify-environment.ps1
anchenyi: Fixed. Thanks!

plugin/skills/microsoft-foundry/foundry-agent/create/references/use-toolbox-in-hosted-agent.md
anchenyi: Fixed. Thanks!

jongio (commented):
Incremental review (3 commits since my last pass). Both auth-checking fixes look correct: the overly broad `|failed|error` alternatives are removed, and unrecognized output now falls back to exit code instead of unconditionally returning false. The toolbox boundary centralization eliminates the duplication across 4 reference files.

One optional structural note below. The `eval` CI check is still failing (pre-existing).

plugin/skills/microsoft-foundry/foundry-agent/create/create-hosted.md
```diff
@@ -196,13 +202,27 @@ azd ai agent init --no-prompt \
...
 
 Read and follow [local-run](references/local-run.md). Complete one representative local invocation before deploying.
 
-### Step 6 -- Add tools (optional)
+### Step 7 -- Add tools (optional)
+
+Tools attach through **toolboxes** -- bundled MCP-compatible endpoints.
 
-Tools attach through **toolboxes** -- bundled MCP-compatible endpoints. Flow:
+### Toolbox creation boundary
```
jongio: Nit: this heading is at `###` (same level as the numbered step headings), which makes it look like a peer step rather than a subsection of Step 7. Consider `####` to clarify the hierarchy, or keep it if you want equal ToC visibility for discoverability by the reference file links.

anchenyi: Fixed. Thanks!

XiaofuHuang (approved):

PR #2719, opened 2026-06-29
fix: Get hooks working in Copilot again

tmeschter:
The different coding agents (VS Code, Copilot, Claude etc.) have varying levels of support for the [Open Plugins spec](https://open-plugins.com/) and varying fallback mechanisms to try and support plugins designed for other coding agents. This makes for a bit of a mess when it comes to supporting hooks.

For one thing, all the coding agents (or at least VS Code, Copilot CLI, and Claude) agree that hooks/hooks.json is a default location to find hooks. However, they do not agree on the format of that file. Further, they do not agree on a mechanism for excluding that file from the set of hooks files they read.

VS Code attempts to solve this by trying to determine if a plugin is following the Open Plugin spec, the Copilot spec, or the Claude spec, and varying its behavior accordingly. For example, for an "open" plugin it defines the `PLUGIN_ROOT` variable; for Claude `CLAUDE_PLUGIN_ROOT`, and for Copilot plugins... neither. However, Copilot itself defines both `PLUGIN_ROOT` _and_ `CLAUDE_PLUGIN_ROOT` at all times. So a "Copilot" plugin may not work in VS Code, even though it is explicitly trying to make it work the same way.

Here's where we currently are regarding hooks:
- In VS Code they work. It picks up the .plugin/plugin.json file, causing it to treat this as an "open" plugin. It honors using hooks/hooks.json as the exclusive source of hook.
- In Copilot hooks don't work at all. It, too, picks up .plugin/plugin.json but it doesn't understand the structure of the `hooks` section at all (it doesn't support the `exclusive` annotation) and so loads no hooks.
- Claude sort of works. It will pick up .claude-plugin/plugin.json, and since that does not specify a hooks file it will use hooks/hooks.json as the default. However, spaces anywhere in the expanded `${CLAUDE_PLUGIN_ROOT}` will break the call to track-telemetry.sh.

Here we make changes so that VS Code and Copilot will continue to use .plugin/plugin.json, but we remove the `hooks` section entirely. They will both fall back to the default of hooks/hooks.json. VS Code will use the `windows`, `osx` and `linux` commands, while Copilot will use `bash` and `powershell` instead.

Claude is pointed to hooks/claude-hooks.json which is in the format it expects and adds in the necessary double-quotes to prevent whitespace issues. Note that it will still try to load hooks/hooks.json as well and and raise an error about it not being in the right format; this seems to be harmless.

Also note that Claude provides no good way to automatically varying behavior between operating systems, so the telemetry reporting effectively only works on system where "bash" is installed (effectively every Mac and Linux system, but only some Windows systems).

Experimentally this allows VS Code and Copilot to report telemetry on Windows systems. I am unable to test Claude fully but it does "see" the hook in claude-hooks.json.

## Checklist


RickWinter (commented):
Copilot looks to be using claude schema.

RickWinter (commented):
README is stale and needs updating with this PR

plugin/hooks/claude-hooks.json
```diff
@@ -0,0 +1,14 @@
...
+        "hooks": [
+          {
+            "type": "command",
+            "command": "\"${CLAUDE_PLUGIN_ROOT}/hooks/scripts/track-telemetry.sh\""
+          }
+        ]
+      }
+    ]
+  }
+}
```
jongio: Nit: missing trailing newline at end of file.

jongio (commented):
The approach here is well-reasoned and the PR description does a great job explaining the compatibility matrix.

Two items:

1. The hooks compatibility table in README.md (line 137-138) will be stale after this lands. Copilot CLI row still references `plugin/hooks/copilot-hooks.json` and Claude Code row says `hooks/hooks.json (no custom hooks)`. Both are now wrong. Worth updating here or in a quick follow-up.

2. Nit: `claude-hooks.json` is missing a trailing newline.

tmeschter:
> Copilot looks to be using claude schema.

@RickWinter In what sense?

plugin/hooks/claude-hooks.json
tmeschter: Fixed.

jongio (commented):
Verified the full changeset. The path quoting asymmetry between hooks.json (unquoted ``) and claude-hooks.json (quoted ``) is correct: Copilot and VS Code resolve paths internally via structured keys, while Claude passes the `command` value to a shell as-is, requiring explicit quotes and the `bash` prefix. No blocking issues.

JasonYeMSFT (approved):

PR #2720, opened 2026-06-29
eval: track turn count and support maxTurns abort in agent runner

JasonYeMSFT:
## Summary

Add `turnCount` to `AgentMetadata` to track the number of assistant turns started (via `assistant.turn_start` events), and an optional `maxTurns` field to `AgentRunConfig`. When the turn count exceeds `maxTurns`, the session is aborted and a warning comment is appended to `agentMetadata.testComments` so the markdown report reflects the reason.

## Changes

- **`AgentMetadata.turnCount`** — new field initialized to `0`; incremented on each `assistant.turn_start` event.
- **`AgentRunConfig.maxTurns`** — optional field; when set, the run aborts once `turnCount > maxTurns`.
- Abort path pushes a ⚠️ test comment: `Run aborted: turn count (N) exceeded maxTurns (M).`

### DIFF
--- a/tests/utils/__tests__/agent-runner.test.ts
+++ b/tests/utils/__tests__/agent-runner.test.ts
@@ -10,11 +10,11 @@ describe("agent-runner", () => {
     test("isComplete flag prevents event processing after completion", () => {
       // This test verifies the fix for the issue where console.log
       // continues to fire after the test completes/times out.
-      
+
       // Create a mock scenario
       let isComplete = false;
       const events: string[] = [];
-      
+
       const mockEventHandler = (eventType: string) => {
         // Stop processing events if already complete
         if (isComplete) {
@@ -22,50 +22,50 @@ describe("agent-runner", () => {
         }
         events.push(eventType);
       };
-      
+
       // Simulate normal event flow
       mockEventHandler("session.start");
       mockEventHandler("assistant.message");
       mockEventHandler("tool.execution_start");
-      
+
       // Simulate completion
       isComplete = true;
-      
+
       // These events should not be processed
       mockEventHandler("tool.execution_complete");
       mockEventHandler("assistant.message_delta");
-      
+
       // Verify only pre-completion events were recorded
       expect(events).toEqual([
         "session.start",
-        "assistant.message", 
+        "assistant.message",
         "tool.execution_start"
       ]);
       expect(events).not.toContain("tool.execution_complete");
       expect(events).not.toContain("assistant.message_delta");
     });
-    
+
     test("isComplete flag is set on session.idle", () => {
       let isComplete = false;
       const events: string[] = [];
-      
+
       const mockEventHandler = (eventType: string) => {
         if (isComplete) {
           return;
         }
         events.push(eventType);
-        
+
         // Simulate the fix: set isComplete when session.idle is received
         if (eventType === "session.idle") {
           isComplete = true;
         }
       };
-      
+
       mockEventHandler("session.start");
       mockEventHandler("assistant.message");
       mockEventHandler("session.idle");
       mockEventHandler("tool.execution_complete"); // Should not be processed
-      
+
       expect(events).toEqual([
         "session.start",
         "assistant.message",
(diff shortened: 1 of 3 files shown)


tests/utils/agent-runner.ts
```diff
@@ -901,6 +914,19 @@ export function useAgentRunner(agentRunnerConfig: AgentRunnerConfig) {
 
           agentMetadata.events.push(event);
 
+          if (event.type === "assistant.turn_start") {
+            agentMetadata.turnCount++;
+            if (runConfig.maxTurns !== undefined && agentMetadata.turnCount > runConfig.maxTurns) {
+              agentMetadata.testComments.push(
+                `⚠️ Run aborted: turn count (${agentMetadata.turnCount}) exceeded maxTurns (${runConfig.maxTurns}).`
+              );
```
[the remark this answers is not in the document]
JasonYeMSFT: Let's not have test for this because the logic is simple enough.

tests/utils/agent-runner.ts
```diff
@@ -901,6 +914,19 @@ export function useAgentRunner(agentRunnerConfig: AgentRunnerConfig) {
...
+              isComplete = true;
+              resolve();
+              void session.abort();
+              return;
+            }
+          }
```
jongio, line 930: The `shouldEarlyTerminate` check just below uses the same abort pattern and has the same relationship with the follow-up loop. If the post-abort follow-up concern (raised by the bot) needs addressing, consider guarding the follow-up loop itself rather than patching each abort site individually. Something like:

```typescript
for (const followUpPrompt of runConfig.followUp ?? []) {
  if (isComplete) break;
  isComplete = false;
  await session.sendAndWait({ prompt: followUpPrompt }, FOLLOW_UP_TIMEOUT);
}
```

That would cover both `maxTurns` and `shouldEarlyTerminate` abort cases with a single guard. Not blocking for this PR though - the behavior is consistent with the existing pattern.

jongio (commented):
Clean, focused change. The `turnCount` tracking and `maxTurns` abort path correctly mirror the existing `shouldEarlyTerminate` pattern (same `isComplete = true` / `resolve()` / `session.abort()` sequence). Adding the warning to `testComments` is a nice improvement over the bare abort - it makes run reports self-explanatory.

The Copilot bot raised a valid point about follow-ups running after abort. Worth noting: this is not new to this PR. The `shouldEarlyTerminate` path (immediately below the new block) has the same behavior. If it matters, a single guard on the follow-up loop (e.g., checking an `aborted` flag before entering) would fix both cases in one shot, but that is a separate concern.

One minor note: `turnCount` is non-optional on `AgentMetadata`, so any code constructing that interface externally now must include it. CI passing confirms all sites are covered.

tests/utils/__tests__/agent-runner.test.ts
```diff
@@ -103,4 +103,123 @@
...
+      eventTypes: string[],
+      maxTurns: number | undefined
+    ): { metadata: MockAgentMetadata; isComplete: boolean; aborted: boolean; maxTurnsExceeded: boolean } {
+      let isComplete = false;
+      let aborted = false;
+      let maxTurnsExceeded = false;
+      const metadata: MockAgentMetadata = { events: [], testComments: [], turnCount: 0 };
+
+      for (const type of eventTypes) {
+        if (isComplete) break;
```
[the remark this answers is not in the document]
JasonYeMSFT: These changes are removed.

tests/utils/agent-runner.ts
JasonYeMSFT: Good idea.

tests/utils/agent-runner.ts
```diff
@@ -915,7 +941,9 @@ export function useAgentRunner(agentRunnerConfig: AgentRunnerConfig) {
 
       // Send follow-up prompts before aggregating stats so tool/skill/token
       // counts include events emitted during follow-up turns.
-      for (const followUpPrompt of runConfig.followUp ?? []) {
+      // Skip follow-ups when the run was aborted due to maxTurns being exceeded.
+      for (const followUpPrompt of (runConfig.followUp ?? [])) {
+        if (isComplete) break;
```
jongio: This always evaluates to `true` here. The `done` promise (line 940) only resolves after one of the three paths sets `isComplete = true` and calls `resolve()`. By the time execution reaches this loop, `isComplete` is guaranteed `true`, so the break fires on the first iteration for all runs (not just aborted ones).

The previous commit (56c78cc) used a dedicated `maxTurnsExceeded` flag, which correctly scoped the guard. Consider restoring that approach, or use a general `let aborted = false;` flag set only in the maxTurns and earlyTerminate paths.

jongio (commented):
The follow-up guard addresses the concern from my earlier review, but the `isComplete` check is too broad. All three completion paths (session.idle at line 910, maxTurns at line 923, shouldEarlyTerminate at line 931) set `isComplete = true` before calling `resolve()`. Since `await done` only resolves once `resolve()` fires, `isComplete` is guaranteed to be `true` when the follow-up loop is reached on line 945. The loop breaks on the first iteration for every run, not just aborted ones. Follow-ups never execute.

A dedicated flag (e.g., `let aborted = false;` set only in the maxTurns and earlyTerminate paths) would correctly distinguish "run was aborted" from "run completed normally". Then check `if (aborted) break;` instead.

tests/utils/agent-runner.ts
JasonYeMSFT: Fixed.

tests/utils/agent-runner.ts
```diff
@@ -915,7 +944,9 @@ export function useAgentRunner(agentRunnerConfig: AgentRunnerConfig) {
 
       // Send follow-up prompts before aggregating stats so tool/skill/token
       // counts include events emitted during follow-up turns.
-      for (const followUpPrompt of runConfig.followUp ?? []) {
+      // Skip follow-ups when the run was aborted due to maxTurns being exceeded.
```
jongio: This says `due to maxTurns being exceeded` but `isAborted` is also set in the `shouldEarlyTerminate` path below. Should be `// Skip follow-ups when the run was aborted.` to cover both triggers.

jongio (commented):
One minor accuracy issue in the new comment.

tests/utils/agent-runner.ts
JasonYeMSFT: Fixed.

tmeschter (approved):

PR #2724, opened 2026-06-30
chore: limit vally worker count

JasonYeMSFT:
## Description

Running 5 (default) concurrent vally workers with resource intensive tests is a little too much for local runs and can significantly slow down the machine. Limit it to 2 so people can better work on other stuff while running vally tests locally.

## Checklist

### DIFF
--- a/tests/run-vally-test.ts
+++ b/tests/run-vally-test.ts
@@ -311,6 +311,7 @@ async function main(): Promise<void> {
   const forwardedArgs = [...options.forwardedArgs];
 
   // default options
+  forwardedArgs.splice(0, 0, "--workers", "2");
   forwardedArgs.splice(0, 0, "--output-dir", "./results");
   forwardedArgs.splice(0, 0, "--executor-plugin", path.join(__dirname, "vally", "vally-executor.ts"));
   forwardedArgs.splice(0, 0, "--grader-plugin", path.join(__dirname, "vally", "vally-graders.ts"));
--- a/tests/run-vally-test.ts
+++ b/tests/run-vally-test.ts
@@ -311,7 +311,10 @@ async function main(): Promise<void> {
   const forwardedArgs = [...options.forwardedArgs];
 
   // default options
-  forwardedArgs.splice(0, 0, "--workers", "2");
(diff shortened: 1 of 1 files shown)


tests/run-vally-test.ts
```diff
@@ -311,6 +311,7 @@ async function main(): Promise<void> {
   const forwardedArgs = [...options.forwardedArgs];
 
   // default options
+  forwardedArgs.splice(0, 0, "--workers", "2");
```
jongio: Per the PR description, this is intended for local runs. It'll also cap CI concurrency (where you'd typically want more parallelism since the machine isn't shared).

Consider guarding it:

```suggestion
  if (!options.forwardedArgs.some(arg => arg.startsWith("--workers")) && !process.env.CI) {
    forwardedArgs.splice(0, 0, "--workers", "2");
  }
```

jongio (commented):
One suggestion: scope the worker limit to local runs only (currently affects CI too).

tmeschter (dismissed):

tests/run-vally-test.ts
```diff
@@ -311,6 +311,7 @@ async function main(): Promise<void> {
...
   forwardedArgs.splice(0, 0, "--output-dir", "./results");
```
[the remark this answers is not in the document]
JasonYeMSFT: I took the approach where the default is set as long as there is no higher level override. Our GitHub Actions runner already overrides the worker option. If it needs more parallellism we will update the value it passes.

tests/run-vally-test.ts
JasonYeMSFT: Addressed. See my response to the Copilot generated comment.

tmeschter (approved):

jongio (commented):
Re-reviewed after the conditional-check commit (`fd3a84f`). The `hasWorkersArg` guard works correctly: `parseCliOptions` only consumes `--skill`, `--pass-rate`, and `--output-dir`, so `--workers` always lands in `forwardedArgs`. CI (`test-all-integration.yml`) already passes `--workers 2` explicitly, preventing the default from triggering in Actions. Both `--workers N` and `--workers=N` forms are handled.

PR #2745, opened 2026-07-04
Escape OData filter input in msbench eval metrics

fallintoplace:
## Summary
- Escapes user-provided `benchmark` and `model` query values in `GET /api/msbench-eval-metrics` with the existing OData literal pattern (`'` -> `''`) to prevent broken queries from malformed values.
- Sanitizes filter construction in `dashboard/api/src/functions/getMsbenchEvalMetrics.ts`.
- Replaces user-visible raw exception text with generic error messages and logs detailed errors server-side via `context.error`.

## Testing
- Not run (not requested).

### DIFF
--- a/dashboard/api/src/functions/getMsbenchEvalMetrics.ts
+++ b/dashboard/api/src/functions/getMsbenchEvalMetrics.ts
@@ -24,7 +24,7 @@ function getEvalTableClient(): TableClient {
 }
 
 /** Escape a value for use inside an OData string literal (single quotes are doubled). */
-function odataLiteral(value: string): string {
+function escapeOdataQuotes(value: string): string {
     return value.replace(/'/g, "''");
 }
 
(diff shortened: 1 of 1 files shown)


dashboard/api/src/functions/getMsbenchEvalMetrics.ts
```diff
@@ -23,6 +23,11 @@ function getEvalTableClient(): TableClient {
     );
 }
 
+/** Escape a value for use inside an OData string literal (single quotes are doubled). */
+function odataLiteral(value: string): string {
```
JasonYeMSFT: I think we should rename this function to `escapeOdataQuotes` to make it more obvious on what it does.

JasonYeMSFT (changes requested):

JasonYeMSFT (approved):

PR #2746, opened 2026-07-04
Fix PR workflow skill path globs for nested skills

fallintoplace:
## Summary
- Fixed PR workflow skill file detection in `.github/workflows/pr.yml` to match nested SKILL files with recursive globs.
  - `plugin/skills/**/SKILL.md`
  - `.github/skills/**/SKILL.md`
- Updated frontmatter validation path mapping so changed `plugin/skills/.../SKILL.md` files are validated against their built `output/skills/.../SKILL.md` counterparts.
- Made the mapping loop newline-safe and quote-safe to avoid argument-splitting issues in shell processing.

## Testing
- Not run (not requested).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -263,7 +263,7 @@ jobs:
             if [[ -z "$file" ]]; then
               continue
             fi
-            if [[ "$file" == plugin/skills/*/SKILL.md ]]; then
+            if [[ "$file" == plugin/skills/* ]]; then
               mapped="output/skills/${file#plugin/skills/}"
               SKILL_FILES+=("../$mapped")
             else
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -211,6 +211,7 @@ jobs:
         id: changed-skills
         uses: tj-actions/changed-files@22103cc46bda19c2b464ffe86db46df6922fd323 # v47.0.5
         with:
+          separator: "\n"
           files: |
             plugin/skills/**/SKILL.md
             .github/skills/**/SKILL.md
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -212,8 +212,8 @@ jobs:
         uses: tj-actions/changed-files@22103cc46bda19c2b464ffe86db46df6922fd323 # v47.0.5
         with:
           files: |
-            plugin/skills/*/SKILL.md
-            .github/skills/*/SKILL.md
+            plugin/skills/**/SKILL.md
+            .github/skills/**/SKILL.md
 
       - name: Check all changed plugin skill files
         id: changed-plugin-skills
@@ -258,18 +258,21 @@ jobs:
 
           # Map changed plugin/skills paths to built output/skills paths so we
           # validate the version-stamped copies. .github/skills are unchanged.
-          SKILL_FILES=""
-          for file in $CHANGED_FILES; do
-            if echo "$file" | grep -q '^plugin/skills/'; then
-              mapped=$(echo "$file" | sed 's|^plugin/skills/|output/skills/|')
-              SKILL_FILES="$SKILL_FILES ../$mapped"
-            else
-              SKILL_FILES="$SKILL_FILES ../$file"
+          SKILL_FILES=()
+          while IFS= read -r file; do
+            if [[ -z "$file" ]]; then
+              continue
             fi
-          done
+            if [[ "$file" == plugin/skills/*/SKILL.md ]]; then
+              mapped="output/skills/${file#plugin/skills/}"
+              SKILL_FILES+=("../$mapped")
+            else
+              SKILL_FILES+=("../$file")
+            fi
+          done <<< "$CHANGED_FILES"
 
           # Run frontmatter spec validation
-          if OUTPUT=$(npm run frontmatter -- $SKILL_FILES 2>&1); then
+          if OUTPUT=$(npm run frontmatter -- "${SKILL_FILES[@]}" 2>&1); then
             echo "$OUTPUT"
             echo "✅ All skill frontmatter is valid" >> $GITHUB_STEP_SUMMARY
           else


.github/workflows/pr.yml
```diff
@@ -258,18 +258,21 @@ jobs:
...
-          for file in $CHANGED_FILES; do
-            if echo "$file" | grep -q '^plugin/skills/'; then
-              mapped=$(echo "$file" | sed 's|^plugin/skills/|output/skills/|')
-              SKILL_FILES="$SKILL_FILES ../$mapped"
+          SKILL_FILES=()
+          while IFS= read -r file; do
+            if [[ -z "$file" ]]; then
+              continue
+            fi
+            if [[ "$file" == plugin/skills/*/SKILL.md ]]; then
```
jongio: In bash `[[ ]]` pattern matching, `*` matches any character sequence including `/`, so this works for nested paths. However, it reads differently from the `**` glob in the `changed-files` filter above (where `**` explicitly means recursive).

Since `$CHANGED_FILES` is already filtered to SKILL.md files only, you could simplify to a prefix check:

```suggestion
            if [[ "$file" == plugin/skills/* ]]; then
```

This makes the intent clearer without relying on the (correct but non-obvious) behavior of `*` in `[[ ]]`.

jongio (commented):
One minor suggestion below about pattern clarity in the bash conditional.

.github/workflows/pr.yml
fallintoplace: Thank you. Let me have a look and push the change.

dashboard/sync/src/functions/syncMsbenchEvalMetrics.ts
```diff
@@ -106,12 +117,14 @@ async function runSync(context: InvocationContext, force = false): Promise<{ syn
...
-            const benchmark = report.instance_id || "unknown";
-            const model = report.model || "unknown";
+            const instanceId = toSafeString(report.instance_id);
+            const benchmark = toSafeString(report.benchmark ?? instanceId);
+            const model = toSafeString(report.model);
 
             await tableClient.upsertEntity({
                 partitionKey: date,
-                rowKey: `${benchmark}_${model}`,
+                rowKey: buildMsbenchRowKey(instanceId, benchmark, model),
```
jongio: Changing the row key format is a breaking change for existing table data. Old rows keyed as `benchmark_model` will persist alongside new rows keyed as a SHA-256 hash for the same logical entity. If the dashboard queries all rows for a given partition (date), it'll see duplicates.

Consider either: (a) a one-time migration that deletes old-format rows after the new code deploys, or (b) a query-side dedup that picks the hash-keyed row when both formats exist for the same entity.

dashboard/sync/src/functions/syncMsbenchEvalMetrics.ts
```diff
@@ -13,10 +14,20 @@ interface EvalReport {
     total_steps: number;
     model: string;
     instance_id: string;
+    benchmark?: string;
     resolved?: boolean;
     [key: string]: unknown;
 }
 
+function toSafeString(value: string | undefined): string {
```
jongio: If `instance_id` is empty/whitespace, `toSafeString` returns `"unknown"`. Then `benchmark` falls back to that same `"unknown"` value (via the `?? instanceId` path). If `model` is also empty, you end up with `sha256("unknown|unknown|unknown")` for every report with missing fields. Subsequent upserts silently overwrite each other.

If this data can realistically have missing instance IDs, consider using a random suffix or the blob path as a fallback identity instead of a fixed sentinel.

jongio (commented):
My prior suggestion (prefix-check pattern) was adopted in 43e1074. The pr.yml and eval.yml changes look solid.

Two concerns on the dashboard sync commit (8453df7):

**Scope**: This PR now bundles three separate concerns (workflow glob fix, eval.yml npm-ci switch, and dashboard row-key redesign). Consider splitting the dashboard change into its own PR so each can be reverted independently.

**Row key migration**: Changing the row key format from `benchmark_model` to a SHA-256 hash means existing table rows won't be overwritten by future syncs. If the sync reprocesses a date that was previously synced with the old key format, you'll get duplicate rows for the same logical entity. Worth adding a one-time migration or cleanup step, or at minimum documenting that old-format rows should be purged after deploy.

dashboard/sync/src/functions/syncMsbenchEvalMetrics.ts
fallintoplace: Quick question before I push this: do we want to implement read-side dedupe in this PR (safe, non-destructive), and then do legacy row cleanup in a follow-up migration after we validate behavior? I can do both, but this keeps rollout low-risk.

.github/workflows/pr.yml
```diff
@@ -258,18 +258,21 @@ jobs:
...
+            fi
+            if [[ "$file" == plugin/skills/* ]]; then
+              mapped="output/skills/${file#plugin/skills/}"
+              SKILL_FILES+=("../$mapped")
             else
-              SKILL_FILES="$SKILL_FILES ../$file"
+              SKILL_FILES+=("../$file")
             fi
-          done
+          done <<< "$CHANGED_FILES"
```
jongio, line 273: `tj-actions/changed-files` defaults to space-separated output since no `separator` input is set on the `changed-skills` step. `while read` splits on newlines, so if a PR touches multiple SKILL.md files, the entire space-separated list arrives as one "line" and the loop iterates once with all paths concatenated into `$file`.

Fix options:

1. Add `separator: "\n"` to the `changed-skills` step (cleanest):
```yaml
        with:
          separator: "\n"
          files: |
            plugin/skills/**/SKILL.md
            .github/skills/**/SKILL.md
```

2. Or convert inline before the loop:
```bash
          done < <(tr ' ' '\n' <<< "$CHANGED_FILES")
```

jongio (commented):
One issue: the `while read` loop expects newline-delimited input but `tj-actions/changed-files` defaults to space separation (no `separator` configured on the step). Single-file PRs work fine, but multiple changed SKILL.md files would all land on one iteration as a single concatenated string. See inline comment for fix options.

jongio (commented):
Verified the incremental commit (d8a753c): `separator: "\n"` correctly pairs with the `while IFS= read -r` / here-string pattern, so multi-file PRs will iterate one path per loop as intended.

No new issues. All prior feedback (recursive globs, prefix-check, separator mismatch) is addressed.

JasonYeMSFT:
@fallintoplace Only the frontmatter of top level SKILL.md matters since agents don't treat nested skill.md as skills. Some early contributors created nested SKILL.md files with frontmatter due to misunderstanding. I don't think we should expand the frontmatter validation to nested SKILL.md files.

tmeschter:
@JasonYeMSFT @fallintoplace

> I don't think we should expand the frontmatter validation to nested SKILL.md files.

Agreed. These don't define actual skills, and validating the frontmatter can only create noise.

PR #2747, opened 2026-07-05
Use stable hashed row keys for msbench eval sync

fallintoplace:
## Summary
- Replace unstable `rowKey = benchmark_model` construction with a deterministic key: `base64url(sha256(instance_id|benchmark|model))`.
- Persist `instance_id`, `benchmark`, and `model` as table properties for queryability and traceability.

## Notes
- This addresses invalid Azure Table row key characters (e.g. `/`, `\\`, `#`, `?`) and collision risks from concatenated keys.
- No query/filter behavior changes are included in this change.

### DIFF
--- a/dashboard/sync/src/functions/syncMsbenchEvalMetrics.ts
+++ b/dashboard/sync/src/functions/syncMsbenchEvalMetrics.ts
@@ -20,11 +20,12 @@ interface EvalReport {
 }
 
 function toSafeString(value: string | undefined): string {
-    return value?.trim() ? value : "unknown";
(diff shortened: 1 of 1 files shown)


JasonYeMSFT:
@fallintoplace Could you please elaborate on the problem of the existing code? Do you see it causing a bug or do you think the new code brings some benefits?

fallintoplace:
Thanks for asking. I noticed the current row key could potentially run into invalid-character or collision issues. After taking another look, the migration handling adds more complexity than expected, so I’ll close this for now and revisit it if we see a concrete case.

PR #2748, opened 2026-07-05
Use npm ci in eval workflow dependencies

fallintoplace:
## Summary
- Replace `npm install` with deterministic lockfile installs in the eval workflow.
- Use `npm ci --ignore-scripts` at repo root.
- Install `scripts` dependencies explicitly via `npm ci` so `npm run vally` still works.

## Safety rationale
- `npm ci` uses `package-lock.json` exactly and avoids lockfile drift in CI.
- Scripts are now installed in an explicit step, avoiding hidden `postinstall` behavior from root only.

### DIFF
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -35,7 +35,10 @@ jobs:
           node-version: "22"
 
       - name: Install dependencies
-        run: npm install
+        run: |
+          npm ci --ignore-scripts
+          cd scripts
+          npm ci
 
       - name: Install test dependencies
         working-directory: tests
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -35,10 +35,11 @@ jobs:
           node-version: "22"
 
       - name: Install dependencies
-        run: |
-          npm ci --ignore-scripts
-          cd scripts
-          npm ci
+        run: npm ci --ignore-scripts
+
+      - name: Install stimuli validation dependencies
+        working-directory: scripts
+        run: npm ci --ignore-scripts
 
       - name: Install test dependencies
         working-directory: tests


.github/workflows/eval.yml
```diff
@@ -35,7 +35,10 @@ jobs:
           node-version: "22"
 
       - name: Install dependencies
-        run: npm install
+        run: |
+          npm ci --ignore-scripts
+          cd scripts
+          npm ci
```
jongio: Consider adding --ignore-scripts here too for parity with the root install. The scripts/ dependencies (eslint, vitest, typescript, tsx, micromatch, etc.) are all pure JS/TS packages with no native postinstall hooks, so it should be safe. This would extend the security hardening to both install steps.

Not blocking since the scripts/ lockfile is version-controlled and the deps are well-known packages.

jongio (commented):
The approach here is correct. The root package.json has a `postinstall` hook (`cd scripts && npm install`) that implicitly installed scripts/ dependencies. Since `--ignore-scripts` skips that hook, the explicit `cd scripts; npm ci` step is the right compensation.

Both root and scripts/ have a `package-lock.json`, so `npm ci` will work in both locations.

One minor note on the scripts install below.

jongio (commented):
Incremental review (new commits since my last pass):

The scripts/ install now includes `--ignore-scripts`, which was the one suggestion from my prior review. Both install steps are consistent: lockfile-pinned (`npm ci`) with script suppression for CI safety.

Final state looks correct. No remaining concerns.

.github/workflows/eval.yml
```diff
@@ -35,7 +35,10 @@ jobs:
           node-version: "22"
 
       - name: Install dependencies
-        run: npm install
+        run: |
+          npm ci --ignore-scripts
+          cd scripts
```
JasonYeMSFT: Could you please remove the 
```
cd scripts
npm ci --ignore-scripts
```
part, since the old script doesn't need them?

JasonYeMSFT (changes requested):

.github/workflows/eval.yml
RickWinter: Please keep this install. The later `Validate stimuli metadata` step runs `cd scripts && npm run vally validate-stimulus`, and the scripts package provides its `tsx` runtime. Removing this after the root `npm ci --ignore-scripts` skips the root postinstall will leave that command unable to resolve its runtime and fail the eval workflow.

RickWinter (commented):
This switches the eval workflow to lockfile-pinned installs and explicitly restores the scripts dependencies skipped by the root install. The approach is the right shape. The current discussion proposes removing that required scripts install, which would break the metadata validation step. That must be resolved before merge.

.github/workflows/eval.yml
fallintoplace: I kept this as a separate step because the later metadata validation runs npm run vally from scripts/ and requires those dependencies. Root npm ci --ignore-scripts no longer installs them via postinstall.

JasonYeMSFT (approved):

PR #2769, opened 2026-07-07
fix: update issue triage field guidance

RickWinter:
## Summary
- update issue triage guidance to use the Issue Type field instead of type labels
- add Priority field triage guidance for Urgent, High, Medium, and Low
- keep skill and assign-to-copilot labels as label-based triage

## Validation
- git diff --check


jongio (commented):
The two automated review comments are incorrect:

1. Re: "no way for the agent to set those fields" - this PR adds set_issue_type and set_issue_field as dedicated tools for exactly this purpose. The agent doesn't use update_issue for type/priority; it uses these new tools with proper constraints (types restricted to Bug/Feature/Task, field restricted to Priority).

2. Re: "lock file still contains the old workflow description" - the lock file is updated in this PR. The description, tool list, safe-output configs, and metadata hashes are all regenerated.

RickWinter:
Quick clarification on the timeline. The two automated comments were accurate against the first commit (ebc50edb), which only touched the .md: at that point update_issue couldn't set those fields and the lock still had the old description. Both were resolved in b0dfde0a, which adds set_issue_type (Bug/Feature/Task) and set_issue_field (Priority) as dedicated tools and regenerates the lock. A fresh gh aw compile now produces no diff, so the stale-lock check passes. Net: valid flags, already fixed, no further action needed.

jongio (commented):
Reviewed the migration from type labels to Issue Type and Priority fields. The safe-outputs constraints correctly pair with the prompt guidance: `set_issue_type` restricted to Bug/Feature/Task, `set_issue_field` restricted to the Priority field only, both with max 1. Tool schemas include proper input sanitization (`sanitize: true`, `maxLength` limits) and the `requiresOneOf:field_name,field_node_id` custom validation ensures field identification.

The lock file regeneration is consistent with the .md changes. The backward-compatible label removal in step 6 correctly handles reopened issues that may carry old type labels from before this migration.

No blocking issues found.

JasonYeMSFT (approved):

PR #2771, opened 2026-07-07
feat: add Skills dashboard section with per-test metric sparklines

tmeschter:
## Summary

Adds a new **Skills** section to the dashboard: a master–detail page listing all plugin skills, with a detail view for the selected skill.

For each selected skill, the detail view shows:
- The skill **description** and its **description length** (characters).
- For every associated integration test, three small graphs — **total duration**, **token usage**, and **turns** — over the last **ten days** on the `main` branch.

## Changes

**New Skills page** (`dashboard/skills.html`, `dashboard/src/skills/`)
- `metrics.ts` — pure helpers (`lastNDates`, `buildDaySeries`, `groupByTest`). Multiple runs on the same day are averaged; days with no data render as gaps rather than zeros.
- `App.tsx` — master–detail UI. Fetches the plugin-skill list + descriptions from `/api/static`, and per-test rows from `/api/test-run-metrics?skill=<name>&branch=main`. Renders recharts sparklines. Supports `?skill=` deep-links.
- `main.tsx`, `skills.css` (reuses the shared design-token CSS variables).
- Added the **Skills** nav link to every page and registered `skills.html` in `vite.config.ts`.

**API: `getTokenUsage` → `getTestRunMetrics`**
- Renamed the function and its routes to `test-run-metrics` and `test-run-metrics/filters` to reflect the fuller set of data it now returns.
- Extended the returned rows with `durationMs` and `turns` in addition to token counts.
- Updated the existing Token Usage page caller, the API docs in `dashboard/Readme.md`, and the `apiUrl` example.

**Data plumbing** (`tests/scripts/upload-token-usage.ts`)
- Carries `totalApiDurationMs` → `durationMs` and `apiCallCount` → `turns` from `token-summary.jsonl` into the Azure Table.

**Frontmatter → dashboard**
- The frontmatter CLI (`scripts/src/frontmatter/cli.ts`) and dashboard collector (`scripts/src/dashboard/collectors/frontmatter.ts`) now surface `description`, `descriptionLength`, and `path`, so the Skills page can list plugin skills and show descriptions.

## Storage impact

Additive only. Newly uploaded Azure Table rows gain two properties (`durationMs`, `turns`). Azure Table Storage is schemaless, so there is **no migration** and existing rows are untouched — legacy rows without these fields render as gaps in the graphs. The storage account, table name, and `TOKEN_USAGE_TABLE_NAME` env var are intentionally left unchanged to avoid a data migration.

## Validation

- `dashboard`: `npm run build` ✅, `tsc --noEmit` clean for the new code ✅
- `dashboard/api`: `npm run build` ✅
- `scripts`: `npm test` — 463 tests pass ✅, `npm run typecheck` ✅
- `tests`: `npm run typecheck` ✅


tests/scripts/upload-token-usage.ts
```diff
@@ -61,6 +63,10 @@ interface AggregatedUsage {
     outputTokens: number;
     cacheReadTokens: number;
     cacheWriteTokens: number;
+    /** Total LLM API duration in ms, summed across the test's records. */
+    durationMs: number;
+    /** Number of LLM round-trips ("turns"), summed across the test's records. */
+    turns: number;
```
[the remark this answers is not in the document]
tmeschter: Fixed in bd362486: `durationMs`/`turns` are now optional on `AggregatedUsage`, so absence is preserved end-to-end instead of persisting as 0.

tests/scripts/upload-token-usage.ts
```diff
@@ -189,11 +195,15 @@ function aggregateByTest(records: TokenSummaryRecord[]): AggregatedUsage[] {
     const output = Number(r.outputTokens) || 0;
     const cacheRead = Number(r.cacheReadTokens) || 0;
     const cacheWrite = Number(r.cacheWriteTokens) || 0;
+    const durationMs = Number(r.totalApiDurationMs) || 0;
+    const turns = Number(r.apiCallCount) || 0;
```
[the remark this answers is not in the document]
tmeschter: Fixed in bd362486: replaced `Number(...) || 0` with `optionalMetric()`, which only yields a value when the source field is present and finite; otherwise it stays undefined.

tests/scripts/upload-token-usage.ts
```diff
@@ -207,6 +217,8 @@ function aggregateByTest(records: TokenSummaryRecord[]): AggregatedUsage[] {
         outputTokens: output,
         cacheReadTokens: cacheRead,
         cacheWriteTokens: cacheWrite,
+        durationMs,
+        turns,
```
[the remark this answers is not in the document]
tmeschter: Fixed in bd362486: aggregation now uses `addOptional()`, which keeps the metric undefined when no record carried it and only sums the present values.

tests/scripts/upload-token-usage.ts
```diff
@@ -291,6 +303,8 @@ async function main(): Promise<void> {
           cacheReadTokens: usage.cacheReadTokens,
           cacheWriteTokens: usage.cacheWriteTokens,
           totalTokens,
+          durationMs: usage.durationMs,
+          turns: usage.turns,
```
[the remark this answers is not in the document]
tmeschter: Fixed in bd362486: `durationMs`/`turns` are now spread into the entity only when present, so absent metrics are never written to the schemaless Table row.

dashboard/src/skills/App.tsx
```diff
@@ -0,0 +1,252 @@
...
+                    tick={{ fontSize: 10 }}
+                    interval="preserveStartEnd"
+                    minTickGap={16}
+                />
+                <YAxis hide domain={["auto", "auto"]} />
+                <Tooltip
+                    formatter={(value) => [format(Number(value)), ""] as [string, string]}
+                    labelFormatter={(l) => String(l)}
+                    contentStyle={{ fontSize: 12 }}
+                />
```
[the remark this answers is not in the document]
tmeschter: Fixed in bd362486: the tooltip formatter now returns a "No data" placeholder for null/undefined days instead of coercing them to 0.

dashboard/src/skills/App.tsx
```diff
@@ -0,0 +1,259 @@
...
+        const params = new URLSearchParams({ skill: selected, branch: "main" });
+        fetch(apiUrl(`/api/test-run-metrics?${params}`))
+            .then((res) => {
+                if (!res.ok) throw new Error(`API error: ${res.status}`);
+                return res.json() as Promise<MetricsRow[]>;
+            })
+            .then((data) => setRows(data))
+            .catch((err) => setRowsError(err.message))
+            .finally(() => setRowsLoading(false));
+    }, [selected]);
```
jongio, line 191: This `useEffect` fires a new fetch each time `selected` changes but never cancels the previous one. If the user switches from skill A to skill B while A's request is still in flight, A's late response calls `setRows(dataA)` and overwrites B's correct data.

Fix with an `AbortController`:

```typescript
useEffect(() => {
    if (!selected) return;
    const controller = new AbortController();
    setRowsLoading(true);
    setRowsError(null);
    setRows([]);
    const params = new URLSearchParams({ skill: selected, branch: "main" });
    fetch(apiUrl(`/api/test-run-metrics?${params}`), { signal: controller.signal })
        .then((res) => {
            if (!res.ok) throw new Error(`API error: ${res.status}`);
            return res.json() as Promise<MetricsRow[]>;
        })
        .then((data) => setRows(data))
        .catch((err) => {
            if (err.name !== "AbortError") setRowsError(err.message);
        })
        .finally(() => {
            if (!controller.signal.aborted) setRowsLoading(false);
        });
    return () => controller.abort();
}, [selected]);
```

The `finally` guard prevents a late abort from flipping `rowsLoading` back to `false` after the next effect already set it to `true`.

jongio (commented):
Two items on this PR.

The metrics fetch `useEffect` in `App.tsx` (line 171) doesn't cancel in-flight requests when the selected skill changes. If a user clicks through skills quickly, a slow response from a previous selection can overwrite the current one. An `AbortController` in the cleanup function fixes this (see inline comment).

The new pure helpers in `metrics.ts` (`lastNDates`, `buildDaySeries`, `groupByTest`) have clear edge cases (empty rows, missing metrics, multiple runs per day) and no unit tests. Worth adding since these functions drive all the sparkline data.

dashboard/src/skills/App.tsx
tmeschter: Fixed in 71472257: added an AbortController to cancel the in-flight request on skill change, ignore AbortErrors, and guard the loading toggle against a superseded effect.

scripts/src/frontmatter/cli.ts
```diff
@@ -682,6 +684,10 @@ export interface FrontmatterSkillResult {
   errors: string[];
   warnings: string[];
   checks: Record<string, boolean>;
+  /** Raw description text ("" when missing). */
+  description: string;
+  /** Character length of the description. */
+  descriptionLength: number;
```
JasonYeMSFT: Can we remove this field and get the length from description?

JasonYeMSFT (dismissed):

dashboard/src/skills/metrics.ts
```diff
@@ -0,0 +1,90 @@
...
+    value: number | null;
+}
+
+/** The three metrics graphed per test. */
+export type MetricKey = "durationMs" | "totalTokens" | "turns";
+
+/**
+ * Return the last `n` calendar days (inclusive of `today`) as YYYY-MM-DD
+ * strings, oldest first. `today` defaults to now (UTC date).
+ */
```
jongio, line 31: Worth adding tests for these helpers. Specific edge cases to cover:

- `lastNDates`: n=0, n=1, date boundary around midnight UTC
- `buildDaySeries`: empty rows array, rows where the metric field is absent (should yield null, not zero), multiple runs on the same day (should average), rows with dates outside the window (should be ignored)
- `groupByTest`: empty input, stable alphabetical ordering

A regression in the averaging or null-gap logic would silently corrupt the sparklines without any build or type error to catch it.

jongio (commented):
The pure helpers in `metrics.ts` (`lastNDates`, `buildDaySeries`, `groupByTest`) still lack unit tests. They drive all the sparkline data and handle edge cases (multi-run averaging, null gaps for missing metrics) that won't surface through type checking alone.

scripts/src/frontmatter/cli.ts
jongio: The consumer already handles this: `skillsFromHealthData` in App.tsx falls back to `description.length` when the metadata field is missing. And in `buildJsonResult` the value is always `description.length`, so the field carries no information the consumer can't derive on its own.

tmeschter: Done in 970ed82e: removed `descriptionLength` from the CLI JSON and the collector metadata. The Skills page now derives it via `description.length` in `skillsFromHealthData`.

dashboard/src/skills/metrics.ts
tmeschter: Added in 970ed82e: `dashboard/src/skills/__tests__/metrics.test.ts` (vitest, `npm test`). Covers all the listed cases — lastNDates n=0/n=1/UTC midnight & month boundaries; buildDaySeries empty rows, absent-metric→null, real-zero preserved, same-day averaging, out-of-window rows ignored; groupByTest empty input and stable alphabetical ordering.

dashboard/src/skills/App.tsx
```diff
@@ -0,0 +1,265 @@
...
+/** Extract plugin skills (with descriptions) from the frontmatter category. */
+function skillsFromHealthData(data: HealthData): Skill[] {
+    const items = data.categories?.frontmatter?.items ?? [];
+    const skills: Skill[] = [];
+    for (const item of items) {
+        const path = String(item.metadata?.path ?? "");
+        // Only plugin skills; the frontmatter check also covers .github/skills.
+        if (!path.startsWith("plugin/skills/")) continue;
+        const description = String(item.metadata?.description ?? "");
+        skills.push({ name: item.name, description, descriptionLength: description.length });
```
RickWinter, line 48: `description` has already been truncated to 500 characters by the frontmatter collector, so this reports roughly 503 for every longer description instead of its actual length. That understates skills approaching the 1024-character limit and makes this dashboard metric unreliable. Can we capture the raw length in `collectFrontmatterMetrics` before sanitization, then use that field here with `description.length` only as a compatibility fallback?

tmeschter: Fixed.

JasonYeMSFT (approved):

PR #2775, opened 2026-07-08
feat: add canvas-first entry gate for Foundry agent creation

qinezh:
## Summary
- Adds a canvas-first entry gate (Step 1a) to the Foundry agent create workflow when running inside the GitHub Copilot app with the `foundry-agent-canvas` extension installed.
- New `check-canvas-entry.sh` / `.ps1` scripts detect the runtime environment (`AI_AGENT=github_copilot_app_agent`) and whether the canvas extension is installed, then emit `[ACTION]` to direct the agent to open the canvas UI.
- Updates `create-hosted.md` and `quick-start-hosted.md` to run the canvas-entry check before the existing environment verification, with a loop guard (`<canvas-context>` check) and user opt-out support.

## Motivation
Currently the agent ignores the canvas extension and proceeds with the CLI-based create flow. With this change, users in the Copilot app are redirected to the canvas first (sign in, select subscription/project, configure, then Send), which then invokes the skill for remaining steps.

### DIFF
--- a/plugin/skills/microsoft-foundry/foundry-agent/create/create-hosted.md
+++ b/plugin/skills/microsoft-foundry/foundry-agent/create/create-hosted.md
@@ -42,7 +42,7 @@ For prompt agents (LLM + instructions, no container), use [create-prompt.md](cre
 
 Two pre-flight checks — run each script and act on its `[OK]` / `[WARN]` / `[ACTION]` summary prefixes.
 
-**1a — Canvas-first entry (GitHub Copilot app).** Run the canvas-entry check first (it can short-circuit the rest). **Skip 1a** if the user opts out, e.g. "skip the canvas" / "use the CLI".
+**1a — Canvas-first entry (GitHub Copilot app).** Detects whether the runtime is the GitHub Copilot app (`AI_AGENT=github_copilot_app_agent`) and the Foundry Agent Canvas extension is installed. If both are true, the canvas must be opened first so the user can authenticate and select a Foundry project before scaffolding. Run this check first (it can short-circuit the rest). **Skip 1a** if the user opts out, e.g. "skip the canvas" / "use the CLI".
 
 ```bash
 ./scripts/check-canvas-entry.sh     # macOS / Linux
--- a/plugin/skills/microsoft-foundry/foundry-agent/create/create-hosted.md
+++ b/plugin/skills/microsoft-foundry/foundry-agent/create/create-hosted.md
@@ -40,7 +40,19 @@ For prompt agents (LLM + instructions, no container), use [create-prompt.md](cre
 
 ### Step 1 -- Verify the environment
 
-Run the bundled verification script before any other create/deploy command:
+Two pre-flight checks — run each script and act on its `[OK]` / `[WARN]` / `[ACTION]` summary prefixes.
+
+**1a — Canvas-first entry (GitHub Copilot app).** Run the canvas-entry check first (it can short-circuit the rest). **Skip 1a** if the user opts out, e.g. "skip the canvas" / "use the CLI".
+
+```bash
+./scripts/check-canvas-entry.sh     # macOS / Linux
+./scripts/check-canvas-entry.ps1    # Windows (pwsh)
+```
+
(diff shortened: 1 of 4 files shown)


plugin/skills/microsoft-foundry/foundry-agent/create/create-hosted.md
```diff
@@ -40,7 +40,19 @@ For prompt agents (LLM + instructions, no container), use [create-prompt.md](cre
 
 ### Step 1 -- Verify the environment
 
-Run the bundled verification script before any other create/deploy command:
+Two pre-flight checks — run each script and act on its `[OK]` / `[WARN]` / `[ACTION]` summary prefixes.
```
anchenyi, line 43: Could you add more description here? For example, tell agent that it needs to check if users are in GitHub Copilot App and reasons for checking this.

jongio (commented):
The `eval` CI check ("Lint skill and eval specs") is failing, but this is a pre-existing issue on `main` (consistently failing since at least mid-June). Not caused by this change.

The Copilot bot flagged two issues on the earlier commit (`e437f4b`): install-path leak in `[OK]` output, and inconsistent `canvasId` syntax. Both are fixed in the current HEAD.

plugin/skills/microsoft-foundry/foundry-agent/create/scripts/check-canvas-entry.sh
```diff
@@ -0,0 +1,52 @@
...
+[ -n "$home_dir" ] && [ -n "${COPILOT_AGENT_SESSION_ID:-}" ] && \
+  dirs+=("$home_dir/.copilot/session-state/$COPILOT_AGENT_SESSION_ID/extensions/$ext")
+
+installed_at=""
+for d in "${dirs[@]}"; do
+  if [ -f "$d/extension.mjs" ]; then installed_at="$d"; break; fi
+done
+
+if [ -z "$installed_at" ]; then
+  echo "[WARN] Foundry Agent Canvas is not installed -- canvas-first gate does not apply; continue with the normal create workflow."
```
anchenyi, line 46: Should we provide installation guide here?

qinezh: no need for now - it would be better to be installed by users from canvas extension marketplace.

plugin/skills/microsoft-foundry/foundry-agent/create/create-hosted.md
qinezh: Added more description for the action '1a — Canvas-first entry'. Let me know if anything else need to be added.

anchenyi (dismissed):

jongio (commented):
One maintainability nit on the script output (inline below).

plugin/skills/microsoft-foundry/foundry-agent/create/scripts/check-canvas-entry.sh
```diff
@@ -0,0 +1,52 @@
...
+  exit 0
+fi
+
+echo "[OK] Foundry Agent Canvas is installed."
+echo '[ACTION] Canvas-first gate applies. If the message <canvas-context> already lists canvas="agent-builder", the user has already driven the canvas -- proceed with the create workflow. Otherwise open_canvas (canvasId: "agent-builder"), have the user sign in / pick a project / configure / Send, then stop -- do not scaffold.'
```
jongio: nit: This `[ACTION]` line embeds the full decision algorithm (check `<canvas-context>`, call `open_canvas`, stop). The same logic is already documented in Step 1a of both markdown files. If the workflow ever changes, you'd need to update both the script strings and the markdown.

Consider trimming to just `[ACTION] Canvas-first gate applies.` and letting the markdown be the single source of truth for what the agent does next. The script's job is detection; the skill doc already prescribes behavior.

qinezh: good advice, thanks!

jongio (commented):
Incremental review (30a95fb): scripts now emit short `[ACTION]` prefix, keeping behavioral spec in the markdown docs only. Detection logic handles missing git, unset HOME/USERPROFILE, and absent extension directory correctly on both platforms.

anchenyi (approved):

PR #2814, opened 2026-07-10
feature: Rework azure-prepare to be script-based

tmeschter:
## Description

The azure-prepare skill helps create a new Azure application and/or prepare an existing application for deployment to Azure. It describes a complex workflow that must account for multiple possible languages, application frameworks, and deployment approaches. While the workflow is very regular the complexity means there are many opportunities for the coding agent to skip steps or carry them out incorrectly. It may skip reading an essential reference file or make "small" modifications to the order or content of a series of command line tool invocations that renders them useless. It may call an MCP tool that returns far more information than can fit in the context window, and flail around trying to understand it.

In the end the LM does tend to get most of the expected result (generally, a set of workable IaC files and a deployment play, ready for azure-validate to take over) but the approach, time, and token usage can vary wildly. The deployment plan may even look good but not actually work because the LM forgot to do things like check for available quota. The azure-validate and azure-deploy skills may correct this, but it generally requires more time, tokens, and turns.

Here we replace much of the workflow logic with a script that works with the LM to keep the overall workflow on rails. The LM invokes the script which performs some deterministic work and tells the LM what it needs to do next. The LM takes care of that, and then calls the script again. The LM and script track progress and shared information in a set of .json files.

The end result is that the LM behavior is much more consistent from one run to another, as it only has to concern itself with the most recent instruction from the script rather than reasoning over the whole of the SKILL.md and the reference files. If it makes a mistake, the script will generally catch it and force the LM to correct it to make progress. Storing state in files means that essential information is not lost during compaction.

Note that this does not necessarily reduce the number of turns or total token usage. In fact, these are generally higher now as the LM actually goes through all of the expected steps rather than inadvertently skipping some. However, I believe the more consistent behavior will make it easier to identify and implement improvements of all sorts, including both bug fixes and performance improvements.

## Checklist


evals/azure-prepare/e2e-eval.yaml
```diff
@@ -50,7 +50,7 @@ stimuli:
       area: output
       skill: azure-prepare
       systemPrompt: '{"mode":"append","content":"Skip the Provisioning Limit Checklist (Step 6 in the plan template). Use reasonable default values for quota/limit columns instead of running az quota commands or invoking the azure-quotas skill. Focus your effort on generating the infrastructure and application files."}'
-      earlyTerminate: '[{"type":"skill-call","skill":"azure-validate"},{"type":"tool-call-count","count":40}]'
+      earlyTerminate: '[{"type":"skill-call","skill":"azure-validate"},{"type":"tool-call-count","count":90}]'
```
tmeschter, line 53: Repeated updates to the .json file and script invocations mean that the tool count is much higher now, though there is room for improvement.

plugin/skills/azure-prepare/references/analyze.md
```diff
@@ -1,99 +0,0 @@
-# Analyze Workspace
```
tmeschter, line 1: These high-level reference files (analyze.md, scan.md, etc.) have been replaced by operations in the script, and blocks of text that are returned by the script to the LM at every major step in the workflow. These blocks of text are under plugin/skills/azure-prepare/scripts/lib/guidance.

plugin/skills/azure-prepare/scripts/references/recipes/azcli/commands.md
```diff
@@ -1,91 +1,91 @@
...
-  --resource-group <rg-name> \
-  --location <location>
-
-# Set secret
-az keyvault secret set \
-  --vault-name <kv-name> \
-  --name <secret-name> \
-  --value <secret-value>
-```
+# Azure CLI Commands
```
tmeschter, line 1: The files in plugin/skills/azure-prepare/scripts/references were originally under plugin/skills/azure-prepare/references. There is no longer a chain of Markdown links leading from SKILL.md to these files; instead the script tells the LM to read these files as needed. If we kept them in the original location they would be flagged as orphans.

In many cases the steps described in these files could be rolled into the script itself rather than leaving them up to the LM to execute. In other cases the information in the files may not be needed at all. Dealing with them is left to a future change as this PR is already quite complicated.

Issue #2325, opened 2026-05-20
Replace skill instructions with scripts


tmeschter:
We have several skills that describe narrow workflows with very well defined sets of steps. These are good candidates to have those steps moved into a bash/PowerShell script that performs the same actions, with the skill updated to use the script. This would make the execution of the workflow more deterministic as well as reduce token usage (as the agent will only see the final output of the script rather than the output of every intermediate step).

Here are the opportunities identified by Copilot:

 ### Highly Scriptable (concrete CLI runbooks, mostly mechanical)
 
 | Skill | Key File(s) | What Could Be Scripted |
 |---|---|---|
 | azure-quotas | `SKILL.md:80-225` | Install extension → `az quota list` → match names → `az quota show` → `az quota usage show` → compare regions → `az quota update` → check status. Nearly a shell script already. |
 | azure-storage | `SKILL.md:22-50` | List accounts → list containers → list blobs → download/upload. Fully mechanical CLI flow. |
 | azure-deploy | `SKILL.md:57-70`, `references/recipes/` | Check plan → checklist → load recipe → RBAC check → execute deploy → post-deploy → verify → report. |
 | azure-resource-lookup | `SKILL.md:49-109` | Check for MCP tool → generate ARG query → run `az graph query` → shape output. |
 
 ### Partially Scriptable (mechanical core, judgment at key points)
 
 | Skill | Scriptable Parts | Judgment Needed For |
 |---|---|---|
 | azure-cost | Prereqs → azqr → discover → query costs → metrics → report | Scope, recommendation prioritization |
 | azure-compliance | Scope → azqr → analyze → categorize | Remediation prioritization |
 | airunway-aks-setup | Verify cluster → install controller → GPU → provider → deploy → summarize | Human confirmation each phase |
 | azure-validate | Verify plan → run commands → build → verify RBAC → record proof | Fixing failures |
 | azure-reliability | ARG query → classify → load checks → apply zone redundancy → verify | Consent, migration timing |
 | entra-app-registration | Register app → auth → permissions → credential → OAuth | App type, redirect URIs |
 | entra-agent-id | Blueprint → principal → agent identities → credentials → token exchange | Tenant/permissions |
 | azure-upgrade | Source/target → readiness → collect settings → upgrade → validate | Confirmation, choices |
 | azure-compute | VM troubleshooter, CRG, EMM workflows have CLI command sequences | Recommender is judgment-heavy |
[...]

Issue #2180, opened 2026-05-06
Optimize reference file utilization


JasonYeMSFT:
As of May 6th 2026, based on the telemetry we have, these reference files exist but have 0 reference count. We should
- Automate tracking the unused reference files
- Investigate if any dead reference files should be cleaned up or updated to fix routing

A few skills are particularly problematic since they have been invoked but most of their reference files are never used:
- azure-cloud-migrate
- azure-compute
- azure-diagnostics
- azure-upgrade

These skills are never used and thus their reference files are never used.
- airunway-aks-setup
- azure-cost
- azure-kubernetes
- entra-agent-id

```
# Unused reference files (# unused / # total references)

airunway-aks-setup/references/gpu-profiles.md
airunway-aks-setup/references/model-sizing.md
airunway-aks-setup/references/powershell-notes.md
airunway-aks-setup/references/steps/step-1-verify.md
airunway-aks-setup/references/steps/step-2-controller.md
airunway-aks-setup/references/steps/step-3-gpu.md
airunway-aks-setup/references/steps/step-4-provider.md
airunway-aks-setup/references/steps/step-5-deploy.md
airunway-aks-setup/references/steps/step-6-summary.md
airunway-aks-setup/references/troubleshooting.md
10 / 10

appinsights-instrumentation/references/container-apps.md
1 / 11

azure-ai/references/sdk/azure-ai-contentsafety-java.md
azure-ai/references/sdk/azure-ai-contentsafety-ts.md
azure-ai/references/sdk/azure-ai-translation-ts.md
azure-ai/references/sdk/azure-ai-vision-imageanalysis-java.md
4 / 15

azure-aigateway/references/auth-best-practices.md
azure-aigateway/references/sdk/azure-ai-contentsafety-ts.md
azure-aigateway/references/sdk/azure-mgmt-apimanagement-dotnet.md
3 / 8

azure-cloud-migrate/references/services/app-service/app-engine-to-app-service.md
azure-cloud-migrate/references/services/app-service/assessment.md
azure-cloud-migrate/references/services/app-service/beanstalk-to-app-service.md
azure-cloud-migrate/references/services/app-service/code-migration.md
azure-cloud-migrate/references/services/app-service/global-rules.md
azure-cloud-migrate/references/services/app-service/heroku-to-app-service.md
azure-cloud-migrate/references/services/container-apps/assessment-guide.md
[...]

PR #2875, opened 2026-07-17
feat: link to open issues from each skill dashboard page

### LINKED ISSUE #2870: Link to open issues
Update the "Skill" section of the dashboard so that each skill page has a link to the open issues associated with the skill. The skill name can be used as a label.

tmeschter:
Resolves #2870.

## What
Each skill page on the Skills dashboard now links to that skill's open GitHub issues, using the skill name as the issue label filter.

## Changes
- `dashboard/src/skills/App.tsx` — added an exported `issuesUrl(skillName)` helper and a **"View open issues for `<skill>` ↗"** link in the skill detail header. It targets `github.com/microsoft/GitHub-Copilot-for-Azure/issues?q=is:issue state:open label:"<skill-name>"` (URL-encoded), opening in a new tab with `rel="noopener noreferrer"`.
- `dashboard/src/skills/skills.css` — styling for `.skills-issues-link`.
- `dashboard/src/skills/__tests__/issuesUrl.test.ts` — unit tests for the URL helper (base URL, open/label query, special-character encoding).

## Notes
- Repo slug is hardcoded to `microsoft/GitHub-Copilot-for-Azure` (no existing repo constant in the dashboard).
- The skill name is used verbatim (URL-encoded) as the label value; no name mapping applied.
- The link always renders even when a skill has zero open issues (the search page simply shows an empty result).

## Validation
- `npm test -- src/skills` → 17/17 pass.
- `tsc --noEmit` → no new errors in `src/skills` (one pre-existing unrelated error in `token-usage/App.tsx` confirmed present on baseline).


dashboard/src/skills/App.tsx
```diff
@@ -254,6 +266,15 @@ export default function App() {
                             <p className="skills-desc-length">
                                 Description length: {selectedSkill.descriptionLength} characters
                             </p>
+                            <p className="skills-issues-link">
+                                <a
+                                    href={issuesUrl(selectedSkill.name)}
```
RickWinter, line 285: These links only work when a repository label exactly matches every skill name, but the repo currently has no labels for `airunway-aks-setup`, `azure-reliability`, or `python-appservice-deploy`. The generated searches for all three return zero results, and `airunway-aks-setup` already appears in open issues #2180 and #2325, so that page fails to surface associated work. Let's create and apply the missing labels before this ships, or make the link fall back to a search that can find skills without dedicated labels.

tmeschter: I've created labels for those three skills. #2180 and #2325 span multiple skills, so I wouldn't necessarily expect them to be tagged with any specific label.

dashboard/src/skills/App.tsx
```diff
@@ -254,6 +266,15 @@ export default function App() {
...
+                                    target="_blank"
+                                    rel="noopener noreferrer"
+                                >
+                                    View open issues for {selectedSkill.name} ↗
+                                </a>
```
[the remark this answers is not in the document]
tmeschter: Added a `title` attribute ("Open issues for <name> in a new tab") so users and screen readers get a hint that the link opens in a new tab, matching the convention used elsewhere in the dashboard (e.g. the snapshot viewer link).

dashboard/src/skills/skills.css
```diff
@@ -83,6 +83,20 @@
...
+}
+
+.skills-issues-link a {
+    color: var(--color-focus, #3b82f6);
+    text-decoration: none;
+}
+
+.skills-issues-link a:hover {
+    text-decoration: underline;
+}
```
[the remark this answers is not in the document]
tmeschter: Added a `.skills-issues-link a:focus-visible` rule with a visible outline and underline so the link is discoverable for keyboard users.

dashboard/src/skills/__tests__/issuesUrl.test.ts
```diff
@@ -0,0 +1,24 @@
+import { describe, it, expect } from "vitest";
+import { issuesUrl } from "../App";
+
```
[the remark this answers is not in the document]
tmeschter: Extracted `issuesUrl` and `ISSUES_REPO` into a standalone `src/skills/issuesUrl.ts` module. The unit tests (and any future callers) now import from there instead of `App.tsx`, so they no longer pull in React/recharts.

dashboard/src/skills/issuesUrl.ts
```diff
@@ -0,0 +1,11 @@
+/** Repository whose issues are surfaced from each skill page. */
+export const ISSUES_REPO = "microsoft/GitHub-Copilot-for-Azure";
```
RickWinter: Where is this export used beyond the use in this file?

tmeschter: Good catch - `ISSUES_REPO` was only referenced within `issuesUrl.ts` itself, so I removed the `export` and made it a module-private `const`. Only `issuesUrl` is exported (and imported by `App.tsx` and the unit test).

JasonYeMSFT (approved):

PR #2876, opened 2026-07-17
feat: show file count per skill on dashboard skill page

### LINKED ISSUE #2869: Number of files in the skill
Update the "Skill" section of the dashboard so that each skill page has a count of the number of files in the built/output skill.

tmeschter:
Closes #2869.

## What
Each skill's page in the dashboard `Skills` section now shows a count of the number of files in the built/output skill.

## Changes
- **`scripts/src/dashboard/collectors/frontmatter.ts`** — added a `countFilesRecursive` helper and an `addFileCounts` step that populates `metadata.fileCount` for each skill from its directory under `output/skills/`. The count is recursive and includes nested sub-skill files. Applied to both the success and non-zero-exit recovery paths; collector version bumped to `1.1.0`.
- **`dashboard/src/skills/App.tsx`** — added `fileCount` to the `Skill` type, read it from the health-data metadata, and render `Files: N` in the skill detail header.
- **`dashboard/src/skills/skills.css`** — added a matching `.skills-file-count` style.
- **Tests** — added collector tests verifying `fileCount` is computed recursively (including nested sub-skills) and defaults to 0 when the directory is missing.

## Validation
- `scripts` dashboard suite: 188/188 pass (frontmatter collector: 14/14).
- `dashboard` skills tests: 14/14 pass.
- `scripts` typechecks clean; skills App compiles clean (the only `tsc` error is a pre-existing recharts type issue in the unrelated `token-usage/App.tsx`, confirmed present on the base branch).

### DIFF
--- a/dashboard/src/skills/App.tsx
+++ b/dashboard/src/skills/App.tsx
@@ -25,6 +25,7 @@ interface Skill {
     name: string;
     description: string;
     descriptionLength: number;
+    fileCount: number;
 }
 
 /** Minimal shape of the health data returned by /api/static. */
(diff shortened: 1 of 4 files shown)


RickWinter (approved):

RickWinter (approved):

dashboard/src/skills/App.tsx
```diff
@@ -25,6 +25,7 @@ interface Skill {
     name: string;
     description: string;
     descriptionLength: number;
+    fileCount: number;
```
JasonYeMSFT, line 28: Is it feasible to collect the full list of files for each skill? Combining the telemetry, we can identify references files that are used/not used and use it to guide reorganization of content.

JasonYeMSFT (approved):

dashboard/src/skills/App.tsx
tmeschter: I've filed #2890 to track that.

PR #2930, opened 2026-07-21
fix: tighten redaction and remove unnecessary data publishing

JasonYeMSFT:
## Description

- Tighten redaction logic to cover storage account key in connection strings
- Remove unnecessary data publishing step for test result investigation

## Checklist

### DIFF
--- a/.github/workflows/test-all-integration.yml
+++ b/.github/workflows/test-all-integration.yml
@@ -326,15 +326,6 @@ jobs:
         if: always() && matrix.skill == 'microsoft-foundry'
         run: npm run quality-report || true
 
-      - name: Export report
-        if: always()
-        id: export-report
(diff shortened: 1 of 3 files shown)


saikoumudi (approved):

.github/workflows/test-all-integration.yml
```diff
@@ -326,15 +326,6 @@ jobs:
         if: always() && matrix.skill == 'microsoft-foundry'
         run: npm run quality-report || true
 
-      - name: Export report
```
tmeschter, line 329: Are the reports already being exported via some other mechanism?

tmeschter (approved):

.github/workflows/test-all-integration.yml
JasonYeMSFT: They are published to the storage account in the step below.

PR #2932, opened 2026-07-21
feat: replace azure-diagnostics messaging connectivity probe with a script

### LINKED ISSUE #2511: Replace azure-diagnostics skill messaging namespace connectivity probe with a script
## Summary

The candidate is the **messaging namespace connectivity / DNS / port probe** in the `azure-diagnostics` skill (`plugin/skills/azure-diagnostics/troubleshooting/messaging/service-troubleshooting.md`).

## Candidate description

When diagnosing Service Bus / Event Hubs connectivity, the skill runs a fixed reachability probe against the namespace FQDN:

1. `curl -v https://<namespace>.servicebus.windows.net/` (HTTPS reachability — expect an Atom feed / 401, not a connection failure)
2. `nslookup <namespace>.servicebus.windows.net` (DNS resolution)
3. TCP connectivity to the well-known messaging ports defined just above in the same file — AMQP `5671`/`5672`, HTTPS `443`, and Kafka `9093` for Event Hubs.

- **Fully deterministic** — given one input (the namespace FQDN), the steps are fixed: resolve DNS, test HTTPS, probe each known port.
- **Output-heavy and fragile to parse by hand** — `curl -v` and `nslookup` emit verbose, OS-formatted output where only "connected / resolved / which IP / which port blocked" matters.
- **Inherently cross-platform** — it benefits from a parallel bash + PowerShell pair (`Test-NetConnection` / `Resolve-DnsName` on Windows vs `curl` / `nc` / `nslookup` on Linux) producing one normalized result.

**Sketch — `test-messaging-connectivity.{sh,ps1}`:**
- **Input:** `--namespace <fqdn>` (and optionally `--kafka` to include port 9093).
- **Output:** a normalized per-check report — DNS resolved (+ IP), HTTPS reachable, and reachable/blocked per port (5671, 5672, 443, 9093) — so the agent gets a clear verdict instead of raw verbose `curl`/`nslookup` output.

> Choosing which namespace to test, and diagnosing a blocked port as IP-firewall vs. corporate proxy vs. NSG (the prose at L9–L14), require judgment and stay in prose. The script is scoped to the mechanical probe + normalized report only.

## Affected file and lines

- [`troubleshooting/messaging/service-troubleshooting.md` — ports/firewall context (L9–L14)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-diagnostics/troubleshooting/messaging/service-troubleshooting.md#L9-L14)
- [`troubleshooting/messaging/service-troubleshooting.md` — Quick Connectivity Test (L16–L24)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-diagnostics/troubleshooting/messaging/service-troubleshooting.md#L16-L24)

## Next steps

tmeschter:
## Summary

Fixes #2511. Replaces the inline messaging namespace connectivity probe in the `azure-diagnostics` skill with a tested, cross-platform script pair.

Previously `troubleshooting/messaging/service-troubleshooting.md` inlined a fragile `curl -v` + `nslookup` probe. This adds `scripts/test-messaging-connectivity.{sh,ps1}` that run the mechanical probe and emit one normalized report.

## Changes

- **Add** `plugin/skills/azure-diagnostics/scripts/test-messaging-connectivity.sh` and `.ps1`:
  - Input: a namespace (full FQDN or bare name; `.servicebus.windows.net` appended automatically) plus optional `--kafka` / `-Kafka` for Event Hubs Kafka port 9093.
  - Checks: DNS resolution -> HTTPS reachability (443) -> TCP ports 5671, 5672, 443 (+9093).
  - Output: a per-check table plus a plain-language summary. A blocked port is a valid diagnostic result (exit 0); only invalid arguments exit non-zero.
- **Edit** `troubleshooting/messaging/service-troubleshooting.md`: replace the inline probe with markdown-linked script references, example invocations, and a one-line description. The judgment-based firewall/ports prose stays in markdown.
- **Add** an integration eval stimulus in `evals/azure-diagnostics/eval.yaml` that presents a ""cannot connect at all"" Service Bus scenario and asserts the skill surfaces `test-messaging-connectivity`.

## Validation

- Smoke-tested both scripts (Git Bash + PowerShell): reachable path, DNS-fail path, `--kafka`, and usage error. Caught and fixed a bug where the bash `nslookup` fallback returned the DNS server IP instead of the resolved host.
- `npm run build` (version stamping) succeeds; scripts copied to `output/`.
- `npm run references` and `npm run frontmatter` pass on the built skill.
- `npm run vally validate-stimulus` passes.


plugin/skills/azure-diagnostics/troubleshooting/messaging/service-troubleshooting.md
```diff
@@ -15,12 +15,21 @@ If the client **cannot connect at all**:
 
 ### Quick Connectivity Test
 
+Run the connectivity probe script with the namespace. It resolves DNS, tests HTTPS reachability, and probes the messaging ports (AMQP `5671`/`5672`, HTTPS `443`), returning a normalized report instead of raw `curl`/`nslookup` output. Add `--kafka` / `-Kafka` to also probe the Event Hubs Kafka port `9093`.
+
+```powershell
+..\..\scripts\test-messaging-connectivity.ps1 -Namespace <namespace>
+```
```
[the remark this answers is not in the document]
tmeschter: This should be a Markdown link.

plugin/skills/azure-diagnostics/scripts/test-messaging-connectivity.sh
```diff
@@ -0,0 +1,190 @@
...
+
+while [ $# -gt 0 ]; do
+    case "$1" in
+        --kafka)
+            INCLUDE_KAFKA=1
+            ;;
+        -h|--help)
+            grep '^#' "$0" | sed 's/^# \{0,1\}//'
+            exit 0
+            ;;
```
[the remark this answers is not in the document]
tmeschter: Fixed in a9fa671c — the --help printer now filters out shebang lines (grep -v '^#!'), so the help text no longer shows !/usr/bin/env bash.

plugin/skills/azure-diagnostics/scripts/test-messaging-connectivity.ps1
```diff
@@ -0,0 +1,134 @@
...
+}
+
+# ── HTTPS reachability ────────────────────────────────────────────────────────
+# On success the namespace returns an Atom feed or HTTP 401 — either proves the
+# endpoint is reachable. A connection failure means blocked.
+function Get-HttpsResult {
+    try {
+        $resp = Invoke-WebRequest -Uri "https://$fqdn/" -Method Get -TimeoutSec 15 `
+            -UseBasicParsing -SkipHttpErrorCheck -ErrorAction Stop
+        return "reachable (HTTP $($resp.StatusCode))"
```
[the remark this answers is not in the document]
tmeschter: Fixed in a9fa671c — removed -SkipHttpErrorCheck (unavailable in Windows PowerShell 5.1) so the call binds on both editions. HTTP-error responses (e.g. 401, which proves reachability) are now read from the caught exception's Response.StatusCode, with a TCP-443 probe as a fallback. Verified the script runs clean on both PowerShell 7 and Windows PowerShell 5.1.

evals/azure-diagnostics/eval.yaml
```diff
@@ -111,6 +111,30 @@ stimuli:
...
+  # surface the test-messaging-connectivity probe script rather than raw
+  # curl/nslookup commands.
+  - name: "Service Bus namespace connectivity probe"
+    prompt: "My app can't connect to my Azure Service Bus namespace 'contoso' at all. How can I check whether it's a network or firewall problem — like which of the messaging ports are actually reachable?"
+    tags:
+      type: integration
+      tier: full
+      cost: llm
+      area: output
+    graders:
```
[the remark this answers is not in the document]
tmeschter: Addressed in a9fa671c — added an arlyTerminate (assistant-message-match on the script name + a tool-call-count cap) so the run stops as soon as the probe script is surfaced, keeping cost in line with the routing stimuli. rea: output is intentional here: this is an output-content assertion (does the skill surface the probe script?) rather than a routing test, so it's deliberately not counted in skill-invocation reporting. Added a comment in the YAML explaining this.

plugin/skills/azure-diagnostics/troubleshooting/messaging/service-troubleshooting.md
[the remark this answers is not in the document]
tmeschter: Fixed in a9fa671c — the scripts are now referenced as Markdown links, and the examples state the working directory (the skill root, plugin/skills/azure-diagnostics) with a # from plugin/skills/azure-diagnostics comment on each command, using skill-root-relative paths (./scripts/...).

plugin/skills/azure-diagnostics/scripts/test-messaging-connectivity.ps1
```diff
@@ -0,0 +1,134 @@
...
+    Also probe the Event Hubs Kafka endpoint on port 9093.
+.EXAMPLE
+    .\test-messaging-connectivity.ps1 -Namespace contoso
+    # Tests contoso.servicebus.windows.net (DNS, HTTPS, AMQP 5671/5672, 443)
+.EXAMPLE
+    .\test-messaging-connectivity.ps1 -Namespace contoso.servicebus.windows.net -Kafka
+    # Also probes Event Hubs Kafka port 9093
+#>
+param(
+    [Parameter(Mandatory)][string]$Namespace,
```
tmeschter: Mandatory parameters can be problematic as PowerShell will prompt for them, which doesn't work well in non-interactive scenarios such as this. Remove `[Parameter(Mandatory)]` and replace it with an explicit check.

plugin/skills/azure-diagnostics/scripts/test-messaging-connectivity.ps1
```diff
@@ -0,0 +1,134 @@
...
+    [switch]$Kafka
+)
+
+$ErrorActionPreference = "Stop"
```
tmeschter: We don't want `Write-Error` to terminate the script. Remove this so we get the default for `$ErrorActionPreference`, and adjust the rest of the script accordingly.

plugin/skills/azure-diagnostics/scripts/test-messaging-connectivity.ps1
tmeschter: Removed [Parameter(Mandatory)] from \ and replaced it with an explicit [string]::IsNullOrWhiteSpace check that writes an error and exits 2. This avoids PowerShell prompting for the value in non-interactive scenarios.

plugin/skills/azure-diagnostics/scripts/test-messaging-connectivity.ps1
tmeschter: Removed \Continue = 'Stop' so the script uses the default preference and a stray non-terminating error no longer aborts the whole probe. The DNS/HTTPS/TCP operations that must fail hard already use explicit -ErrorAction Stop inside their try/catch blocks, so behavior there is unchanged.

evals/azure-diagnostics/eval.yaml
```diff
@@ -111,6 +111,36 @@ stimuli:
...
+      type: integration
+      tier: full
+      cost: llm
+      # area: output — this is an output-content assertion (does the skill surface
+      # the probe script?), not a routing test, so it is intentionally not counted
+      # in skill-invocation reporting.
+      area: output
+      # Cap cost: terminate as soon as the probe script is surfaced. The
+      # output-matches grader still sees it in the accumulated assistant messages.
+      earlyTerminate: '[{"type":"assistant-message-match","contentPattern":"test-messaging-connectivity"},{"type":"tool-call-count","count":3}]'
```
JasonYeMSFT: Should this be tool-call-result check instead? Is saying nothing but directly running the script is a reasonable outcome?

plugins/azure-skills/skills/azure-diagnostics/troubleshooting/messaging/service-troubleshooting.md
```diff
@@ -15,12 +15,26 @@ If the client **cannot connect at all**:
...
+
+```powershell
+# from plugin/skills/azure-diagnostics
+.\scripts\test-messaging-connectivity.ps1 -Namespace <namespace>
+```
 ```bash
-# Test endpoint reachability (expect Atom feed XML on success)
-curl -v https://<namespace>.servicebus.windows.net/
+# from plugin/skills/azure-diagnostics
+./scripts/test-messaging-connectivity.sh <namespace>
```
RickWinter: This bash example will fail on Linux because the new `.sh` file is committed as mode `100644`, so `./scripts/test-messaging-connectivity.sh` is not executable after checkout. Either commit the script with the executable bit or invoke it through `bash` in both bash examples. As written, the quick connectivity test can hand users a permission error before it runs the probe.

```suggestion
bash ./scripts/test-messaging-connectivity.sh <namespace>
```

tmeschter: Fixed. The .sh was committed as mode `100644`; it's now marked executable (`100755`), and both bash examples invoke it via `bash ./scripts/test-messaging-connectivity.sh` so the quick connectivity test runs cleanly on Linux/macOS regardless of whether the executable bit survives the build copy.

evals/azure-diagnostics/eval.yaml
tmeschter: Good call — the eval was reworked along exactly these lines. It's no longer an `assistant-message-match` on the script name; it now asserts the agent actually *runs* the probe via the `shell-command-invoked` grader, with an early-terminate that fires on the script's tool call (`tool-call-match` on `test-messaging-connectivity\.(sh|ps1)`). So the agent running the script directly — with or without narration — is the intended passing outcome, and the check keys off the tool call rather than the message text.

evals/azure-skills/azure-diagnostics/eval.yaml
```diff
@@ -111,6 +111,51 @@ stimuli:
...
+      # full plugin skill set (the post-multi-plugin skill-loading model loads
+      # every plugin skill unless requiredSkills pins the ones under test).
+      requiredSkills:
+        - azure-diagnostics
+      # Terminate the moment the agent starts the probe script — we only need to
+      # observe the attempt, not wait for the (network-timeout-laden) run to
+      # finish. The shell-command-invoked grader below matches the same tool_call
+      # start event, so early termination does not drop the signal. A
+      # tool-call-count cap is a cost fallback if the agent never runs the script.
+      earlyTerminate: '[{"type":"tool-call-match","toolPattern":"bash|powershell|pwsh|run_in_terminal","argsPattern":"test-messaging-connectivity\\.(sh|ps1)"},{"type":"tool-call-count","count":12}]'
```
JasonYeMSFT, line 139: Could you please double check if the tool-call-match condition works fine with the shell-command-invoked grader?

evals/azure-diagnostics/eval.yaml
tmeschter: Verified — the `tool-call-match` early-terminate and the `shell-command-invoked` grader are compatible because they key off the *same* `tool_call` event (the invocation), not the completion/result. I re-ran the stimulus and checked the per-run correlation:

- Every run where the probe script was invoked: the early-terminate fired **and** `shell-command-invoked` passed.
- Runs that failed did so only because the agent never invoked the script at all (nothing for either the terminate or the grader to match) — there is **no** case where the script ran but the grader missed it due to early termination.

So the terminate condition does not race the grader: terminating on the tool call still leaves the `tool_call` event in the trajectory for the grader to score. (The pass/fail variance you may see across runs is skill-routing stochasticity — whether the agent engages azure-diagnostics vs. another skill — which is independent of the grader interaction.)

evals/azure-skills/azure-diagnostics/eval.yaml
tmeschter: Verified — the `tool-call-match` early-terminate and the `shell-command-invoked` grader are compatible because they key off the *same* `tool_call` event (the invocation), not the completion/result. I re-ran the stimulus and checked the per-run correlation:

- Every run where the probe script was invoked: the early-terminate fired **and** `shell-command-invoked` passed.
- Runs that failed did so only because the agent never invoked the script at all (nothing for either the terminate or the grader to match) — there is **no** case where the script ran but the grader missed it due to early termination.

So the terminate condition does not race the grader: terminating on the tool call still leaves the `tool_call` event in the trajectory for the grader to score. (The pass/fail variance you may see across runs is skill-routing stochasticity — whether the agent engages azure-diagnostics vs. another skill — which is independent of the grader interaction.)

JasonYeMSFT (approved):

saikoumudi (approved):

PR #2935, opened 2026-07-21
feat: extract azure-diagnostics dump-everything blocks into scripts

### LINKED ISSUE #2509: Replace azure-diagnostics skill combined dump-everything diagnostic blocks with scripts
## Summary

The candidate is the **combined "dump everything" diagnostic scripts** in the `azure-diagnostics` skill — blocks that are *already written as chained shell scripts* in the Container Apps and App Service references.

## Candidate description

Both per-service references ship a single inlined `bash` block that chains several `az` calls with `&&` and labeled `echo` separators to collect all diagnostic info for an app in one go:

- **Container Apps** — `echo "=== ... ===" && az containerapp show ... && az containerapp revision list ... && az containerapp logs show ...` (inputs: `APP`, `RG`)
- **App Service** — the analogous `az webapp` configuration / deployment / log chain (inputs: `APP`, `RG`)

- **Already a deterministic script** — fully chained, zero branching, the only inputs are the app name and resource group. It is essentially asking to be a real `.sh`/`.ps1` file.
- **Output-heavy** — each runs 3–4 `az` calls printing raw JSON/log blobs; a script can label and summarize each section instead of dumping everything.
- **A near-duplicate pattern across two files** — the labeled-echo + sequential `az ... --query` config/revision/deployment/log fetch is the same shape for both services, arguing for one parameterized script (service-type switch) or two siblings sharing a helper.

**Sketch — `containerapp-diagnostics.{sh,ps1}` + `appservice-diagnostics.{sh,ps1}` (shared helper):**
- **Input:** `--name`, `--resource-group`.
- **Output:** labeled, summarized sections (config, revisions/deployments, recent logs) rather than concatenated raw JSON — replacing brittle inline `&&` quoting with a maintained, self-explaining script.

> Interpreting the gathered config/logs afterward is the agent's job and stays in prose. The script only **collects and labels**.

## Affected file and lines

- [`references/container-apps/README.md` — "Get All Diagnostic Info" combined block (L96–L105)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-diagnostics/references/container-apps/README.md#L96-L105)
- [`references/app-service/README.md` — "Combined Diagnostic Script" block (L175–L183)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-diagnostics/references/app-service/README.md#L175-L183)

## Next steps

tmeschter:
## Summary

Resolves #2509.

Replaces the two inline "dump everything" diagnostic blocks in the `azure-diagnostics` skill with maintained, cross-platform scripts.

Previously both per-service references shipped a single inlined `bash` block chaining several `az` calls with `&&` and labeled `echo` separators. These are now extracted into `.sh` + `.ps1` scripts under the skill's `scripts/` folder, following the existing `azure-quotas` convention.

## Changes

- **New scripts** in `plugin/skills/azure-diagnostics/scripts/`:
  - `containerapp-diagnostics.{sh,ps1}` — labels & collects revisions, registry config, ingress config, recent logs (`--tail 20`)
  - `appservice-diagnostics.{sh,ps1}` — labels & collects app config, recent deployments, app settings (names), custom domains
  - Args: `--name`/`-Name` and `--resource-group`/`-ResourceGroup` (bash also accepts positional args and an optional `--subscription`)
  - Each prints a leading/trailing summary line describing what it collected. Scripts only **collect and label** — interpreting the output stays the agent's job in prose.
- **Updated references**: `container-apps/README.md` and `app-service/README.md` now link to the scripts (markdown links) with PowerShell and bash sample invocations instead of the inline blocks.

## Behavior

Script behavior is intentionally **identical** to the original inline blocks (same `az` queries, same `--query` projections). The Azure CLI defaults to JSON output, so no explicit `-o json` was added.

## Validation

- `npm run build` ✅
- `npm run frontmatter` (built output) ✅
- `npm run references` ✅ — 26 skills pass, scripts properly linked, no orphaned/escaped references
- `npm run tokens compare` — no `azure-diagnostics` regressions
- Bash `-n` + PowerShell parser syntax checks — all 4 scripts OK

Note: the eval suite (`evals/azure-diagnostics/eval.yaml`) and former integration tests only exercise skill **routing/invocation**, not script execution, so they're unaffected.


plugin/skills/azure-diagnostics/scripts/appservice-diagnostics.ps1
```diff
@@ -0,0 +1,53 @@
...
+.PARAMETER ResourceGroup
+    Resource group that contains the web app.
+.PARAMETER Subscription
+    Azure subscription ID. Defaults to the current subscription.
+.EXAMPLE
+    .\appservice-diagnostics.ps1 -Name my-app -ResourceGroup my-rg
+    # Collects config, recent deployments, app settings, and custom domains for my-app
+#>
+param(
+    [Parameter(Mandatory)][string]$Name,
```
tmeschter: Avoid mandatory parameters in PowerShell. If the parameter is not provided PowerShell will interactively wait for the user to provide the value; this is undesirable in non-interactive scenarios. Instead, explicitly check the value of the parameter and exit early with an appropriate error message.

plugin/skills/azure-diagnostics/scripts/containerapp-diagnostics.ps1
```diff
@@ -0,0 +1,53 @@
...
+.PARAMETER ResourceGroup
+    Resource group that contains the Container App.
+.PARAMETER Subscription
+    Azure subscription ID. Defaults to the current subscription.
+.EXAMPLE
+    .\containerapp-diagnostics.ps1 -Name my-app -ResourceGroup my-rg
+    # Collects revisions, registry/ingress config, and recent logs for my-app
+#>
+param(
+    [Parameter(Mandatory)][string]$Name,
```
tmeschter: Same issue with mandatory parameters as in the other script.

plugin/skills/azure-diagnostics/scripts/containerapp-diagnostics.sh
```diff
@@ -0,0 +1,67 @@
...
+# Support both --flag and positional styles.
+POSITIONAL=()
+while [ $# -gt 0 ]; do
+    case "$1" in
+        --name|-n)             APP="$2"; shift 2 ;;
+        --resource-group|-g)   RG="$2"; shift 2 ;;
+        --subscription|-s)     SUBSCRIPTION="$2"; shift 2 ;;
+        *)                     POSITIONAL+=("$1"); shift ;;
+    esac
+done
```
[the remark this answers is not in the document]
tmeschter: Fixed in 4f3cc030cc3df1765da48f073315775adb44cb0c. The parser now validates that a value follows each flag (prints `Error: option '--name' requires a value.` + usage and exits 1 instead of hitting a `set -u` unbound-variable error), and unknown `--...`/`-x` options now error out with usage instead of being silently treated as positional args.

plugin/skills/azure-diagnostics/scripts/appservice-diagnostics.sh
```diff
@@ -0,0 +1,69 @@
...
+# Support both --flag and positional styles.
+POSITIONAL=()
+while [ $# -gt 0 ]; do
+    case "$1" in
+        --name|-n)             APP="$2"; shift 2 ;;
+        --resource-group|-g)   RG="$2"; shift 2 ;;
+        --subscription|-s)     SUBSCRIPTION="$2"; shift 2 ;;
+        *)                     POSITIONAL+=("$1"); shift ;;
+    esac
+done
```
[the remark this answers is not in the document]
tmeschter: Fixed in 4f3cc030cc3df1765da48f073315775adb44cb0c. Same hardening applied here: each flag now requires a following value (clear error + usage, exit 1) and unknown options are rejected instead of mis-parsed as positional args.

plugin/skills/azure-diagnostics/scripts/appservice-diagnostics.ps1
tmeschter: Fixed in 4f3cc030cc3df1765da48f073315775adb44cb0c. Removed `[Parameter(Mandatory)]` from `-Name`/`-ResourceGroup` and instead check them explicitly, calling `Write-Error` + `exit 1` when missing, so the script fails fast in non-interactive scenarios rather than prompting.

plugin/skills/azure-diagnostics/scripts/containerapp-diagnostics.ps1
tmeschter: Fixed in 4f3cc030cc3df1765da48f073315775adb44cb0c. Same change here: `-Name`/`-ResourceGroup` are no longer mandatory params; the script checks them and exits early with an error message instead of prompting.

JasonYeMSFT (dismissed):

evals/azure-skills/azure-diagnostics/script-invocation.eval.yaml
```diff
@@ -0,0 +1,90 @@
...
+  diagnostic scripts (containerapp-diagnostics, appservice-diagnostics) when
+  asked to collect all diagnostics for a Container App or App Service web app.
+  Grades routing plus the presence of a shell tool call running the script.
+
+tags:
+  type: integration
+  skill: azure-diagnostics
+
+defaults:
+  runs: 5
```
JasonYeMSFT: The per stimulus config for runs doesn't actually work today. Since all the stimulis in this file override it with 1, let's set the default runs to 1.
 https://github.com/microsoft/vally/issues/430

JasonYeMSFT (dismissed):

evals/azure-skills/azure-diagnostics/script-invocation.eval.yaml
tmeschter: Fixed in e55d3ca1. Moved 
uns: 1 to the defaults block and removed the per-stimulus config.runs overrides (which are a no-op today per microsoft/vally#430).

evals/azure-skills/azure-diagnostics/script-invocation.eval.yaml
```diff
@@ -0,0 +1,90 @@
...
+  # Asserts: isSkillInvoked + tool call running containerapp-diagnostics.(sh|ps1).
+  # Terminates at that tool call so the script is not executed.
+  - name: "Container App diagnostics uses the bundled script"
+    prompt: "Collect all the diagnostic info for my Azure Container App named my-app in resource group my-rg."
+    tags:
+      type: integration
+      tier: full
+      cost: llm
+      area: response-quality
+      earlyTerminate: '[{"type":"tool-call-match","toolPattern":"bash|powershell|pwsh|run_in_terminal","argsPattern":"containerapp-diagnostics\\.(sh|ps1)"},{"type":"tool-call-count","count":6}]'
```
JasonYeMSFT: I think you need to use tool-call-result condition or the grader checking for this tool call will fail.

JasonYeMSFT (changes requested):

evals/azure-skills/azure-diagnostics/script-invocation.eval.yaml
tmeschter: Good catch — fixed in c9ee3f18. Switched both stimuli's earlyTerminate from \	ool-call-match\ (fires at execution_start) to \	ool-call-result\ (fires at execution_complete). The built-in \	ool-calls\ grader only counts a \
equired\ match when it sees the tool_result, so terminating at start left a dangling tool_call it couldn't match. The bundled scripts are read-only \z\ queries that no-op without resources, so letting the call complete once is harmless and the run still aborts right after.

JasonYeMSFT (approved):

saikoumudi (approved):

PR #2936, opened 2026-07-21
feat: replace azure-diagnostics AKS baseline sweep with aks-baseline script

### LINKED ISSUE #2506: Replace azure-diagnostics skill AKS cluster baseline diagnostic sweep with a script
## Summary

The candidate is the **AKS cluster baseline diagnostic sweep** in the `azure-diagnostics` skill — a fixed, read-only set of `az aks` / `kubectl` commands that is duplicated across four files.

## Candidate description

For nearly every AKS investigation, the skill runs the same ordered, read-only "what's the state of my cluster" sweep:

1. `az aks show -g <rg> -n <cluster> --query "provisioningState"`
2. `az aks nodepool list ...`
3. `az monitor activity-log list -g <rg> --max-events 20 -o table`
4. `kubectl get nodes -o wide`
5. `kubectl get pods -A --field-selector=status.phase!=Running,status.phase!=Succeeded`
6. `kubectl get pods -n kube-system ...`
7. `kubectl get events -A --sort-by=.lastTimestamp` (often `| head -40`)

- **A fixed, ordered, read-only sequence** with no branching — pure data gathering.
- **Heavily duplicated** — the same commands recur verbatim across `general-diagnostics.md`, `aks-troubleshooting.md`, and two CLI-fallback blocks in `command-flows.md`.
- **Output-heavy** — ~7 commands emit large blobs; the agent currently pastes raw output. The unhealthy-pod field-selector and `events --sort-by | head` are output-reduction wins a script does once, correctly.

**Sketch — `aks-baseline.{sh,ps1}`:**
- **Input:** `-g/--resource-group`, `-n/--cluster`, optional `--namespace`.
- **Output:** a single labeled digest (provisioning state, nodepool summary, node readiness, unhealthy pods, kube-system health, recent warning events) so the agent gets one summarized result instead of 7 raw dumps.

> The Evidence-Order / Scope-Bucket classification and root-cause interpretation (`aks-troubleshooting.md` L37–60) require judgment and stay in prose. The script only **gathers and digests**.

## Affected file and lines

- [`troubleshooting/aks/general-diagnostics.md` — baseline sweep (L7–L23)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-diagnostics/troubleshooting/aks/general-diagnostics.md#L7-L23)
- [`troubleshooting/aks/aks-troubleshooting.md` — Safe Fallback Checks (L72–L81)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-diagnostics/troubleshooting/aks/aks-troubleshooting.md#L72-L81)
- [`troubleshooting/aks/references/command-flows.md` — Cluster & Kubernetes baseline CLI fallback (L11–L33)](https://github.com/microsoft/GitHub-Copilot-for-Azure/blob/3890cbfb65c548ce8daa96cabd1d8de63f7bbcca/plugin/skills/azure-diagnostics/troubleshooting/aks/references/command-flows.md#L11-L33)

## Next steps

tmeschter:
## Description

Replaces the AKS cluster baseline diagnostic sweep in the `azure-diagnostics` skill with a reusable, cross-platform script, resolving #2506.

Previously the same fixed, ordered, read-only "what's the state of my cluster" sweep (`az aks show`, `az aks nodepool list`, `az monitor activity-log list`, `kubectl get nodes/pods/events`) was duplicated verbatim across three AKS files. The agent had to generate ~7 commands and paste ~7 raw output blobs on every run.

### Changes

- **New scripts** `plugin/skills/azure-diagnostics/scripts/aks-baseline.sh` and `aks-baseline.ps1` (co-located with the skill, matching the `azure-quotas` convention). They run the read-only sweep once and print a single labeled digest: provisioning state, node pool summary, recent activity log, node readiness, unhealthy pods, kube-system health, and recent warning events, plus a closing summary. Params: `-g`/`-n` (required), optional `--namespace`/`--subscription`. Each step is guarded so a single failure (e.g. `kubectl` not authenticated) prints a note and the sweep continues; both scripts are strictly read-only.
- **Updated markdown** in `general-diagnostics.md`, `aks-troubleshooting.md`, and `references/command-flows.md` to reference the script via a markdown link, briefly describe what it does, and show sample bash + PowerShell invocations, replacing the duplicated inline command blocks. Judgment prose (Evidence Order, Scope Buckets) and pod-specific deep-dive commands (`kubectl describe`, `kubectl logs --previous`) remain inline.

### Validation

- `references` validator passes (no broken/escaped links, no orphaned files)
- Token budgets: all changed files within limits (command-flows.md 905 < 1000)
- `npm run build` stamps versions and copies both scripts into `output/skills/azure-diagnostics/scripts/`
- `bash -n` and PowerShell parser both clean; `.sh` stored with LF line endings

## Checklist

- [x] Tests pass locally (`cd tests && npm test`) — no functional tests exercise these scripts; validated via `references`, token check, build, and script syntax checks
- [x] **If modifying skill descriptions:** N/A — SKILL.md frontmatter/description unchanged

## Related Issues

Fixes #2506


plugin/skills/azure-diagnostics/scripts/aks-baseline.ps1
```diff
@@ -0,0 +1,132 @@
...
+    Optional namespace for an extra pod overview.
+.PARAMETER Subscription
+    Optional Azure subscription id. Defaults to the current subscription.
+.EXAMPLE
+    .\aks-baseline.ps1 -ResourceGroup my-rg -Cluster my-cluster
+.EXAMPLE
+    .\aks-baseline.ps1 -ResourceGroup my-rg -Cluster my-cluster -Namespace payments
+#>
+param(
+    [Parameter(Mandatory)][Alias("g")][string]$ResourceGroup,
```
tmeschter: Avoid mandatory parameters. If they are not provided the script will interactively wait for the value to be provided, which doesn't work for non-interactive scenarios like plugin scripts. Instead, explicitly check that the parameter has a value; if not, exit early with an appropriate error message explaining what needs to be provided.

plugin/skills/azure-diagnostics/scripts/aks-baseline.ps1
```diff
@@ -0,0 +1,132 @@
...
+    [Parameter(Mandatory)][Alias("n")][string]$Cluster,
+    [string]$Namespace,
+    [string]$Subscription
+)
+
+$ErrorActionPreference = "Continue"
```
tmeschter: This is the default value for `$ErrorActionPreference`, so this line is not necessary.

plugin/skills/azure-diagnostics/scripts/aks-baseline.sh
```diff
@@ -0,0 +1,131 @@
...
+while [ $# -gt 0 ]; do
+    case "$1" in
+        -g|--resource-group) RESOURCE_GROUP="${2:-}"; shift 2 ;;
+        -n|--cluster) CLUSTER="${2:-}"; shift 2 ;;
+        --namespace) NAMESPACE="${2:-}"; shift 2 ;;
+        --subscription) SUBSCRIPTION="${2:-}"; shift 2 ;;
+        -h|--help) usage ;;
+        *) echo "Unknown argument: $1" >&2; usage ;;
+    esac
+done
```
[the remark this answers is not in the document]
tmeschter: Fixed. usage() now takes an exit code (-h/--help exits 0, error paths exit 1), and each option validates a value is present via a require_value helper before shift 2, so a missing value (e.g. a trailing -g) now prints an error and exits instead of looping. Verified: --help -> exit 0, -g with no value -> exit 1.

plugin/skills/azure-diagnostics/scripts/aks-baseline.sh
```diff
@@ -0,0 +1,131 @@
...
+fi
+
+# 6. kube-system health --------------------------------------------------------
+section "6. kube-system health"
+run "kube-system pods" kubectl get pods -n kube-system -o wide
+
+# 7. Recent warning events -----------------------------------------------------
+section "7. Recent warning events (last 40, sorted by time)"
+run "warning events" bash -c \
+    "kubectl get events -A --field-selector=type=Warning --sort-by=.lastTimestamp 2>/dev/null | tail -n 40"
```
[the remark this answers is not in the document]
tmeschter: Fixed. The warning-events subshell now runs `set -o pipefail;` before the pipeline, so a kubectl failure propagates through tail and is surfaced by run().

plugin/skills/azure-diagnostics/scripts/aks-baseline.ps1
tmeschter: Fixed. Removed the [Parameter(Mandatory)] attributes and instead explicitly check -ResourceGroup/-Cluster; if either is missing the script writes a clear error, prints usage, and exits 1. Verified it exits non-interactively (no prompt) when a required value is omitted.

plugin/skills/azure-diagnostics/scripts/aks-baseline.ps1
tmeschter: Fixed. Removed the redundant $ErrorActionPreference = "Continue" line since Continue is already the default.

JasonYeMSFT (dismissed):

plugins/azure-skills/skills/azure-diagnostics/scripts/aks-baseline.sh
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env bash
```
RickWinter, line 1: This script is added with mode `100644`, but every bash example invokes it directly as `./scripts/aks-baseline.sh`. On Linux and macOS that fails with permission denied before any diagnostics run. Let's either set the executable bit on this file or change the documented invocations to `bash ./scripts/aks-baseline.sh` so the fallback works from a fresh checkout.

plugins/azure-skills/skills/azure-diagnostics/scripts/aks-baseline.sh
```diff
@@ -0,0 +1,139 @@
...
+        --query "[].{time:eventTimestamp, operation:operationName.value, status:status.value, resource:resourceId}" \
+        -o table
+
+# 4. Node readiness ------------------------------------------------------------
+section "4. Node readiness"
+run "node readiness" kubectl get nodes -o wide
+
+# 5. Unhealthy pods ------------------------------------------------------------
+section "5. Unhealthy pods (not Running/Succeeded)"
+if kubectl get pods -A --field-selector=status.phase!=Running,status.phase!=Succeeded 2>/dev/null | grep -q .; then
```
RickWinter: This phase selector misses Running pods with container-level failures, which includes common incident states like `CrashLoopBackOff` and high restart counts. The baseline can report "No unhealthy pods" while `kubectl get pods -A -o wide` would show the failing workload. Let's make this check inspect readiness, waiting reasons, or include the all-pods overview that the old flow had. The PowerShell script uses the same selector and should be kept in sync.

plugins/azure-skills/skills/azure-diagnostics/scripts/aks-baseline.sh
tmeschter: There is no way for a plugin to set the executable bit. In practice the LM will always be running this with the equivalent of `bash ./scripts/aks-baseline.sh` so the shebang is irrelevant.

plugins/azure-skills/skills/azure-diagnostics/scripts/aks-baseline.sh
tmeschter: Good catch. The phase-based field selector (`status.phase!=Running,!=Succeeded`) missed pods that stay in phase `Running` while a container fails, including `CrashLoopBackOff` and `ImagePullBackOff`.

Step 5 now lists all pods (`kubectl get pods -A -o wide`) and flags a pod when it is **not Ready** (READY numerator != denominator on a Running pod), has a **non-Running / non-terminal STATUS** (CrashLoopBackOff, ImagePullBackOff, Error, Pending, etc.), or has a **high restart count** (>= 5). Terminal pods (Completed/Succeeded) are excluded so finished jobs aren't falsely flagged. The PowerShell script uses the identical READY/STATUS/RESTARTS logic — I verified both flag the same rows against a sample `kubectl` table (CrashLoopBackOff, ImagePullBackOff, not-ready, and flapping pods all caught; a Completed job correctly ignored).

Fixed in 9c824069.

JasonYeMSFT (approved):

saikoumudi (approved):

PR #3053, opened 2026-08-10
feature: add discover-azure-skills skill

### LINKED ISSUE #2794: Build the installer skill in the core plugin for plugin discovery
**Standalone work item**

## Scope
- Add an installer skill to the core `azure` plugin.
- Skill knows all plugins in `azure-skills` and helps users find a skill by scenario.

## Acceptance criteria
- Prompting e.g. "I want to do X in Azure, find a skill that can help" surfaces the relevant plugin/skill.

JasonYeMSFT:
## Description

A new discover-azure-skills skill to help discover available skills among all the plugins we own. The discover-azure-skills skill can be triggered when the user explicitly asks the agent to find a new skill/plugin or asks the agent to do an Azure related task. However, it won't be triggered when the user asks informational questions on how to do an Azure related task.

To achieve this trigger condition, a SessionStart hook that injects a higher level instruction is necessary. This PR adds a new SessionStart event hook that injects instructions as additional context to the session. The injected instructions costs ~64 tokens.

The plugin build script is refactored to support building per-plugin hooks. The repo now maintains a shared set of hooks and optionally hooks per plugin. The build script can merge the shared hooks and the per-plugin hooks at compile time. Right now, we share telemetry hook for all plugins and only add this new hook for discover skills to the azure-skills plugin.

## Checklist

## Related Issues

resolves #2794


JasonYeMSFT:
Screenshot 2026-08-10 at 10 41 50 AM

Example UX in Copilot CLI.

plugins/azure-skills/skills/discover-azure-skills/references/install/other.md
```diff
@@ -0,0 +1,9 @@
+# Steps
+
+1. Install each skill using `skills` package
+
+```
+npx skills add https://github.com/microsoft/azure-skills/tree/main/.github/plugins/{plugin-dirname}/skills/{skill-name}
```
JasonYeMSFT, line 6: Notes for reviewers:

1. How do you feel about requiring npx? I haven't found any other well-known tool that handles skill installation as well as skills.sh while being more multi-platform friendly.
2. VS Code is currently categorized under "other" because its built-in mechanism doesn't support specifying which plugin to install from a repo. https://github.com/microsoft/vscode/issues/330047
3. Skills installed in this way won't send telemetry since hooks won't be installed.

plugins/azure-skills/skills/discover-azure-skills/SKILL.md
```diff
@@ -0,0 +1,42 @@
...
+
+3. Discover relevant skills
+
+Eliminate the skills that obviously aren't relevant by their names. Then for each remaining skill, read their description from the API: https://raw.githubusercontent.com/microsoft/azure-skills/main/.github/plugins/{plugin-dirname}/skills/{skill-name}/SKILL.md
+
+Use the descriptions to further eliminate skills that aren't relevant.
+
+4. Discover the plugin name of the relevant skills
+
+For each relevant skill, discover their plugin name by reading the `plugin.json` from the GitHub Contents API: https://raw.githubusercontent.com/microsoft/azure-skills/main/.github/plugins/{plugin-dirname}/.plugin/plugin.json
```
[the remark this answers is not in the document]
JasonYeMSFT: This is intended. Eventually skills are consumed from azure-skills repo, not github-copilot-for-azure.

plugins/azure-skills/skills/discover-azure-skills/references/install/other.md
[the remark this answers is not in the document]
JasonYeMSFT: This is intended. The skills will be consumed from the azure-skills repo.

evals/azure-skills/discover-azure-skills/eval.yaml
```diff
@@ -0,0 +1,111 @@
...
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  - name: "Find a skill to generate Azure Kusto graph"
+    prompt: "Find an Azure agent skill that can generate an Azure Kusto graph."
+    tags:
+      type: integration
+      tier: full
+      cost: llm
+      area: output
```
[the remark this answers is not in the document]
JasonYeMSFT: This is intended. This stimuli doesn't early terminate and checks for output.

plugins/azure-skills/skills/discover-azure-skills/SKILL.md
```diff
@@ -0,0 +1,42 @@
...
+license: MIT
+metadata:
+  author: Microsoft
+  version: "0.0.0-placeholder"
+---
+
+Follow these steps to discover the available azure skill matching the given task description.
+
+1. List plugins
+
```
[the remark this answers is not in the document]
JasonYeMSFT: This is a simple skill that doesn't need such complex structure.

plugins/azure-skills/skills/discover-azure-skills/version.json
```diff
@@ -0,0 +1,6 @@
+{
+  "version": "1.0",
```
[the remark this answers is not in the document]
JasonYeMSFT: This is fine. The generated version numbers will be 1.0.x instead of 1.1.x.

plugins/azure-skills/skills/discover-azure-skills/SKILL.md
```diff
@@ -0,0 +1,42 @@
+---
+name: discover-azure-skills
+description: "Searches the Azure skills catalog and recommends installable agent skills by matching an Azure task to skill metadata and plugin installation guidance. WHEN: \"search the Azure skills catalog\", \"find an Azure skill for this task\", \"recommend Azure skills not currently installed\", \"which plugin contains this Azure skill\"."
```
yunjchoi: Does the skill only get triggered when the user shares an intent to discover a skill?

yunjchoi: Would it be possible to update the description to clarify that this skill serves as a fallback for Azure tasks that are not covered by any enabled skills? It should also explicitly instruct the model not to invoke this skill when an existing enabled skill already supports the user's requested task.

evals/azure-skills/discover-azure-skills/eval.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - type: skill-invocation
+        config:
+          required:
+            - discover-azure-skills
+      - type: output-not-matches
+        config:
+          pattern: "(?i)fatal error|unhandled exception|stack trace"
+
+  - name: "Find a skill to generate Azure Kusto graph"
+    prompt: "Find an Azure agent skill that can generate an Azure Kusto graph."
```
yunjchoi: If we update this skill so that it's invoked based on the user's scenario rather than the intent to discover a skill, I think this evaluation will need to be updated as well.

plugins/azure-skills/skills/discover-azure-skills/SKILL.md
JasonYeMSFT: > Does the skill only get triggered when the user shares an intent to discover a skill?

This is what I plan to do. In my opinion, making this skill to proactively recommend skills is too aggressive and will also run into scalability issues. To recommend a skill from everything we offer without knowing what the user may ask, we will need to load everything into the context. It's not possible load everything into the context as we add more plugins/skills.

RickWinter:
Does this skill require me to ask for discovery intentionally? Would the customer need to do "discover" explicitly? Should the skill be more of a guide that happens based on intent to do something Azure?

JasonYeMSFT:
> Does this skill require me to ask for discovery intentionally? Would the customer need to do "discover" explicitly? Should the skill be more of a guide that happens based on intent to do something Azure?

I expect the user to at least express the intent to learn how to do something in Azure to trigger this skill. I intend to avoid having the skill triggered when it's not the case.

We can discuss more details of the design in this issue. https://github.com/microsoft/GitHub-Copilot-for-Azure-pr/issues/358

evals/azure-skills/discover-azure-skills/eval.yaml
JasonYeMSFT: I rewrote the test prompts to cover these cases:
- Ask agent to search a skill not in the core plugin
- Ask agent to find a skill in the core plugin
- Ask a generic how-to question
- Give a direct command asking the agent to do something

plugins/azure-skills/skills/discover-azure-skills/SKILL.md
JasonYeMSFT: I reworked the skill to allow triggering it from direct task commands when no existing skill/tool in the context seems relevant. I added a eval suite to check skill invocation in such case.

JasonYeMSFT:
I was able to implement the discover-azure-skills skill to make it work like a catch-all skill when the user issue commands for Azure tasks without explicitly asking for finding new skills. However, given that we only have one additional plugin available right now, the overhead we introduce by making agents trying hard to find new skills is very significant. I still prefer taking an conservative approach for now, grow the platform to have more plugins/skills and tune the skill to make more aggressive recommendations later.

hooks/azure-skills/scripts/discover-skills.ps1
```diff
@@ -0,0 +1,4 @@
+Write-Output '{"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": "Make sure to use discover-azure-skills once when the user asks you to execute a task related to Azure if there is no relevant skill/mcp tool in the context. Avoid using discover-azure-skills when the user asks informational questions on how to do something."}}'
+
+# Output success to stdout (required by hooks)
+Write-Success
```
RickWinter: `Write-Success` is defined only inside `track-telemetry.ps1` and this script does not import it. Every Windows Copilot CLI or VS Code session therefore emits a command-not-found error during `SessionStart` after writing the context. Replace this with a local success response or an explicit `exit 0` so the hook completes cleanly.

hooks/azure-skills/scripts/discover-skills.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
RickWinter: This script has a shebang but is committed as mode `100644`, while the existing hook script is executable. The Copilot CLI, Claude Code, and Cursor manifests invoke it as a command, so Unix sessions can fail with `Permission denied` before the discovery context is emitted. Mark this file executable in Git.

JasonYeMSFT: The first iteration of this skill won't be aggressive in recommending skills. This script is removed for the 1st iteration.

hooks/azure-skills/scripts/discover-skills.ps1
JasonYeMSFT: The first iteration of this skill won't be aggressive in recommending skills. This script is removed for the 1st iteration.

plugins/azure-skills/skills/discover-azure-skills/references/install/other.md
JasonYeMSFT: VS Code has fixed the issue I linked.

yunjchoi (dismissed):

evals/azure-skills/discover-azure-skills/eval.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+tags:
+  type: integration
+  skill: discover-azure-skills
+
+defaults:
+  runs: 5
+  timeout: "10m"
+  executor: integration-test-agent-runner
+  model: claude-sonnet-4.6
```
tmeschter: Is this model still available to us?

tmeschter (commented):
The comment references a session start hook but I don't see that in the PR.

JasonYeMSFT:
I updated the description. The session start hook has been removed from the changes of this PR.

evals/azure-skills/discover-azure-skills/eval.yaml
JasonYeMSFT: For some reason Copilot CLI made it available again. I updated it to claude-sonnet-5 to stay consistent with the rest of the eval suites.

tmeschter (approved):

PR #3146, opened 2026-09-02
fix: add cross-platform Cursor telemetry runner

tmeschter:
## Description

Adds a Node.js telemetry dispatcher for Cursor so the existing PowerShell hook runs on Windows and the Bash hook runs on macOS and Linux. The dispatcher preserves hook input/output and child exit status, while resolving the platform scripts relative to itself.

Since we're already dependent on `npx` to invoke Azure MCP to report the telemetry events we should be OK to take a dependency on `node` as well.

Updates both Cursor telemetry events to use the dispatcher and extends the telemetry hook tests with platform selection, failure propagation, pass-through behavior, and an installed-plugin fixture layout.

## Checklist

- [x] **If modifying skill descriptions:** verified routing correctness with integration tests (not applicable; no skill descriptions changed)

## Testing

- `npm test -- --run src/__tests__/telemetry-hooks.test.ts`
- `npm --prefix scripts run typecheck`
- `npm --prefix scripts run lint`
- `npm run build`

## Related Issues

Fixes microsoft/GitHub-Copilot-for-Azure-pr#363

### DIFF
--- a/hooks/cursor-hooks.json
+++ b/hooks/cursor-hooks.json
@@ -4,13 +4,13 @@
     "postToolUse": [
       {
         "type": "command",
-        "command": "bash ${CURSOR_PLUGIN_ROOT}/hooks/scripts/track-telemetry.sh"
+        "command": "node \"${CURSOR_PLUGIN_ROOT}/hooks/scripts/track-telemetry.js\""
       }
     ],
     "afterMCPExecution": [
       {
         "type": "command",
-        "command": "bash ${CURSOR_PLUGIN_ROOT}/hooks/scripts/track-telemetry.sh"
+        "command": "node \"${CURSOR_PLUGIN_ROOT}/hooks/scripts/track-telemetry.js\""
       }
     ]
   }
--- a/scripts/src/__tests__/telemetry-hooks.test.ts
+++ b/scripts/src/__tests__/telemetry-hooks.test.ts
@@ -1,6 +1,7 @@
 import { spawnSync } from "node:child_process";
 import {
   chmodSync,
+  cpSync,
   existsSync,
   mkdtempSync,
   mkdirSync,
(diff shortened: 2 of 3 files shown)


scripts/src/__tests__/telemetry-hooks.test.ts
```diff
@@ -61,17 +92,7 @@ function fixture(name: string): Record<string, unknown> {
...
-    ".cursor",
-    "plugins",
-    "cache",
-    "cursor-public",
-    "azure",
-    "revision",
-    "skills",
-    "azure-cost",
-  );
+  const skillRoot = join(PLUGIN_ROOT, "skills", "azure-cost");
```
JasonYeMSFT, line 95: This needs to use a different skill after this PR merges. https://github.com/microsoft/GitHub-Copilot-for-Azure/pull/3151

hooks/cursor-hooks.json
```diff
@@ -4,13 +4,13 @@
     "postToolUse": [
       {
         "type": "command",
-        "command": "bash ${CURSOR_PLUGIN_ROOT}/hooks/scripts/track-telemetry.sh"
+        "command": "node \"${CURSOR_PLUGIN_ROOT}/hooks/scripts/track-telemetry.js\""
```
JasonYeMSFT, line 7: Let's try to actively suppress error in case the client machine doesn't have node installed.

JasonYeMSFT (approved):

hooks/cursor-hooks.json
tmeschter: I'll handle that in a follow up.

tmeschter: Tracked in #3156 so we can address and test missing-Node suppression as a focused follow-up.

PR #3155, opened 2026-09-03
chore: restrict dashboard access control

JasonYeMSFT:
## Checklist

- [ ] **If modifying skill descriptions:** verified routing correctness with integration tests (In `tests/`, `npm run test:vally -- --plugin <plugin-dirname> --skill <skill>`)


tmeschter (approved):
We should take this change now, but we should also think about how we can set up the eval runs and dashboard so that the entire system is internal.

PR #3172, opened 2026-09-09
fix: split telemetry scripts by client

### LINKED ISSUE #3144: Split telemetry scripts by client
Copilot, VS Code, Claude, Cursor, and other clients we try to support all have somewhat different behavior when it comes to hooks, including different schemas for the input JSON objects. Right now the differences in behavior are handled within the various script functions, leading to a lot of conditional logic and functions that are longer and harder to understand than they need to be.

We should break up the telemetry.ps1/telemetry.sh. The existing telemetry.ps1/.sh will continue to serve as the entry point for the hooks and handle client detection. At that point it will hand off to client-specific logic in client-specific files. Within these files we can assume we already know the client which will significantly simplify the logic.

This should make it easier to maintain the telemetry scripts over the long term with changes for one client less likely to break telemetry reporting for others.

tmeschter:
## Summary

- keep `track-telemetry.sh` and `track-telemetry.ps1` as shared entry points for client detection, logging, version lookup, path ownership, publication, and hook responses
- move Copilot CLI, Claude Code, Cursor, and VS Code payload parsing and event classification into paired client-specific Bash and PowerShell handlers
- use a shared installation-path allowlist so any client can report skills installed by another client
- expand fixture-driven regression coverage across all clients, shells, event types, cross-client installation layouts, ownership deduplication, and input encoding

## Validation

- `npm test -- --run src/__tests__/telemetry-hooks.test.ts` - 55 tests passed
- `cd scripts && npm run typecheck`
- `cd scripts && npm run lint`
- `npm run check:shell-scripts`
- `npm run build`

Closes #3144

### DIFF
--- a/hooks/scripts/clients/copilot-cli.sh
+++ b/hooks/scripts/clients/copilot-cli.sh
@@ -1,0 +1,34 @@
+# Sourced by track-telemetry.sh after Copilot CLI detection.
+
+process_telemetry_event() {
+    local raw_input="$1"
+    local tool_name
+    local path_to_check
+    local path_lower
+    local candidate
+
+    clientName="copilot-cli"
+    sessionId=$(extract_json_field "$raw_input" "sessionId")
+    tool_name=$(extract_json_field "$raw_input" "toolName")
+    [ -n "$tool_name" ] || return 0
+
+    if [ "$tool_name" = "skill" ]; then
+        candidate=$(extract_toolargs_field "$raw_input" "toolArgs" "skill")
+        track_skill_by_name "$candidate"
+    fi
+
+    path_to_check=$(extract_toolargs_path "$raw_input" "toolArgs")
+    path_lower=$(normalize_path_lower "$path_to_check")
+    if [ "$tool_name" = "view" ] && [ -n "$path_to_check" ] \
+        && is_azure_skills_path "$path_lower" && [[ "$path_lower" == *"/skill.md" ]]; then
+        track_skill_read "$path_to_check"
+    fi
+
+    if [[ "$tool_name" == azure-* ]]; then
+        track_tool_invocation "$tool_name"
+    fi
+
+    if [ -z "$skillName" ] && [ -n "$path_to_check" ] && is_azure_skills_path "$path_lower"; then
+        capture_reference_path "$path_to_check"
+    fi
+}
--- a/hooks/scripts/clients/copilot-cli.sh
+++ b/hooks/scripts/clients/copilot-cli.sh
@@ -19,7 +19,7 @@ process_telemetry_event() {
 
     path_to_check=$(extract_toolargs_path "$raw_input" "toolArgs")
     path_lower=$(normalize_path_lower "$path_to_check")
-    if [ "$tool_name" = "view" ] && [ -n "$path_to_check" ] \
+    if is_file_read_tool "$tool_name" && [ -n "$path_to_check" ] \
         && is_azure_skills_path "$path_lower" && [[ "$path_lower" == *"/skill.md" ]]; then
         track_skill_read "$path_to_check"
     fi
--- a/hooks/scripts/track-telemetry.ps1
+++ b/hooks/scripts/track-telemetry.ps1
@@ -261,6 +261,13 @@ function Get-ToolInputPath {
     return $null
 }
 
+# Clients have used more than one file-read tool name over time. Accept the
+# complete compatibility list regardless of which client emitted the payload.
+function Test-FileReadTool {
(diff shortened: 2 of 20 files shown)


hooks/scripts/track-telemetry.ps1
```diff
@@ -1,202 +1,235 @@
...
+    $targetPathNorm = Normalize-PathLower $TargetPath
+    return $targetPathNorm.StartsWith("$skillsRootNorm/", [System.StringComparison]::Ordinal)
+}
+
+# Match local plugin development paths configured through --plugin-dir.
+function Test-LocalSkillPath {
+    param([string]$NormalizedPath)
+    if (-not $NormalizedPath -or -not $env:AZURE_SKILLS_PLUGIN_ROOT) { return $false }
+    $localRoot = Normalize-PathLower $env:AZURE_SKILLS_PLUGIN_ROOT
+    return $NormalizedPath.Contains("$localRoot/skills/")
```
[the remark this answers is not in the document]
tmeschter: Fixed in bcd0b992. The normalized PowerShell root now uses `TrimEnd('/')`, which handles both trailing slash styles after path normalization. Added regression tests for trailing `/` and `\` values.

hooks/scripts/track-telemetry.sh
```diff
@@ -1,488 +1,412 @@
...
 
-# Extract the skill version from a SKILL.md frontmatter (metadata.version).
-# Prints nothing if the file or version cannot be read.
+# Match local plugin development paths configured through --plugin-dir.
+is_local_skill_path() {
+    local normalized_path="$1"
+    [ -n "$AZURE_SKILLS_PLUGIN_ROOT" ] || return 1
+    local local_root
+    local_root=$(normalize_path_lower "$AZURE_SKILLS_PLUGIN_ROOT")
+    [[ "$normalized_path" == *"${local_root}/skills/"* ]]
```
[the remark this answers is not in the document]
tmeschter: Fixed in bcd0b992. The Bash helper now removes trailing normalized slashes before appending `/skills/`. Added regression tests for both trailing `/` and `\` values.

scripts/src/__tests__/telemetry-hooks.test.ts
```diff
@@ -268,69 +385,130 @@ describe("Cursor telemetry dispatcher", () => {
   );
 });
 
-describe.each(shells)("Cursor telemetry hook ($name)", shell => {
-  const skillRoot = createCursorSkillCache();
+describe.each(shells)("Telemetry hook ($name)", shell => {
+  it.each(clientCases)("reports a $id SKILL.md read as a skill invocation", client => {
```
JasonYeMSFT, line 401: Is this $id replaced by clientCase.id by the test framework?

hooks/scripts/clients/copilot-cli.sh
```diff
@@ -0,0 +1,34 @@
...
+    [ -n "$tool_name" ] || return 0
+
+    if [ "$tool_name" = "skill" ]; then
+        candidate=$(extract_toolargs_field "$raw_input" "toolArgs" "skill")
+        track_skill_by_name "$candidate"
+    fi
+
+    path_to_check=$(extract_toolargs_path "$raw_input" "toolArgs")
+    path_lower=$(normalize_path_lower "$path_to_check")
+    if [ "$tool_name" = "view" ] && [ -n "$path_to_check" ] \
```
JasonYeMSFT: Copilot CLI also uses tool named "Read". Let's apply the same tool name list we used to have to every client. I am seeing reference file read payloads result in shouldTrack=false here.

JasonYeMSFT (changes requested):

scripts/src/__tests__/telemetry-hooks.test.ts
tmeschter: Yes. This is Vitest's `it.each` object interpolation syntax: `$id` is replaced with the `id` property from the current `clientCases` object. For example, the generated test title includes `copilot`, `claude`, `cursor`, or `vscode`.

hooks/scripts/clients/copilot-cli.sh
tmeschter: Fixed in 27dfb51d. All four client handlers now use the shared compatibility list (`view`, `Read`, and `read_file`) instead of assuming one read-tool name per client. Added a client-by-tool-name matrix covering all 12 combinations in both Bash and PowerShell; all 83 telemetry tests pass.

JasonYeMSFT (approved):

PR #3186, opened 2026-09-11
add azure service retirement skill and related routing evaluation

mugurehiya:
## Checklist

- [ ] **If modifying skill descriptions:** verified routing correctness with integration tests (In `tests/`, `npm run test:vally -- --plugin <plugin-dirname> --skill <skill>`)

### DIFF
--- a/plugins/azure-skills/skills/azure-service-retirement/SKILL.md
+++ b/plugins/azure-skills/skills/azure-service-retirement/SKILL.md
@@ -1,0 +1,31 @@
+---
+name: azure-service-retirement
+description: "Find Azure Advisor retirement deadlines and affected resources. WHEN: Azure service, SKU, runtime, API, feature, or version retirement, deprecation, shutdown, end of support, end of life; AKS or Kubernetes version support ending; Service Health retirement ID; retirement-required migration. DO NOT USE FOR: upgrades, patching, modernization, migrations, or Advisor requests without explicit retirement lifecycle context."
+license: MIT
+metadata:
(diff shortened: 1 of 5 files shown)


plugins/azure-skills/skills/azure-service-retirement/SKILL.md
```diff
@@ -0,0 +1,31 @@
...
+| Property | Value |
+|---|---|
+| MCP tool | `advisor_recommendation_list` |
+| Required | `subscription` |
+| Optional filters | `subCategory`, `trackingIds`, `retirementDate` |
+
+## Workflow
+
+1. Infer `subscription` only from established context; otherwise ask for it.
+2. Read and follow [Retirement Query Rules](references/query-rules.md).
```
tmeschter, line 25: If the agent is always expected to read query-rules.md, the content should be in this file instead of a separate file.

tmeschter (commented):
The major thing you're missing are evals that check the end-to-end behavior of the skill--not just that the skill is invoked, but that the appropriate MCP tools are invoked with the proper arguments, and that the output of the LM matches expectations.

tmeschter:
>The major thing you're missing are evals that check the end-to-end behavior of the skill--not just that the skill is invoked, but that the appropriate MCP tools are invoked with the proper arguments, and that the output of the LM matches expectations.

And once you have those evals you can use them to compare the LM behavior with and without the skill and establish whether or not it helps. You can also check those evals on different LMs and see if the skill may need tweaks.

